PR #39, opened 2016-11-03
WIP: Allow a configurable directory instead of using package layout

### LINKED ISSUE #24: Allow direct configuration of the fragment directory.
I think that towncrier should be a bit less opinionated and should not force you to use a specific package format and have the fragment directory under a specific location and with a specific name...

at least at the API level.

For very simple project I would like to have something like this

```
from towncrier import update
update(fragments_path='some/path/to/fragments/dir', update_file='some/path/to/NEWS', version='1.2.3')
```

no import of the source code should be done and the tool should be used for non-python projects

Thanks!

dstufft:
Instead of ``<package>/<section>/<newsfragments>`` being the only real option, this allows instead to define your own top level news fragment directory. This could be something like ``./news`` and then inside of that directory there is either news fragments directly, or sub directories for different sections.

Fixes #24

### DIFF
--- a/src/towncrier/__init__.py
+++ b/src/towncrier/__init__.py
@@ -62,9 +62,16 @@ def __main(draft, directory, project_version, project_date):
         ("misc", {"name": "Misc", "showcontent": False}),
     ])
 
+    if config.get("directory"):
+        base_directory = os.path.abspath(config["directory"])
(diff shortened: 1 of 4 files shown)


hawkowl:
sure!

PR #41, opened 2016-11-20
TOML & other stuff

hawkowl:


tardyp:
@hawkowl would you like to finish up this work and make a release? this would be an important change so that buildbot can fully uses this project

PR #42, opened 2016-12-07
dont swallow the toml parsing exception

tardyp:
Even if toml error reporting is far from perfect,
if you have a syntax error in your toml, you want to be informed
rather than switching to .ini parsing


tardyp:
@hawkowl ping

PR #66, opened 2017-08-03
Add issue_format= configuration

njsmith:
By default you still get '#foo' for integers, and 'foo' for
non-integers, but you can also define a format string to generate, for
example, a link to the issue tracker.

Also:

- Redo sorting in a more rational way, based on sorting values instead
  of rendered text.

- Add issue_format= to towncrier's own pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,7 @@
     package = "towncrier"
     package_dir = "src"
     filename = "NEWS.rst"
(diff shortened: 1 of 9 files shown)


njsmith:
Notes:

- towncrier seems to often refer to the numbers as "ticket numbers", but since everyone including towncrier seems to be using github these days, I called it `issue_number`. If you want it to be `ticket_number` instead I can change it :-)

- This and #64 are going to have merge conflicts all over the place...

Issue #68, opened 2017-08-03
Line-wrapping logic is too eager, breaks URLs


njsmith:
Towncrier just generated an entry for me like:
```
- Don't crash on Windows (`#10 <https://github.com/python-trio/trustme/issues/10>`__)
```

Or, well, it was supposed to. What it actually generated was:
```
- Don't crash on Windows (`#10 <https://github.com/python-
  trio/trustme/issues/10>`__)
```

This is not very helpful. To be safe, it should only break lines on whitespace.

PR #69, opened 2017-08-03
Don't break long lines mid-word or at hyphens, to avoid breaking URLs

njsmith:
Fixes gh-68

This should probably have a test added, but I don't dare touch those
files and generate even more conflicts while gh-66 and gh-68 are
outstanding...


hawkowl:
@njsmith A test would be nice

PR #91, opened 2017-09-14
Add help entry for --version

obestwalter:
I was wondering if this is possible but did not see the option with `towncrier --help`. I tried it anyway, saw that it works and came here to tell the world about it :)


obestwalter:
The faillures have nothing to do with my innocent little change. Interpreters not found.

michaeljoseph:
👍 
@hawkowl maybe travis needs a nudge?

PR #93, opened 2017-09-15
draft output should go completely to stdout

obestwalter:
At least to me it does make no sense that the header goes to stderr and the rest to stdout.

### DIFF
--- a/src/towncrier/__init__.py
+++ b/src/towncrier/__init__.py
@@ -103,7 +103,7 @@ def __main(draft, directory, project_version, project_date, answer_yes):
             "What is seen below is what would be written.\n",
             err=to_err,
         )
-        click.echo("%s%s" % (top_line, rendered))
+        click.echo("%s\n%s" % (top_line, rendered))
     else:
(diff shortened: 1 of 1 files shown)


src/towncrier/__init__.py
```diff
@@ -103,8 +103,7 @@ def __main(draft, directory, project_version, project_date, answer_yes):
             "What is seen below is what would be written.\n",
             err=to_err,
         )
-        click.echo(top_line, err=to_err)
-        click.echo(rendered)
+        click.echo("%s%s" % (top_line, rendered))
```
michaeljoseph: Looks like this breaks the tests since you dropped the newline.
Maybe just remove the `err` kwarg?

obestwalter:
Thanks @michaeljoseph - added the missing newline.

michaeljoseph:
The test failures look like tox 😤 travis miscommunications 👊 
So 👍 from me.

hawkowl:
Travis removed pypy and python3.3. But this seems fine, thanks! The original idea was that the draft could be piped out, but I don't think many people would want to do that...

PR #95, opened 2017-10-18
Add support for custom config files

EliotBerriot:
This is a simple PR that introduce a new `--config` flag in the CLI, which accept a path to a custom config file. 

We needed this in one project :)

### DIFF
--- a/src/towncrier/_settings.py
+++ b/src/towncrier/_settings.py
@@ -22,7 +22,13 @@ _underlines = ["=", "-", "~"]
 
 
 def load_config(from_dir):
-    fn = os.path.join(from_dir, "pyproject.toml")
+    if os.path.isfile(from_dir):
+        # maybe from_dir is actually a config_file path
+        fn = from_dir
+    else:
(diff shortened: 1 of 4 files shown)


src/towncrier/_settings.py
```diff
@@ -22,7 +22,13 @@
 
 
 def load_config(from_dir):
-    fn = os.path.join(from_dir, "pyproject.toml")
+    if os.path.isfile(from_dir):
+        # maybe from_dir is actually a config_file path
+        fn = from_dir
+    else:
+        # we fallback to default behaviou of finding the config_file
```
ewjoachim: behaviour

EliotBerriot: Fixed thanks :)

EliotBerriot:
We don't need or use this anymore, so feel free to close the PR :)

hawkowl:
It's kind of intended to have it in the one place to match with the pyproject PEP, so I am not sure I'd want to change it. Thanks though, I'll reconsider if more people think it's an idea.

PR #96, opened 2017-10-18
Update README w/ up-to-date info

webknjaz:


hawkowl:
Thanks for pointing this out! In my 3.3 removal patch (since Travis dropped it), I just changed it to "3.4+" for simplicity, so I'll close this. Thanks again for noticing it needed updating!

PR #100, opened 2017-12-27
Make bin/towncrier work when run from other directories

njsmith:
This is handy for running towncrier from a checkout.


njsmith:
Test failures are because the CI is broken and trying to test on missing python versions.

PR #102, opened 2017-12-27
Several fixes to newsfragment extension handling

### LINKED ISSUE #99: Newsfragment removal fails for files with extra extensions
In #65, I made towncrier ignore extensions on newsfragment filenames, so that you could use names like `newsfragments/17.feature.rst` and your editor would understand this was a `.rst` file.

Apparently I missed a case, because when you use towncrier in for-real-release-mode, it currently sees the above file, moves it into your release notes, and then tries to remove `newsfragments/17.feature`, which fails because that's not what the file is called:

```
~/trio$ towncrier
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
Removing news fragments...
/home/njs/trio/newsfragments/1.removal
/home/njs/trio/newsfragments/1.removal
/home/njs/trio/newsfragments/241.removal
/home/njs/trio/newsfragments/241.removal
/home/njs/trio/newsfragments/377.removal
/home/njs/trio/newsfragments/377.removal
/home/njs/trio/newsfragments/136.feature
/home/njs/trio/newsfragments/377.feature
/home/njs/trio/newsfragments/136.feature
/home/njs/trio/newsfragments/277.bugfix
/home/njs/trio/newsfragments/277.bugfix
Is it okay if I remove those files? [Y/n]: 
fatal: pathspec '/home/njs/trio/newsfragments/1.removal' did not match any files
Done!
```

### LINKED ISSUE #101: towncrier happily reads 1.feature.rst~, ignoring 1.feature.rst
Another ugly bit of fallout from #65. Sorry.

I see two options:

- Use `git ls-files` to find newsfragments, so we only pick up files that are actually under VCS

- Instead of blindly accepting anything that looks like `<anything>.<category>.*`, let projects specify an extension in `pyproject.toml` and then only look for `<anything>.<category>.<configured extension>`

njsmith:
This still needs tests, but it fixes #99 and ameliorates #101 well enough to
let me generate some actual release notes. It also adds some better
error checking. (Each commit is self-contained.)

As noted in #101, it probably would be better if towncrier had a
configuration option specifying what filename extension to use,
instead of blindly accepting all extensions. That would have been
cleverer of me. I think all these changes are probably still useful
in their own right though.

### DIFF
--- a/src/towncrier/__init__.py
+++ b/src/towncrier/__init__.py
@@ -67,8 +67,8 @@ def __main(draft, directory, project_version, project_date, answer_yes):
             directory, config['package_dir'], config['package']))
         fragment_directory = "newsfragments"
 
-    fragments = find_fragments(
-        base_directory, config['sections'], fragment_directory)
+    fragments, fragment_filenames = find_fragments(
+        base_directory, config['sections'], fragment_directory, definitions)
 
     click.echo("Rendering news fragments...", err=to_err)
 
@@ -114,9 +114,7 @@ def __main(draft, directory, project_version, project_date, answer_yes):
         stage_newsfile(directory, config['filename'])
 
         click.echo("Removing news fragments...", err=to_err)
-        remove_files(
-            base_directory, fragment_directory, config['sections'],
-            fragments, answer_yes)
+        remove_files(fragment_filenames, answer_yes)
 
         click.echo("Done!", err=to_err)
 
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -78,6 +78,11 @@ def find_fragments(base_directory, sections, fragment_directory, definitions):
             fragment_filenames.append(full_filename)
             with open(full_filename, "rb") as f:
                 data = f.read().decode('utf8', 'replace')
+            if (ticket, category) in file_content:
+                raise ValueError(
+                    "multiple files for {}.{} in {}"
+                    .format(ticket, category, section_dir)
+                )
             file_content[ticket, category] = data
 
         content[key] = file_content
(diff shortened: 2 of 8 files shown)


hawkowl:
Tests fail :(

njsmith:
Ah, darn. Tests are failing because the tests are running on non-git-managed directories. I was assuming from the git stuff that towncrier already does that that wasn't supported, but maybe I'm wrong.

What is towncrier's story on git?

hawkowl:
Towncriers only Git integration allows deleting topfiles from Git. So it's
pure "better if it's available".

njsmith:
Sigh fine. Dropped that commit, rebased on master, and added tests for the other things. Still need to come up with a more reasonable fix for #101, probably switching to an explicit newsfragment extension configuration thing, but these two changes are least contain the damage and are ready to go in.

Interesting note: github apparently does not understand git. I originally wrote the "Factor out some common code" commit later, and then during rebase moved it earlier. So despite what github is saying, it actually comes *first* in git's history. Github's UI apparently sorts git commits based on their commit time metadata, rather than using history graph.

njsmith:
Annnnd finally tests are passing and codecov's happy.

nicoddemus:
This is great, I just stumbled on this in `pytest`. It would be great to have this merged and released. 😁

nicoddemus:
Any plans for a new release which includes this? 😁 

It's the 4th pytest release that I have to remove the files by hand (usually noticed well after the release was made). 😓

nicoddemus:
Not at all, awesome. Thanks @hawkowl and @njsmith for the fix. 👍

PR #106, opened 2018-01-27
make fragment file removal work with extensions 

### LINKED ISSUE #99: Newsfragment removal fails for files with extra extensions

obestwalter:
fixes #99 

When generating the list of files to be removed it will be checked if the generated path exists for a collection of accepted extensions. To keep this change simple it is hardcoded to <no extension>, .rst and .txt - this could be made configurable in the future.

### DIFF
--- a/src/towncrier/_git.py
+++ b/src/towncrier/_git.py
@@ -59,7 +59,7 @@ def find_file_path(section_dir, category_name, ticket):
         if os.path.exists(path):
             return path
 
-    raise FileNotFoundError(
+    raise Exception(
         "can't find the file for fragment %s at %s - "
         "use no extension or one of the accepted (%s)" % (
             namestub, section_dir, accepted_extensions[1:]))
--- a/src/towncrier/_git.py
+++ b/src/towncrier/_git.py
@@ -24,10 +24,8 @@ def remove_files(base_dir, fragment_directory, sections, fragments,
             for tickets in category_items.values():
 
                 for ticket in tickets:
-
-                    filename = str(ticket) + "." + category_name
-                    to_remove.append(os.path.join(
-                        section_dir, filename))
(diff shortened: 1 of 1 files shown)


hawkowl:
@obestwalter I think this may have also been fixed in a PR I hadn't got to that I just merged. Does current master fix your problem? If not, I'll fix up the branch conflicts.

obestwalter:
Thank you @hawkowl - I dont' have the possibility to try it at the moment, but I looked at the merged PR and it looks like this will also do the trick. If not I will open a new PR basing on top of the new code.

PR #107, opened 2018-03-22
make project name configurable from cli

davidhyman:
I found that:
> It is usable by projects written in other languages, provided you give it the version of the project when invoking it.

Was not the case, as towncrier wanted to load a Python package in order to get the project name.

I've added another cli option to ignore that behaviour, in the same way as the version does. It now works in non-python projects 🌐

### DIFF
--- a/src/towncrier/__init__.py
+++ b/src/towncrier/__init__.py
@@ -35,11 +35,23 @@ def _get_date():
 @click.option('--date', 'project_date', default=None)
 @click.option('--yes', 'answer_yes', default=False, flag_value=True,
               help="Do not ask for confirmation to remove news fragments.")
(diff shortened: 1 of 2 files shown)


davidhyman:
@hawkowl sorted - but now travis is being uncooperative, doesn't seem to be picking up the PR 😞 
If I can do anything to help it along, let me know!

acabarbaye:
It would be awesome if this fix could be merged!

mauritsvanrees:
But I see you are busy at PyCon. Fair enough. ;-)
https://youtu.be/A5tPhUKljuY

mauritsvanrees:
Thank you! I have immediately released [zestreleaser.towncrier](https://pypi.org/project/zestreleaser.towncrier/) 1.0.0b3 which makes use of the new features.

Issue #111, opened 2018-04-10
Make package optional


mauritsvanrees:
I am trying out`towncrier`. I see that the config file requires a `package` entry, which is used to get the version and the name for the `title_format`. In my case, this should not be needed:

- I will pass the version on the command line.
- I have this in my config file: `title_format = "{version} ({project_date})"`. So the package name is not shown here.

(The package name is passed to the main template as well, but by default it is not used there.)

So I would like to avoid setting the package name in the configuration, for two reasons really:

1. Like I said above: I do not need it.
2. `towncrier` fails if it cannot import the package. This assumes that the package-to-be-released is installed in the same virtualenv as `towncrier`, or all its dependencies have been installed, which in most cases is not true for me.

For the second reason it would help if https://github.com/hawkowl/towncrier/pull/107 is released, so I can pass the package name on the command line. A workaround would be to cheat by putting `package = "towncrier"` in the configuration, which can obviously be imported. :-)

What I would like best:

- If package is not defined in the config and not passed on the command line, use an empty string. (This may mean that the version cannot be determined automatically, but that can already be passed on the command line.)
- If `package` is defined but cannot be imported, fall back to using `package` as the name. For Python projects, 99% of the time this will be the same anyway. (I could make this a separate issue, but it is related.)

hawkowl:
Merged! Will do a release.

PR #112, opened 2018-04-23
Made `package` optional.

mauritsvanrees:
When the version is passed on the command line, and the `title_format` does not use the package name, and it is not used for the path to the news fragments, then no package name is needed, so we should not enforce it.

Closes issue #111.


hawkowl:
Thanks!

PR #119, opened 2018-05-30
Allow for multiple fragments with the same ticket and type

Cadair:
This extends the fragment format to be `<issue>.<type>(.<counter>)` i.e have an optional `.counter` after the type, which needs to be an integer.


hawkowl:
@Cadair Could you write a changelog entry for this? :)

Cadair:
done :)

Cadair:
ping? :angel: 

Anything else that needs to happen on this?

PR #127, opened 2018-09-05
Add a --pyproject option to towncrier.check

### LINKED ISSUE #123: Using pyproject.toml triggers special pip behavior
I shouldn't have to kick pip into a new, incompatible mode just to configure towncrier.  towncrier should provide an alternate configuration mechanism that avoids pep 518 and pyproject.toml completely.

exarkun:
Allows for a work-around for #123


hawkowl:
thanks @exarkun

dstansby:
Would it be possible to add this as a command line option too?

Issue #132, opened 2018-11-02
Could there be a new news fragment for major changes such as `.major`?


acabarbaye:
It sometimes happens that significant refactoring or architecture changes should be described or treated differently than newly introduced features. Therefore, could there be a new news fragment for such changes like `.major` or so and which would have a different section in the Changelog?

Sup3rGeo:
I was doing my research for a good way of automatically dealing with versioning and it seems that having this `.major` would allow to define semantic versioning automatically from the news.

Although I am not sure if `major` would be the best name for it:

Qualitative description: semantic versioning item
- bugfix: patch
- new, backward compatible feature: minor
- incompatible changes: major

So it could be called maybe `incompatible`, or something better (but still a qualitative description, following the logic of the table).

acabarbaye:
@Sup3rGeo : This project is doing exactly what you mention about automatic semantic versioning based on news files: https://github.com/ARMmbed/autoversion. However, as you mention, a news file describing `major` changes responsible for `major` version bump is needed to be fully complete.

hawkowl:
You can define custom fragments using the `toml` config: https://github.com/hawkowl/towncrier#further-options

PR #133, opened 2018-11-07
Add a `major` fragment

acabarbaye:
See issue #132

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,7 @@ News Fragments
 ``towncrier`` has a few standard types of news fragments, signified by the file extension.
 These are:
 
+- ``.major``: Signifying a major change.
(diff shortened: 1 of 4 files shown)


kinkerl:
i woud love to see this upstream. i would have used this today :)

davidhyman:
@acabarbaye this PR needs a newsfile 🙂

acabarbaye:
@hawkowl : Could you have a look at this PR so that it can be merged ?

hawkowl:
The existing defaults are matching Twisted's model, which I'm not sure I want to change.

You probably just want to configure it per-project, as attrs does it? https://github.com/python-attrs/attrs/blob/master/pyproject.toml

PR #143, opened 2019-03-11
Accept multiple dots in filenames

### LINKED ISSUE #142: towncrier misses newsfragment filenames like `fix-1.2.3.bugfix`
In projects that don't use the convention of embedding ticket numbers within the newsfragment names, filenames like `fix-1.2.3.bugfix` are possible. towncrier misses such newsfragments without any errors and there's high possibility that this won't be noticed until much later. towncrier should accept such filenames and either add them to the release notes or throw an error.

p12tic:
Fixes #142.

### DIFF
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -11,6 +11,28 @@ from collections import OrderedDict
 from jinja2 import Template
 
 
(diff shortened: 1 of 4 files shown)


hawkowl:
Thanks!

PR #147, opened 2019-05-06
Move version top strings to the Jinja template

hawkowl:


seberg:
This is nice, but seems backward incompatible if someone uses their own template file and the default string? I do not mind this, but there seems to options:
1. Note in the release notes that users who want the old behaviour need to to set the `title_format` manually now.
2. Use my version of forcing users who want to use this to set `title_format = false`?

PR #156, opened 2019-07-17
ENH: Allow to manually define which pyproject.toml file to use

seberg:
This adds `--pyproject` to `towncrier`, the same way as what exists
for `towncrier.check`.

---

This probably needs a few additional things (such as tests)? I tried it locally briefly, and it seems to work (is pretty straight forward after all).

### DIFF
--- a/src/towncrier/__init__.py
+++ b/src/towncrier/__init__.py
@@ -13,7 +13,7 @@ import pkg_resources
 
 from datetime import date
 
-from ._settings import load_config
+from ._settings import load_config, load_config_from_file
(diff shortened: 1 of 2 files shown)


seberg:
xref gh-157 which solves the same issue (uses `--config` though).

hawkowl:
I merged 157, I prefer the config name over pyproject (since pyproject is python specific).

Thanks!

seberg:
I agree config is nicer, could change `towncrier.check` to also accept `--config`, but low priority.

PR #158, opened 2019-07-25
ENH: Allow no bullet point framgents and file templating

seberg:
This adds a single file mode to allow creating a single file for
each release. It also allows to disable automatic indentation of
bullet point items, since not all projects necessarily use
bullet points for release notes. In single file mode, the output
file is a format string of the `name`, `version`, and `project_date`
and the file is fully replaced (`start_line` is not supported).

`name`, `version`, and `project_date` are now available in the main
template, allowing to render the title there. To make this useful,
the `title_format` should be set to `false`.

---

Will create a newsfragment soon. Is this something you would like? Basically, I would like to be able to create most of our current numpy style release notes automatically (for example: https://github.com/numpy/numpy/blob/master/doc/release/1.17.0-notes.rst).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -125,6 +125,7 @@ Towncrier has the following global options, which can be specified in the toml f
     issue_format = "format string for {issue} (issue is the first part of fragment name)"
     underlines: "=-~"
     wrap = false  # Wrap text to 79 characters
+    all_bullets = true  # make all fragments bullet points
 ```
 If a single file is used, the content of this file are overwritten each time.
 
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -117,7 +117,7 @@ def indent(text, prefix):
 # Takes the output from find_fragments above. Probably it would be useful to
 # add an example output here. Next time someone digs deep enough to figure it
 # out, please do so...
-def split_fragments(fragments, definitions):
+def split_fragments(fragments, definitions, all_bullets=True):
 
     output = OrderedDict()
 
@@ -126,7 +126,14 @@ def split_fragments(fragments, definitions):
 
         for (ticket, category, counter), content in section_fragments.items():
 
-            content = indent(content.strip(), u"  ")[2:]
+            if all_bullets:
+                # By default all fragmetns are append by "-" automatically,
+                # and need to be indented because of that.
+                # (otherwise, assume they are formatted correctly)
+                content = indent(content.strip(), u"  ")[2:]
+            else:
+                # Assume the text is formatted correctly
+                content = content.rstrip()
 
             if definitions[category]["showcontent"] is False:
                 content = u""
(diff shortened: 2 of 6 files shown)


seberg:
Sorry, also needs lots of test of course, although wouldn't mind a quick getting back on that.

seberg:
Just noticed that gh-147 does the title part already nicely. Probably nicer since `versioninfo.name` is more elegant. I think that one breaks backward compatibility if no `title_format` is given? Which I do not mind too much.

Anyway, happy to pull this out or do the opposite, but I would very much appreciate if we can try to keep things moving here @hawkowl? I expect other projects may have similar needs as NumPy.

hawkowl:
@seberg Hi! I finished off #147, would you be able to rebase the other changes?

seberg:
Thanks so much, I will do that today or tomorrow!

seberg:
OK, assuming the existing tests pass, this now adds two things (slight modification from before):

1. `single_file = false` option. If given, the output filename can be specified using the same formatting as the (old-style) `title_format`. In that case the file is always overwritten.
2. `no_bullets = true` option. If given, towncrier will not assume that all fragments should be rendered as bullet points (which means indenting them, plus the `-` already being in the template). Since I think it might (during the development) happen that we mix bullets and non bullets, I decided to sort fragments starting with bullet points to the front for the moment.

(this may still settle a bit, but should be good to look at.)

seberg:
Can I ping you on this? Being able to handle non-bullet points is pretty important to me. The single file per release I could probably handle differently. I do not mind taking a different approach here though or fixing things up a bit more.

seberg:
Oone more gentle ping @hawkowl, I do not mind working a bit more here though.

hawkowl:
@seberg I cherry picked out the single file changes into another branch, and I'll poke at some tests for them.

I don't quite understand the bullet point fragments, though, since making entries bulletpoints or not is the responsibility of the template. Could you just make an example output of what the option lets you do that editing the template doesn't? I can't quite figure it out from the code.

seberg:
Thanks. Making it bullet points or not should be the responsibility of the template, the issue is that the code (not the template) does indenting by two spaces (this could be done in the jinja2 template in a similar manner, but is not). In NumPy our snipplets normally have a subheading:
```
Deprecations
============

Deprecate `np.alen`
  -------------------
  `np.alen` was deprecated. Use `len` instead. (`gh-14181 <https://github.com/numpy/numpy/pull/14181>`__)
```
and the above is how it gets rendered by towncrier, not the spaces of the following lines which are not in the original) fragment.

Since we sometimes also use bullet points (Sections where everything is short), I thought I may want to allow both, and just sort the bullet points to the top, to be cleaned up manually before the final release (which is the reason for the sorting changes). I can work without the sorting change, although I doubt there is much harm in it.

hawkowl:
Thanks! I added some tests and merged it in here: https://github.com/hawkowl/towncrier/pull/162

seberg:
Thank you so much!

PR #160, opened 2019-08-12
ENH: Add `--no` option to not delete files

seberg:
Currently, if both no and yes are given, no will win unfortunately,
but that seems not a huge issue?


seberg:
I am not sure this is actually very important, it seemed nice for CI, but OTOH, in that context it is probably OK to just delete the files as well, since it restarts cleanly every time in any way.

ThiefMaster:
I would like to call towncrier during my sphinx build to include the latest changelog in the docs. But when writing docs I tend to build them locally to check that everything is fine - I won't want fragments to be deleted in that case.

Or alternatively, at least write anything that's not for the actual news file to stderr in `--draft` so the useful output can be piped elsewhere.

altendky:
Without suggesting that there's no functionality to add to towncrier, fyi there is https://pypi.org/project/sphinxcontrib-towncrier/.  I have not looked at it much myself yet.  I personally do run towncrier actually from my sphinx `conf.py` so that it will run in the Read the Docs build.  This is hacky and shared for context.

ThiefMaster:
Oh, that sphinx plugin looks really interesting! Gonna have a look at it :)

PR #161, opened 2019-08-31
Cherrypicked changes for single-file

hawkowl:
Cherry-picked from https://github.com/hawkowl/towncrier/pull/158

### DIFF
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -11,17 +11,19 @@ from __future__ import absolute_import, division
 import os
 
 
-def append_to_newsfile(directory, filename, start_line, top_line, content):
+def append_to_newsfile(directory, filename, start_line, top_line, content, single_file=True):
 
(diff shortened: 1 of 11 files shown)


src/towncrier/_writer.py
```diff
@@ -11,17 +11,19 @@
...
 
 
-def append_to_newsfile(directory, filename, start_line, top_line, content):
+def append_to_newsfile(directory, filename, start_line, top_line, content, single_file=True):
 
     news_file = os.path.join(directory, filename)
 
-    if not os.path.exists(news_file):
-        existing_content = u""
+    if single_file:
```
seberg: I think my intention here was the opposite, i.e. `single_file=False` is now the old behaviour (maybe the name is not very good). When a new file is created, I felt it is easier to simply overwrite. Not overwriting can be fine I guess, but right now the default behaviour is broken (will overwrite) I think. I can create a PR just flipping this one place.

PR #166, opened 2019-11-12
Allow for defining arbitrary project version and name in configuratio…

### LINKED ISSUE #165: Ability to use towncrier in non-python project
At the moment it's not possible, due to the fact, that towncrier tries to detect version if not passed in command line.

I'll create a pull request in a minute with ability to pass both project_name and project_version through config files, allowing the config based version to be maintaned by the tools like bumpversion as well, and just use the standard set of commands without the need to pass arbitrary arguments in command line.

fizyk:
…n - close #165

    This allows for the towncrier to be seamlessly used in non-python projects


tjanez:
We've successfully cherry-picked the changes in this PR to [Oasis Labs' towncrier fork](https://github.com/oasislabs/towncrier/) and [used it with the Oasis Core repo](https://github.com/oasislabs/oasis-core/pull/2496), which is a Go & Rust project.

src/towncrier/newsfragments/165.feature.rst
```diff
@@ -0,0 +1,2 @@
+Allow to define ``project_version`` and ``project_name`` in configuration.
+This allows to use towncrier seamlessly in the non-pythyon projects.
```
altendky, line 2:
```suggestion
Allow definition of ``project_version`` and ``project_name`` in the configuration file.
This allows use of towncrier seamlessly with non-Python projects.
```

altendky (changes requested):
Also needs a catch up.  Anyways, thanks for the work.  Sorry I went and duplicated it, though that made the review easy since they were so close.  `:]`  I can't merge anything right now but I'll try to get that fixed so we can get this integrated once we work through the minor tidbits.  My apologies it has taken so long to even get any feedback.

.gitignore
```diff
@@ -21,4 +21,5 @@ apidocs/
 .tox/
 .coverage.*
 .vscode
+.idea
```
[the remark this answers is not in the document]
fizyk: Would separate commit with this be enough, or you'd like to have separate pull request?

src/towncrier/_settings.py
```diff
@@ -144,6 +144,8 @@ def parse_toml(base_path, config):
         "single_file": single_file,
         "filename": config.get("filename", "NEWS.rst"),
         "directory": config.get("directory"),
+        "project_version": config.get("project_version"),
+        "project_name": config.get("project_name"),
```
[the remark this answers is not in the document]
fizyk: I think it's a matter of preference, and short version is as good as longer. One does not use library_version or library_name variables after all...

.gitignore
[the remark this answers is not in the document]
altendky: Don't worry about it here, just commenting for the future.  `:]`

src/towncrier/_settings.py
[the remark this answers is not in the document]
fizyk: Although comming back, I used the project_ prefixes since I was extending the use of already existing variables in code

src/towncrier/newsfragments/165.feature.rst
fizyk: Thanks, after all these years with English, I still have issues with articles :)

src/towncrier/_settings.py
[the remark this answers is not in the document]
altendky: Yeah, I saw the code used longer names and that makes sense to me.  I couldn't make up my mind so I thought about setuptools and checked Poetry and both of them use just `name` and `version`.  I don't think it ends up being ambiguous.  I don't feel strongly enough to push hard on this, but if you are comfortable with changing to the shorter, I think that's a touch more straightforward.  Take your pick and we'll move on.

altendky (commented):
I started looking through the readme and found a few other references to 'version' that seemed like they could use some updating to account for this new feature.

So I just realized that this also introduces the possibility of specifying the version and the name via both the command line options and the configuration file.  I think that other options, at least on the `build` command, can't do that.  I vaguely feel like perhaps the CLI options should go away, though there would be a deprecation period for that.  Let's just ignore that for now. 
 It can be handled in a separate PR if needed.  But, it would probably be good to add tests that confirm the priority of the command line over the configuration file.

With all this time since you submitted this, I'm glad you are available to get it worked through and finished up quickly.  `:]`  I'm sorry I keep finding tidbits.  I'm still on the newer side of review and appreciate your patience.

src/towncrier/_settings.py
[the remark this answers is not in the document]
fizyk: ok, changed to shorter

README.rst
```diff
@@ -123,6 +123,8 @@ Towncrier has the following global options, which can be specified in the toml f
     single_file = true  # if false, filename is formatted like `title_format`.
     filename = "NEWS.rst"
     directory = "directory/of/news/fragments"
+    version = "1.2.3"  # project version if maintained separately
+    name = "arbitrary project name"
```
fizyk, line 127: Although now I think that this particular one can be ambigous against a section name. Especially in the context of the title_format option

fizyk:
Also, I'd gladly squash commits and rework commit message if that'd be something you want.

altendky:
My personal preference is to never change public history, so no need to go squashing etc for my sake.

README.rst
altendky: I was going to make myself a bit more familiar with this.  But, `directory` is already present in both places and I'm not sure it would make sense to include section/type-specific information in the title.  At least not without some separation like iterating over the sections or selecting one by name.  I'll follow up after my exploration.

altendky (commented):
There are still a few references to 'version' in the readme that will mislead people into thinking there isn't a new option for specifying it.  If you would rather, I could go through that.  Just let me know.

README.rst
fizyk: And I just realised I haven't pushed that commit 🤦🏻

altendky: https://github.com/twisted/towncrier/blob/5c431028a3b699c74b162014e907272cbea8ac81/src/towncrier/build.py#L140-L143

So `name`, `version`, and `project_date` are hard coded and the only fields available to `title_format`.  I think this makes the short names match up better.  Does this satisfy your concern or did I misunderstand?

fizyk: 🤦🏻 yes they make more sense. Why did I thought of it as a section name, don't know. Must have been the late hour.

src/towncrier/newsfragments/165.feature.rst
```diff
@@ -0,0 +1,2 @@
+Allow definition of ``project_version`` and ``project_name`` in the configuration file.
```
altendky, line 1:
```suggestion
Allow definition of the project ``version`` and ``name`` in the configuration file.
```

altendky (changes requested):
Also see https://github.com/ClearcodeHQ/towncrier/pull/2 since GitHub didn't let me add a suggestion for a line not already in the diff context.

I think maybe that's it.

altendky:
@fizyk looks like you lost most of the changes in the readme.  https://github.com/twisted/towncrier/pull/166/commits/215a80aabf30f3cb4585546f5c343ca9368fb14a

fizyk:
> @fizyk looks like you lost most of the changes in the readme. [215a80a](https://github.com/twisted/towncrier/commit/215a80aabf30f3cb4585546f5c343ca9368fb14a)

🤦🏻‍♂️ fixed it

altendky:
@fizyk, sorry for the delay here.  Looks good to me.  I've got the CI fixed up properly now so I'm going to go ahead and update the branch here from master.  Hopefully that'll just work and I'll merge this but I wanted to give you a heads up so you don't go trying to push new work without pulling and get a surprise.

altendky (approved):

PR #167, opened 2019-11-21
Fix formatting in README

blueyed:
See https://github.com/blueyed/towncrier/tree/fix-readme#further-options for the fixed version.

Kept in two separate commits - the second just aligns the style.

I'm fine with squash-merging, or only picking the first commit.


hawkowl:
Thanks! <3

PR #168, opened 2019-11-24
Fix formatting of toml configs

sigmavirus24:
Previously they were formatted as markdown blocks but this is reStucturedText.

This should ensure they display properly in the README both on GitHub and PyPI.


hawkowl:
Fixed in #167

Issue #163, opened 2019-09-15
Towncrier incarnation with `python -mtowncrier` fails


seberg:
On current master, running towncrier with `python -mtowncrier` fails with `_main()` not being found. This probably is not an issue for me right now, since I can just call the script which works fine.

altendky:
It looks like this was fixed in #170 and released in 19.9.0rc1 (though we have no tag for that in this repo...).  @seberg, are you comfortable closing this?

seberg:
Thanks!

PR #170, opened 2019-12-02
Fix the use of -mtowncrier. #163

nedbat:
This makes the -m invocation work.

I added a test, but:
- I didn't know in what file you wanted it
- I didn't know if Trial had a way to manage the current directory

Let me know if I should change anything.

### DIFF
--- a/src/towncrier/test/test_project.py
+++ b/src/towncrier/test/test_project.py
@@ -43,7 +43,7 @@ class VersionFetchingTests(TestCase):
 class InvocationTests(TestCase):
     def test_dash_m(self):
         """
-        `python -m towncrier` works.
+        `python -m towncrier` invokes the main entrypoint.
         """
         temp = self.mktemp()
         new_dir = os.path.join(temp, "dashm")
--- a/src/towncrier/test/test_project.py
+++ b/src/towncrier/test/test_project.py
@@ -2,6 +2,7 @@
 # See LICENSE for details.
 
 import os
+from subprocess import check_output
 
 from twisted.trial.unittest import TestCase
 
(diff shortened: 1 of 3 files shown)


charris:
Also need this.

nedbat:
@charris I don't know what you are referring to.  Is there a larger conversation somewhere? Or did you mean to link to something that should be included in this PR?

charris:
I run several versions of Python for testing and development, plus upgrading the OS twice a year. In that circumstance it is helpful to have towncrier work as a module so as to keep track of which Python it is installed with.

src/towncrier/test/test_project.py
```diff
@@ -36,3 +37,24 @@ def test_tuple(self):
...
+        temp = self.mktemp()
+        new_dir = os.path.join(temp, "dashm")
+        os.makedirs(new_dir)
+        orig_dir = os.getcwd()
+        try:
+            os.chdir(new_dir)
+            with open("pyproject.toml", "w") as f:
+                f.write('[tool.towncrier]\n' 'directory = "news"\n')
+            os.makedirs("news")
+            out = check_output(["python", "-m", "towncrier", "--help"])
```
cdunklau: Probably should use `sys.executable` instead of `'python'` here, to avoid relying on PATH

cdunklau: ...hmm I didn't know this would show up as a "review", my intention was just to comment.

glyph, line 56:
```suggestion
            out = check_output([sys.executable, "-m", "towncrier", "--help"])
```

src/towncrier/test/test_project.py
```diff
@@ -2,6 +2,7 @@
 # See LICENSE for details.
 
 import os
```
glyph, line 4:
```suggestion
import os
import sys
```

src/towncrier/test/test_project.py
```diff
@@ -36,3 +37,24 @@ def test_tuple(self):
 
         version = get_version(temp, "mytestproja")
         self.assertEqual(version, "1.3.12")
+
+
+class InvocationTests(TestCase):
+    def test_dash_m(self):
+        """
+        `python -m towncrier` works.
```
glyph, line 45: "works" is one of those non-words called out by https://jml.io/pages/test-docstrings.html so let's say what working means

```suggestion
        `python -m towncrier` invokes the main entrypoint.
```

PR #173, opened 2020-01-07
Improve news fragment file name parsing

tjanez:
The new version uses category as the reference point to later infer the issue number and counter value.
As a consequence, it will return `(None, None, None)` if the base name doesn't contain a valid category.

The new parser allows using file names like `123.feature.1.ext` which are convenient when one wants to use an appropriate extension (e.g. `rst`, `md`) to enable syntax highlighting.


tjanez:
> The test cases look great to me. Thanks for this contribution!

No problem. I'm glad towncrier's development is ramping up again.

PR #174, opened 2020-01-09
Make check subcommand detect if only configured news file has changed

### LINKED ISSUE #152: Running check after building the changelog

I'm trying something like:

```bash
git checkout -b newbranch 
# Make a change.
# Commit the change.
# Add the newsfragment
# Bump the version number
bump2version patch
# Check for newsfragments
python -m towncrier.check
# Update changelog
towncrier --yes
python -m towncrier.check
# which fails.
```

tjanez:
This should enable the `check` subcommand to be used as a CI lint step and not fail when a pull request only modifies the configured news file (i.e. when the news file is being assembled for the next release).

Closes #152.

### DIFF
--- a/src/towncrier/check.py
+++ b/src/towncrier/check.py
@@ -34,7 +34,15 @@ def __main(comparewith, directory, config):
     try:
         files_changed = (
             _run(
-                ["git", "diff", "--name-only", comparewith + "..."], cwd=base_directory
+                ["git",
+                 "diff",
+                 "--name-only",
+                 # Only show files that were Added (A), Copied (C), Modified (M)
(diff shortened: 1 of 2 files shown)


altendky (commented):
@tjanez, would you have time to add some tests?  If not, let me know and maybe I can continue this effort.  Thanks for your contribution.

altendky:
If you get back to this before I manage to find some reviews, feel free to merge #296 back in here and we can continue with this PR and close mine.  I just didn't want to hijack your branch.  Thanks again.

adiroiban:
I think that this can be closed as the code was moved to #296 where tests were added.

PR #271, opened 2020-07-11
Allow specifying project name in config file

### LINKED ISSUE #270: Allow specifying project name in config file
The project name can presently be specified on the CLI or via an incremental version.  It would be nice to be able to set it in the config file.  In my particular case it is `QTrio` which when not specified takes the package `qtrio` and titles it to `Qtrio`.

altendky:
Fixes #270

### DIFF
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -539,3 +539,52 @@ class TestCli(TestCase):
             """
             ).lstrip(),
         )
+
+    def test_name_in_config(self):
+        """
+        When a name is specified in the config it is used.
+        """
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            with open("pyproject.toml", "w") as f:
+                f.write(
+                    '[tool.towncrier]\n'
+                    'name="SoMEThiNg"'
+                )
+            os.mkdir("newsfragments")
+            with open("newsfragments/123.feature", "w") as f:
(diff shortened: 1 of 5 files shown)


src/towncrier/test/test_build.py
```diff
@@ -539,3 +539,52 @@ def test_bullet_points_false(self):
...
+
+            self.assertEqual(0, result.exit_code, result.output)
+            with open("NEWS.rst", "r") as f:
+                output = f.read()
+
+        self.assertEqual(
+            output,
+            dedent(
+                """
+            SoMEThiNg 7.8.9 (01-01-2001)
```
Julian, line 578: This was likely already an issue previously, but perhaps we should have a test for name containing invalid `reST`?

Also relatedly (in other contexts) personally I always wrap the package name in at least double backticks, otherwise if you're running spellcheck and your package isn't an english word, you'll get spelling errors.

altendky: Sounds like this should be satisfied in another PR? Maybe...  Or is this a direct request for a change here?  The value in checking ahead of time would be early and direct complaints at the cost of maybe not checking in the right environment?

Julian: No both are likely fine as future work given that they already exist if someone somehow creates a package with a backtick in its name or something, but yeah I guess shouldn't block this PR.

altendky:
Closing as a duplicate of the pre-existing #166.

PR #275, opened 2020-10-07
Users can now write the content in the "create" method.

saroad2:
Now, you don't have to create the news fragment and then edit it. You can write the content dynamically!

### DIFF
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -16,12 +16,13 @@ from ._settings import load_config_from_options
 @click.command(name="create")
 @click.option("--dir", "directory", default=None)
 @click.option("--config", "config", default=None)
+@click.option("-i", "--interactive", is_flag=True, default=False)
 @click.argument("filename")
-def _main(directory, config, filename):
-    return __main(directory, config, filename)
+def _main(directory, config, filename, interactive):
+    return __main(directory, config, filename, interactive)
 
 
-def __main(directory, config, filename):
+def __main(directory, config, filename, interactive):
     """
     The main entry point.
     """
@@ -59,10 +60,29 @@ def __main(directory, config, filename):
     if os.path.exists(segment_file):
         raise click.ClickException("{} already exists".format(segment_file))
 
+    content = _get_news_content(interactive)
+
     with open(segment_file, "w") as f:
-        f.writelines(["Add your info here"])
+        f.write(content)
 
     click.echo("Created news fragment at {}".format(segment_file))
+
+
+def _get_news_content(interactive):
+    if not interactive:
+        return "Add your info here"
+    content = []
+    click.echo(
+        "Please write your news content. Press enter without content to finish"
+    )
+    count = 1
+    while True:
+        content_line = click.prompt("{}".format(count), default="", show_default=False)
+        if content_line == "":
+            break
+        content.append(content_line)
+        count += 1
+    return "\n".join(content)
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -61,7 +61,10 @@ def __main(ctx, directory, config, filename, edit):
     if os.path.exists(segment_file):
         raise click.ClickException("{} already exists".format(segment_file))
 
-    content = _get_news_content(edit)
+    if edit:
+        content = _get_news_content_from_user()
+    else:
+        content = "Add your info here"
 
     if content is None:
         click.echo("Abort creating news fragment.")
@@ -73,9 +76,7 @@ def __main(ctx, directory, config, filename, edit):
     click.echo("Created news fragment at {}".format(segment_file))
 
 
-def _get_news_content(edit, default_content="Add your info here"):
-    if not edit:
-        return default_content
+def _get_news_content_from_user():
     content = click.edit(
         "# Please write your news content. When finished, save the file.\n"
         "# In order to abort, exit without saving.\n"
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -62,6 +62,9 @@ def __main(directory, config, filename, interactive):
 
     content = _get_news_content(interactive)
 
+    if content is None:
+        click.echo("Abort creating news fragment.")
+        return
     with open(segment_file, "w") as f:
         f.write(content)
 
@@ -71,17 +74,15 @@ def __main(directory, config, filename, interactive):
 def _get_news_content(interactive):
     if not interactive:
         return "Add your info here"
-    content = []
-    click.echo(
-        "Please write your news content. Press enter without content to finish"
+    content = click.edit(
+        "# Please write your news content. When finished, save the file.\n"
+        "# In order to abort, exit without saving.\n"
+        "# Lines started with \"#\" are ignored."
     )
-    count = 1
-    while True:
-        content_line = click.prompt("{}".format(count), default="", show_default=False)
-        if content_line == "":
-            break
-        content.append(content_line)
-        count += 1
+    if content is None:
+        return None
+    content = content.split("\n")
+    content = [line for line in content if not line.startswith("#")]
     return "\n".join(content)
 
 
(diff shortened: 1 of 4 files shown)


saroad2:
Any chance someone take a look at it? It's the only thing preventing me from using *Towncrier* in my projects.
@nedbat ?

saroad2:
@Julian, any feedback on this?

src/towncrier/newsfragments/275.feature
```diff
@@ -0,0 +1 @@
+Users can now write the content in the "create" method.
```
altendky, line 1:
```suggestion
Users can now write the content in the ``create`` subcommand.
```

altendky (commented):
This is a great feature to add, thanks for the work on this.  I am a bit concerned about developing and maintaining our own editor though.  I wasn't familiar with this feature but as I searched around I ran across `click.edit()` which seems applicable.  Do you think it would work well here?

https://click.palletsprojects.com/en/7.x/utils/#launching-editors

saroad2:
Hey @altendky , `click.edit` was an excellent idea! thanks for that.

Now, it allowed us another functionality that wasn't supported before: if the user does not save the file created by the `edit` method, we abort without creating the new news fregmant.

If something else is needed to be done in order to merge this PR, let me know.

src/towncrier/test/test_create.py
```diff
@@ -3,7 +3,9 @@
 
 import os
 from textwrap import dedent
+
 from twisted.trial.unittest import TestCase
+import mock
```
altendky, line 8: I don't like using mock but Click does not yet provide a means for testing code that uses `click.edit()`.  We shouldn't wait up for that to be implemented so we'll just move forward with mock for now..

https://github.com/pallets/click/issues/1720

altendky (changes requested):
Thanks for the continued help with this and the quick turnaround.

src/towncrier/test/test_create.py
saroad2: Mocking is a very powerful tool that allows you to isolate the code components you want to test and assume that the dependant component (the one you are mocking) is working correctly. Using mocks can make your tests clearer, simpler, single-responsibility driven, and shorter.

I understand why people don't like mocks. When people do not use them correctly, they can cause a lot of problems. But, if you do know how to use them and keep using them **the right way**, it will change the way you test your code for life!

altendky: Is that better than the library you would otherwise mock providing an explicitly documented and supported feature such that you don't have to mock?

saroad2: Actually, in my opinion: no. I have 3 reasons for that:

1. *python* mocking library is based on **mock injection**. That means that when you use the python mock library you do not need to alter your code in order for it to be mockable (`mock.patch` is a good example for that).
2. Libraries should not "invent the wheel" and create their own mocking systems. Very smart people at the PSF have thought this through and created a general mocking library that can satisfy most needs.
3. When you rely heavily on a specific dependant library in your testing (in our case, the *click* library) you make it harder for future developers to replace this library if needed. One day you might want to use a different library for arguments parsing (maybe use the good old `argparse` 😉 ), and it will be very annoying to remove all the tests you've already written just because you decided to replace implementation. a good guideline for testing is **Test Behaviors, Not Implementations**. When you use the library's mocking process, you test implementation, not behavior.

As for everything in code, there is no "right" and "wrong", but it's a matter of approach. I hope that the reasons I mentioned above are well reasoned, but I'm sure that different people have different opinions about it.

altendky (approved):
There are some miscellaneous details I might like to tweak.  Would you prefer I raise them here or just follow up afterwards myself?  I'm fine with either but don't want to put you through every little tidbit unless you specifically want to.

Thanks again for this PR.

saroad2:
@altendky no problem mate, if you still think there additional things that need to be done related to this PR, feel free to add comments about them and I'll try to fix those ASAP.

One thing that I would like to ask is that after we do merge this PR, you would publish a new `towncrier` version so I could start using this new ability. As I said, this is the only thing preventing me from using `towncrier` ATM.

altendky:
@saroad2, I try not to make promises on work like this but my intent is to work through the open PRs and bugs and get a release out.  Who knows, maybe I give up part way through and do a release earlier.

src/towncrier/create.py
```diff
@@ -14,14 +14,16 @@
 
 
 @click.command(name="create")
+@click.pass_context
 @click.option("--dir", "directory", default=None)
 @click.option("--config", "config", default=None)
+@click.option("--edit/--no-edit", default=False)  # TODO: default should be true
```
altendky, line 20:
```suggestion
@click.option("--edit/--no-edit", default=False, help="Open an editor for writing the newsfragment content.")  # TODO: default should be true
```

src/towncrier/create.py
```diff
@@ -59,11 +61,32 @@ def __main(directory, config, filename):
...
     with open(segment_file, "w") as f:
-        f.writelines(["Add your info here"])
+        f.write(content)
 
     click.echo("Created news fragment at {}".format(segment_file))
 
 
+def _get_news_content(edit, default_content="Add your info here"):
+    if not edit:
+        return default_content
```
altendky, line 78: I think this would be simpler to read if the condition was moved outside the function call.  There's just not enough logic or code to need to be 'abstracted'.

src/towncrier/newsfragments/275.feature
```diff
@@ -0,0 +1 @@
+Users can now write the content in the ``create`` subcommand.
```
altendky, line 1: I know I already had you change it to this but I think it could be a bit more explicit in some way.  Maybe:

```suggestion
The new ``--edit`` option of the ``create`` subcommand launches an editor for entering the contents of the newsfragment.
```

src/towncrier/create.py
```diff
@@ -59,11 +61,32 @@ def __main(directory, config, filename):
     if os.path.exists(segment_file):
         raise click.ClickException("{} already exists".format(segment_file))
 
+    content = _get_news_content(edit)
+
+    if content is None:
+        click.echo("Abort creating news fragment.")
+        ctx.exit(1)
```
altendky, line 68: Do you know the details of the various ways to exit a click program and which is best?  (and why `:]`)  I wasn't even aware that `ctx.exit()` existed.

seberg:
@altendky just wanted to give some kudos!  It is very cool to see you picking up maintenance of towncrier. It has worked well for us at NumPy the past year (I think we may still be using the master branch, but that hardly matters).

altendky (changes requested):

src/towncrier/create.py
saroad2: `ctx.exit` means exit the program with code.
You can always use the python `return` directive if you want, but it means that the exit code would be 0.

Here, I chose to exit with code 1 because if the user did not write content, it means the operation has "failed", and the exit code should reflect that.

As for the use of `ctx.exit`, *click* has this **real bad habit** of wrapping built-in python methods with *click* functions (for example, wrapping the `print` method in `click.echo`). I don't like it at all, but I take it as a conversion when working with *click* (this is related to our mocking discussion). In our case, we could simply use `sys.exit` instead and it would do the same.

To sum up, `ctx.exit` is the best choice, considering that it is a convention.

altendky: I agree a non-zero exit code is sensible, my only question was how to do it.  I asked about this in the Click discord and the only answer I got was `sys.exit()`, but I think they hadn't heard of `ctx.exit()` either.  `:]`  Anyways, `click.echo()` does do more than print especially when dealing across Python versions and platforms.  There's lots of room for major hassle when trying to handle that at the lower levels.

altendky (approved):

altendky:
@saroad2, thanks again for both the initial submission and continuing to work through this with me.

saroad2:
Thank you so much @altendky ! I had a great time working on this.

I really like `towncrier`. I think it's such a great tool.
I wanted to integrate it into my repos (both open-source and work-related) for quite some time now. I think that once you publish a new version, I'll be able to do so!

PR #276, opened 2020-10-14
title_format overrides default top line

### LINKED ISSUE #180: Rendering issue with release candidate (19.9.0rc1)
When using the release candidate (19.9.0rc1), the title of the new entry in the Changelog is poorly formatted.

_This is using the same config as for 19.2.0._

There is duplication of the line:
```
"version" (2020-04-02)
====================
"package-name" "version" (2020-04-02)
=======================
```

altendky:
Fixes #180


src/towncrier/test/test_format.py
```diff
@@ -327,6 +331,38 @@ def test_line_wrapping_disabled(self):
...
+        """
+        Title format replaces default top line rendering.
+        """
+        self.maxDiff = None
+
+        fragments = {}
+
+        definitions = OrderedDict()
+
+        expected_output = u"""A custom top line
```
adiroiban, line 353: to make the code a bit easier to read I would prefer to have the `expect_outpupt" define just before the call to `self.assertEqual(output, expected_output)`

src/towncrier/test/test_format.py
```diff
@@ -327,6 +331,38 @@ def test_line_wrapping_disabled(self):
...
+
+        template = pkg_resources.resource_string(
+            "towncrier", "templates/default.rst"
+        ).decode("utf8")
+
+        fragments = split_fragments(fragments, definitions)
+        output = render_fragments(
+            template,
+            None,
+            "A custom top line",
```
adiroiban, line 365: also. to make all these render_fragments easier to read, it woud be nice to  call them with named arguments for all argumetns and not only 

```suggestion
        output = render_fragments(
            template=template,
            issue_format=None,
            top_line="A custom top line",
```

src/towncrier/test/test_format.py
```diff
@@ -327,6 +331,38 @@ def test_line_wrapping_disabled(self):
...
+        """
+        self.maxDiff = None
+
+        fragments = {}
+
+        definitions = OrderedDict()
+
+        expected_output = u"""A custom top line
+=================
+"""  # NOQA
```
adiroiban, line 355: minor comment. Why do we need no qa here?

src/towncrier/test/test_format.py
```diff
@@ -327,6 +331,38 @@ def test_line_wrapping_disabled(self):
...
+        self.assertEqual(output, expected_output)
+
+    def test_title_format_specified(self):
+        """
+        Title format replaces default top line rendering.
+        """
+        self.maxDiff = None
+
+        fragments = {}
+
```
adiroiban, line 350: I think that maxDiff ended up after a debugging session :)

```suggestion
```

src/towncrier/test/test_format.py
```diff
@@ -327,6 +331,38 @@ def test_line_wrapping_disabled(self):
...
+        expected_output = u"""A custom top line
+=================
+"""  # NOQA
+
+        template = pkg_resources.resource_string(
+            "towncrier", "templates/default.rst"
+        ).decode("utf8")
+
+        fragments = split_fragments(fragments, definitions)
```
adiroiban, line 361: Maybe use this ... in this way e have a hint that we don't care about fragments or definitions in this test and you don't need to look and see where and how they are created and how they are checked at the end.

```suggestion
        fragments = split_fragments(fragments={}, definitions=OrderedDict())
```

But for this tests it would be interesting to also have at least one definition and one fragment to increase the coverage and make it easy to compare the output for the case in which you run without a custom topline.

adiroiban (approved):
All good. Thanks. Only minor comments.

src/towncrier/test/test_format.py
altendky: I'm guessing it was the indentation?  Switched to `dedent()`.

PR #278, opened 2020-10-16
Strip down integer ticket numbers

### LINKED ISSUE #126: Leading zeros in issue-numbers are not stripped
If the issue-number in the newsfragment-file has leading zeros, these are kept when printing the issue-number: A file named `0020.feature.txt` will result into the entry `- blablabla (#0020)`.

I would expect leading zeros to be removed the example above to result into the entry `- blablabla (#20)`.

altendky:
Fixes #126 .


src/towncrier/test/test_builder.py
```diff
@@ -66,3 +66,15 @@ def test_dots_in_ticket_name_and_counter(self):
             parse_newfragment_basename("baz.1.2.feature.3", ["feature"]),
             ("2", "feature", 3),
         )
+
+    def test_strip(self):
+        self.assertEqual(
```
adiroiban, line 71: Minor comment

would be nice to have a docstring explaing the scope/purpose/expectation of this test :)

This can serve as the developer documentation for this feature.

altendky: https://github.com/twisted/towncrier/pull/278/files#diff-b7254cf4717de60e5d2e57233db4f5cbb47871439b71f2280bed581ff9fcf851R71-R73

```python
        """Leading spaces and subsequent leading zeros are stripped
        when parsing newsfragment names into ticket numbers etc.
        """
```

adiroiban (approved):
Great work. Thanks!

Issue #279, opened 2020-11-25
Get off Travis


altendky:
My perspective is that Travis is pretty much done.  I like GitHub Actions (GHA).  Twisted uses GHA, albeit along with others.  PR incoming.

https://www.traviscistatus.com/#month
![image](https://user-images.githubusercontent.com/543719/100182726-4e20f480-2eab-11eb-8848-0bc02658f539.png)

altendky:
Note #283.

PR #280, opened 2020-11-25
Add GitHub Actions workflow for testing and checking

altendky:
First step of #279.

Followup steps:
- [ ] Make the `All` job required under GitHub branch protections

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,5 @@
+name: CI
+
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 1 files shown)


altendky:
So the workflow isn't actually running here at the moment but you can see them run over at https://github.com/altendky/towncrier/pull/1.  We could merge based on that or I think we could get it to run here either by adding a minimal one to `trunk` first (I can submit another PR and we can merge it) or perhaps by pushing this branch to the `twisted/towncrier` repository instead of my `altendky/towncrier` repository.  There are three options, let me know which you want to go with.

altendky:
Closed in favor of #282.  That one is in the repo and so the actions will run.  Though apparently pushing there triggered the actions here...  but...  whatever.  We'll work that one instead.

PR #286, opened 2020-12-03
Correct check tasks in CI to run the proper tox env

altendky:


.github/workflows/ci.yml
```diff
@@ -84,7 +84,7 @@ jobs:
       run: python -m pip install --upgrade pip tox
 
     - name: Check
-      run: tox -c tox.ini -e ${{ matrix.python.tox }}-${{ matrix.task.tox }}
+      run: tox -c tox.ini -e ${{ matrix.task.tox }}
```
adiroiban, line 87: do we  need to specify the tox  config file? Looks like we are using the default one.

altendky:
@saroad2, would you be interested in doing some reviews on some of my PRs to keep these moving?  If so, I can mention in them as they are ready.

adiroiban (approved):
Looks good. Thanks. Ready to merge.

I would still prefer to have flake + towncrier checks in single job.

I expect that in the future we will have a pre-release / publish job that will check that we can build a source package + wheel at any time ... and automatically publish the package when a tag is created. Similar to what we have for pydoctor / ldaptor and twisted itself.

altendky:
Funny you talk about auto-publishing off a tag seeing as I was able to do that in my projects up until I used towncrier and now i have to go and create a branch and so on and so forth to prep a release.  `:]`  Anyways, I'll add a PR to merge the checks together.  Though, I'm still not clear the significant upside once proper Python/OS coverage is in.  I suppose as long as each check is in a different step then you at least won't have to go scanning for multiple separate checks within a single step that may have failed.

Thanks.

.github/workflows/ci.yml
altendky: Doesn't seem like we do.  That's left from Travis I assume as I've never done that before.

adiroiban:
> Funny you talk about auto-publishing off

:) ... well, the idea is that we only have auto-pypi publish and not auto-release 

You manually create a branch, prepare the release, ask for a review ... and once the release branch is merged you can create the tag ...and then package will be on pypi... and with RTD you also have the documentation published.

You can keep the separate checks if you prefer... it was just my feedback.... minor comment.

altendky:
Oh, I agree with the direction.  Just a little irony in my case.  I did start to explore making towncrier offer either a no-commit-for-release or post-release-cleanup-commit workflow.

And yeah, I'm not excited about working on the docs but that's one of the things I imagine myself getting to.  Maybe.

adiroiban:
Well, towncrier is small library (and this is a good thing)  . H think that the current documentation in the README is enough.

altendky:
Maybe it has evolved sufficiently since the docs PR three years ago.  Haven't gotten to actually looking at it yet.  `:]`

saroad2:
> @saroad2, would you be interested in doing some reviews on some of my PRs to keep these moving? If so, I can mention in them as they are ready.

Feel free to mention me whenever you want a review :)

PR #287, opened 2020-12-03
Add Windows to CI (plus associated fixes)

altendky:


.github/workflows/ci.yml
```diff
@@ -46,6 +48,12 @@ jobs:
         task:
           - name: Test
             tox: tests
+        exclude:
+          # Twisted and thus trial don't work on Windows with CPython 3.9
```
adiroiban, line 52: Maybe link to a Trac ticket for Twisted... I am sure this will be fixed ... or maybe is already fixed and just waiting for a new release.

But maybe it is worth migrating towcrier tests to stdlib unit test. In this way we will not depend on twisted for running the tests.

adiroiban (approved):
Looks good. Thanks!

.github/workflows/ci.yml
exarkun: Trial really has a bunch of nice facilities compared to stdlib unittest.  There are other options, of course, like pytest and testtools, but you're going to run into exactly the same issues with anything that's not part of the stdlib itself.

As far as Twisted's 3.9 release goes, there's https://twistedmatrix.com/trac/ticket/10027 at least

src/towncrier/test/test_check.py
```diff
@@ -22,6 +22,9 @@ def create_project(pyproject_path):
         f.write("Adds levitation")
 
     call(["git", "init"])
+    # The default git branch name may not be master, let's make sure we
+    # are consistent here.
+    call(["git", "checkout", "-b", "master"])
```
exarkun, line 27: Do we have to pick master here?  It wouldn't surprise me if the answer is yes because other things still assume/use master... But if it works to use some other name like "main" then maybe that would be a better choice.

tox.ini
```diff
@@ -33,9 +33,11 @@ deps =
 commands =
    python -V
    coverage --version
-   {envbindir}/trial --version
+   python -m twisted.trial --version
    coverage erase
-   coverage run -p {envbindir}/trial {posargs:towncrier}
+   # coverage fails to run trial on Windows since it is trial.exe
+   # so just use --module.
+   coverage run -p --module twisted.trial {posargs:towncrier}
```
exarkun, line 40: May not make any difference but this could also be `python -m coverage ...`.  I expect the right command is on PATH thanks to tox ... but a few lines above it looks like you thought `python -m twisted.trial ...` was a good idea, so maybe it's a good idea for coverage too?

exarkun (approved):

src/towncrier/test/test_check.py
altendky: I can switch it over.  I think there was one test that explicitly passed master to towncrier as the reference branch.  I could also just make the test check the primary branch before creating the 'PR' branch and then specify no name anywhere.  I'll do one of the two unless you make a specific request.

tox.ini
altendky: The `-m twisted.trial` was for consistency with this line which switched as explained in the comment.  But sure, could do it for everything.  I'll look, or maybe another PR to switch everything if there are several unrelated places where `-m` would make sense.

src/towncrier/test/test_check.py
altendky: So I just removed this all together for now.  This wasn't actually a Windows related issue and only showed up because I happened to install an updated Git in my Windows VM and allowed it to default to `main` instead of `master`.  So, it's a real issue in regards to an assumption made by the tests about external configuration, but not a Windows issue nor one that CI is presently forcing us to fix.

https://github.com/twisted/towncrier/issues/294

tox.ini
altendky: https://tox.readthedocs.io/en/latest/config.html#conf-allowlist_externals

Hmm, so tox only warns about outside access and doesn't, right there anyways, list an option to make that an error.  Anyways, there are several other places this would apply in general so I won't change them here.  I did update the comment to better represent coverage's perspective on the run subcommand.

.github/workflows/ci.yml
adiroiban: True. Thanks for the link.

PR #289, opened 2020-12-03
Test against PyPy 2 and 3

altendky:
Draft for:
- [x] Being combined in compound PR


altendky:
Apparently squash and merge doesn't close included PRs.  So here's a closure based on #291 being merged.

PR #290, opened 2020-12-04
Add macOS to CI

altendky:
Draft for:
- [ ] Being combined in compound PR


altendky:
Apparently squash and merge doesn't close included PRs.  So here's a closure based on #291 being merged.

PR #291, opened 2020-12-05
Expand CI

altendky:
Combines #288, #289, #290, and a little more.

Draft for:
- [x] Empty newsfragment
- [x] Self review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,9 +74,9 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - name: CPython 3.6
-            tox: py36
-            action: 3.6
+          - name: CPython 3.8
+            tox: py38
+            action: 3.8
         task:
           - name: Flake8
             tox: flake8
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,8 @@ jobs:
       fail-fast: false
       matrix:
         python:
+          # Using second most recent minor release for whatever little
+          # increase in stability over using the latest minor.
           - name: CPython 3.8
             tox: py38
             action: 3.8


altendky:
Adding Windows is still separate in #287 since it involved fixing both tests and code as well.

.github/workflows/ci.yml
```diff
@@ -57,9 +74,9 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - name: CPython 3.6
-            tox: py36
-            action: 3.6
+          - name: CPython 3.8
+            tox: py38
+            action: 3.8
```
exarkun, line 79: Any particular reason for this to move to 3.8 instead of 3.9?  It seems reasonable to bump it up, since CPython 3.6 is nearing the end of its supported life.  All else equal, I guess 3.9 would be a better new target since it will be supported for longer ... but maybe there are other concerns?

exarkun (approved):
Cool.  Looks pretty good.  I checked and I see PyPy 2 and PyPy 3 CI jobs are running the tests and seem to behave well (output looks a lot like the CPython output).  I also see the CPy 3.8 and CPy 3.9 jobs look good as do the macOS jobs.  Though a couple of the macOS jobs seem unusually slow - the Py 2.7 and PyPy 3 jobs.  Maybe that's random noise from the CI system.  I do see that at least a couple dependencies are having wheels built for them though.  If CI run time is an issue for towncrier then pursuing wheels from upstream for those might be worthwhile.

Apart from all that, just one question inline.  Address to your satisfaction and then merge.

.github/workflows/ci.yml
altendky: I figured that a little extra stability isn't a bad thing on the checks.  They are there to work, not exercise against latest.  And sure, 3.9 isn't all that new, but it doesn't have a patch release yet and my brain works a bit better without nuance.  Just use the second-most-recent minor and done.  When adding 3.10 to the test matrix we could bump this to 3.9.  (Though sure, I don't think I actually apply this explicitly anywhere else.  Maybe next time I will.  `:]`)

I guess I could afford a brief comment as to the intent.

altendky, line 79:
```suggestion
          # Using second most recent minor release for whatever little
          # increase in stability over using the latest minor.
          - name: CPython 3.8
            tox: py38
            action: 3.8
```

PR #292, opened 2020-12-06
One check job

altendky:
Draft for:
- [x] Revisit after #291 and #287 expand the CI matrix coverage
- [x] Adjust so that all check outputs are present but any single one will result in failure
  - Using `if: always()` like this makes all three run even when a previous step has failed.  This isn't great but I don't have a better working solution and running a few extra steps that will likely fail as well when there's an oddball failure in setup steps doesn't seem super problematic.
  - Here's an example failing build.
    - https://github.com/twisted/towncrier/pull/292/checks?check_run_id=1508471988
    - image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,23 +50,16 @@ jobs:
   check:
     name: Check - ${{ matrix.python.name }}
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        python:
-          - name: CPython 3.6
-            tox: py36
-            action: 3.6
 
     steps:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
 
-    - name: Set up ${{ matrix.python.name }}
+    - name: Set up CPython 3.6
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python.action }}
+        python-version: 3.6
 
     - uses: twisted/python-info-action@v1.0.1
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
-    name: ${{ matrix.task.name}} - ${{ matrix.python.name }}
+    name: Test - ${{ matrix.python.name }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -26,9 +26,6 @@ jobs:
           - name: CPython 3.7
             tox: py37
             action: 3.7
-        task:
-          - name: Test
-            tox: tests
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


altendky:
Hmm...  but `if: always()` will always run those steps even if maybe the checkout step fails.  That's not the intent here.

altendky:
@adiroiban, since we've now got 27 'jobs' thus requiring scrolling (at least on the main PR page) regardless, I don't personally see this as an improvement.  That said, the check job does have separate steps for each check such that you don't have to dig through a single log to get potentially multiple sets of errors, so it doesn't make things a hassle either.  If this wasn't what you had expressed interest in, let me know.  If you like it, go ahead and merge.

.github/workflows/ci.yml
```diff
@@ -75,45 +72,40 @@ jobs:
 
     - name: Codecov
       run: |
-        codecov -n "GitHub Actions - ${{ matrix.task.name}} - ${{ matrix.os.name }} ${{ matrix.python.name }}"
+        codecov -n "GitHub Actions - Test - ${{ matrix.os.name }} ${{ matrix.python.name }}"
 
   check:
-    name: ${{ matrix.task.name}} - ${{ matrix.python.name }}
+    name: Check
```
adiroiban, line 78: maybe rename `Static checks` ? 'Check is a  bit generic'
```suggestion
    name: Static checks
```

.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
-    name: ${{ matrix.task.name}} - ${{ matrix.os.name }} ${{ matrix.python.name }}
+    name: Test - ${{ matrix.os.name }} ${{ matrix.python.name }}
```
adiroiban, line 11:
```suggestion
    name: ${{ matrix.python.name }} ${{ matrix.os.name }} tests
```

adiroiban (approved):
Looks good. Thanks. Much appreciated.

What I want to get is have the important info at the start of the name...

For example we have this... and then is had to see which is which.

image

.github/workflows/ci.yml
altendky: https://github.com/pytest-dev/pytest-twisted/pull/108 Is my 'plan' to address the OS name (🐧, 🪟, 🍎).  I guess the empty box I get instead of the window character is actually kind of like an ultra simplistic window so maybe I should just merge that.  Python names could be more like the ABI tags (https://www.python.org/dev/peps/pep-0425/#abi-tag).  So, cp27 and pp3 etc.

Should I try a setup like this in an effort to allow all info to be visible even in the short little space provided in the popup you shared?  Rather than shuffling around what you get to see.

.github/workflows/ci.yml
altendky: Seems counter to being short.  `:]`  Is there really not a single word that has the desired meaning?  It seems it only needs to be distinct from 'test'.

altendky (commented):
Maybe this is part of the reason for multiple workflows.  Then we don't have the pointless `CI` and can put `Test` or `Check` there instead to save a little space.  But, I generally prefer a single file.

altendky:
Also, with all these jobs forcing vertical scrolling in all cases, is it still useful to have the newsfragment/lint/manifest hidden a layer down instead of being immediately visibly at the top level which is which?  This is the bit I don't understand the benefit of.

adiroiban:
> Should I try a setup like this in an effort to allow all info to be visible even in the short little space provided in the popup you shared? Rather than shuffling around what you get to see.

Nope.  Just merge is as it is :) Thanks


-------


>  is it still useful to have the newsfragment/lint/manifest hidden a layer down instead of being immediately visibly at the top level which is which? This is the bit I don't understand the benefit of.

As you prefer :) It was a minor comment.

PR #296, opened 2020-12-12
Make check subcommand detect if only configured news file has changed

### LINKED ISSUE #152: Running check after building the changelog

### LINKED ISSUE #246: Chicken-and-egg problem with `--check` option
If you add `towncrier --check` as a CI check, it is possible to verify that every PR that is merged has a matching release note. This is useful for ensuring final release notes are complete.

However, when you cut a release, a user will submit a PR that incorporates the final compiled release notes, along with a version number bump and any other release procedures. 

This "release PR" then fails the towncrier CI check, as it doesn't have a news snippet. 

You can make the PR pass CI by adding a 'misc' release note for  the release PR. However this means the release notes for the next version include an entry about cutting the *previous* release. 

There should be a way to avoid this "chicken and egg" problem without irrelevant dummy entries in changelog history.

altendky:
Closes #152
Closes #246

Draft for:
- [ ] Consideration of explicit specification of check mode (release vs feature/fix/etc)
- [ ] Consideration of complete checks to be done (if release, verify no newsfragments exist)


altendky:
Just building on @tjanez's work in #174.  Maybe just by adding the test...

src/towncrier/check.py
```diff
@@ -55,6 +64,10 @@ def __main(comparewith, directory, config):
         click.echo("On trunk, or no diffs, so no newsfragment required.")
         sys.exit(0)
 
+    if files_changed == config["filename"]:
+        click.echo("Only the configured news file has changed.")
```
adiroiban, line 68: I would say add uppercase when skipped, to help detect accidental configuration errors.

```suggestion
        click.echo("Checks SKIPPED. Only news file changes detected that looks like a release branch.")
```

src/towncrier/test/test_check.py
```diff
@@ -13,7 +14,11 @@
 
 def create_project(pyproject_path):
     with open(pyproject_path, "w") as f:
-        f.write("[tool.towncrier]\n" 'package = "foo"\n')
+        f.write(dedent("""\
```
adiroiban, line 17: I think that we should disable the linter warning for this type of indentation... and maybe just use black :)

I feel that dedent here has no technical merit, and is just added to please a tool :)

src/towncrier/test/test_check.py
```diff
@@ -148,3 +156,24 @@ def test_none_stdout_encoding_works(self):
 
         self.assertEqual(0, proc.returncode)
         self.assertEqual(b"", stderr)
+
+    def test_newsfile_modified(self):
+        """
+        A modified newsfile satisfies the check.
```
adiroiban, line 162:
```suggestion
        When only the newsfile is modified on a branch --check will not complain about missing newsfile fragmants.
```

adiroiban (commented):
I am not sure if this will work.

I don't know what is the "configured news" file.

I guess that it will work for a very limited use case.

----------

For a release PR, I am also updating a file containing the version (setup.py / setup.cfg or /_version.py) and sometimes 

And basically, for a release branch I am doing the reverse check.
The news fragments check will fail if they is any newsfragment at all.

My suggestion is to rename to "Make --check work for a release branch." or "Detect when running on a release branch."

Then for a release branch it will not fail if there is no news fragment for the branch, but will fail if there are any newsfragments at all.

I see 2 tasks:
1. detect (heuristic) if towncrier runs for a release branch
2. run different checks for a release branch

One idea is to add a separate `--check-release` argument that will run point 2.
In this way, for point 1 users will add their own external logic to detect a release branch.

----------

I am detecting the release branch based on the branch name (*-release-) , it works for my buildbot CI. Not sure if it works with GitHub.

I am using this helper https://github.com/chevah/brink/blob/master/brink/pavement_commons.py#L129

My branches all start with the ticket id, in this way is easier to detect so I can also check that the file has the right news fragment...and not just any news fragment :)

altendky (commented):
I don't understand the general concerns about ~functionality and~ what the configured newsfile is.  Isn't the configured newsfile the one that would be written to if using `build` instead of `check`?  Perhaps this is just misuse of terminology where newsfile is being improperly used where changelog should be instead.  In my head news fragments were used to build a news file (often called a changelog).  But, I see in the readme `These fragments are also commonly called "topfiles" or "newsfiles" in Twisted parlance.`  The readme doesn't seem to use any distinct name for the output file beyond the `filename` configuration option name.  This is probably a thing to address.

Maybe that clarification helps explain the concept?  The intent was that `check` would check for either a modification to the output file (changelog?) and no modified newsfragments, or one or more added news fragment files.  I see that this will fail in the case that you otherwise edit the output file and also have a news fragment.  Also, it doesn't check for an actual lack of news fragments when doing 'the release check'.

I appreciate the interest in explicit external control.  I think something like a regex in the configuration file to choose based on branch name or such might be a reasonable option as well.  I'll take a step back and see what I can come up with.

As always, thanks for the consideration and suggestions.

adiroiban:
Note that I was not rejecting the PR :) Just adding a comment with feedback based on my usage of towncrier and my release procedure.

I call the resulting file "Release Notes" .I find that end users / customers understand it best (vs NEWS File (GNU) or Changelog File).
Also, I think that Release Notes is a better reminder that the text is designed for general public and is not a dev log of changes.

--------

I guess that we can also add the case when only the "newsfile" is modified in a branch and disable the checks in that case.
My comment, was that when only the "newsfile" is modified, we should check that no news fragment are left in the source code.

----------

Maybe start small by adding a special `--check-release` command line.
It will do the different checks require for a release branch.

altendky:
> Note that I was not rejecting the PR :) Just adding a comment with feedback based on my usage of towncrier and my release procedure.

Sometimes I feel that the comments are sensibly addressed in a follow up and that the existing PR is a good stepping stone.  Per my thoughts after reading your comments, this PR doesn't seem to get either the branch detection bit right nor implement the proper checks for each case.  I think these points should be addressed instead of merging this (roughly) as is.  I think the separation of implementing the checks and the branch detection make sense.  This PR may just get closed and replaced by two to cover those separate pieces.  We'll see when I make it back to working on this one.

adiroiban:
I think that this is now fixed with #337

PR #176, opened 2020-03-09
Fix/revisit get_package

blueyed:
Try importing without changing `sys.path` first.

This makes a difference for pytest, where it would fail to
`from _pytest._version import version`, since `_pytest._version` does
not exist in the source tree (with non-editable installs), but only
the installed package.

It also improves the error reporting:

- prefix with "ERROR" for what might be redirected
- raise the `ImportError` directly (more information than when wrapped
  in `Exception`)

Ref: https://github.com/pytest-dev/pytest/issues/6831

PR #297, opened 2020-12-13
Fix/revisit get_package

altendky:


altendky:
Just building on @blueyed's work in #176.  Maybe just by adding the test...

src/towncrier/_project.py
```diff
@@ -16,26 +16,22 @@
...
-    sys.path.pop(0)
-
-    # Step 2: uhhhhhhh
-    # TBA
-
-    if not module:
-        raise Exception("Can't find your project :(")
+    except ImportError:
+        # Try the dumbest and simplest thing that could possibly work.
+        # Yes, that means importing it. Call the cops, I don't care.
```
adiroiban, line 23:
```suggestion
        # Package is not already available / installed.
        # Force importing it based on the source files.
```

src/towncrier/test/test_project.py
```diff
@@ -39,6 +39,46 @@ def test_tuple(self):
         version = get_version(temp, "mytestproja")
         self.assertEqual(version, "1.3.12")
 
+    def test_import_fails(self):
+        """
+        The project cannot be found.
```
adiroiban, line 44:
```suggestion
        An exception is raised when getting the version failed due to missing Python package files.
```

src/towncrier/test/test_project.py
```diff
@@ -39,6 +39,46 @@ def test_tuple(self):
...
+        """
+        if sys.version_info >= (3, 6):
+            expected_exception = ModuleNotFoundError
+        else:
+            expected_exception = ImportError
+
+        with self.assertRaises(expected_exception):
+            get_version(".", "mytestproj_import_fails")
```
adiroiban, line 52:
```suggestion
            get_version(".", "projectname_without_any_files")
```

src/towncrier/test/test_project.py
```diff
@@ -39,6 +39,46 @@ def test_tuple(self):
...
+        An already installed package will be checked before cwd-found packages.
+        """
+        project_name = "mytestproj_already_installed_import"
+
+        temp = self.mktemp()
+        os.makedirs(temp)
+        os.makedirs(os.path.join(temp, project_name))
+
+        with open(os.path.join(temp, project_name, "__init__.py"), "w") as f:
+            f.write("__version__ = (1, 3, 12)")
```
adiroiban, line 65:
```suggestion
            f.write("__version__ = "version.on.disk")
```

but maybe we don't need at all.

src/towncrier/test/test_project.py
```diff
@@ -39,6 +39,46 @@ def test_tuple(self):
...
+        sys_path_temp = self.mktemp()
+        os.makedirs(sys_path_temp)
+        os.makedirs(os.path.join(sys_path_temp, project_name))
+
+        with open(os.path.join(sys_path_temp, project_name, "__init__.py"), "w") as f:
+            f.write("__version__ = (1, 3, 13)")
+
+        sys.path.insert(0, sys_path_temp)
+        try:
+            version = get_version(temp, project_name)
```
adiroiban, line 76: can we just pass here a path that doesn't exits to check that it is completely ignored by `get_version` ?

src/towncrier/test/test_project.py
```diff
@@ -39,6 +39,46 @@ def test_tuple(self):
...
+            f.write("__version__ = (1, 3, 12)")
+
+        sys_path_temp = self.mktemp()
+        os.makedirs(sys_path_temp)
+        os.makedirs(os.path.join(sys_path_temp, project_name))
+
+        with open(os.path.join(sys_path_temp, project_name, "__init__.py"), "w") as f:
+            f.write("__version__ = (1, 3, 13)")
+
+        sys.path.insert(0, sys_path_temp)
```
adiroiban, line 74: Using `addCleanup` should be more idiomatic for cleanup code.

```suggestion
        sys.path.insert(0, sys_path_temp)
        self.addCleanup(sys.path.pop, 0)
```

adiroiban (approved):
Looks good and I think that this is a big improvement. Only minor comments.

adiroiban (commented):
For `get_version` I think that we should have this logic

See if version is a tuple and join them with a dot.
For anything else, just call `str(version_object)` and strip it.

In this way it will work with incremental and any other version objects.

towncrier shouldn't care about comparing versions...

It can try to be smart and detect that you are not adding a version older than the latest one  in the file... but I think that is too much...and again it should be implemented with a naive compare

So I am +1 for remove the incremental dependency :) and use simple version strings for towncrier :)

I am using a fork of towncrier in which incremental is removed as at least 3 years ago, incremental was not playing nice with private PYPI repos due to a bug in setup tools

altendky:
Am I correct that your comments about `get_version()` are effectively a suggestion for a new ticket?  Just making sure I understand before I go implementing the other suggestions.

adiroiban:
> Am I correct that your comments about `get_version()` are effectively a suggestion for a new ticket?

Yes. Sorry for not being explicit. I was just venting by pain with Incremental :)

src/towncrier/test/test_project.py
```diff
@@ -39,6 +39,46 @@ def test_tuple(self):
...
+        try:
+            version = get_version(temp, project_name)
+        finally:
+            sys.path.pop(0)
```
altendky, line 78:
```suggestion
        sys.path.insert(0, sys_path_temp)
        self.addCleanup(sys.path.pop, 0)

        version = get_version(temp, project_name)
```

src/towncrier/test/test_project.py
altendky: Sounds like a maybe reasonable separate test case?  But it seems valid to test that when it \_does\_ exist it still acts as expected?

src/towncrier/test/test_project.py
altendky: I'm not clear what this would clarify.  `:]`

adiroiban: '1.3.12' vs '1.3.13` is a tiny difference :) .. .but maybe we don't even need to create these files ... and only have `sys_path_temp`

altendky: This is testing that `get_version()` get's the version from the proper place.  Unless we extracted the selecting of the package to be a separate activity (sure, that sounds maybe good) then don't we need to put versions into the packages so we can make sure the proper one was used?

'version on disk' doesn't seem to really distinguish between the two, but 'installed' vs. 'source' might?  Though seems reasonable to keep them as 'versions' instead of 'names'.  Perhaps `(1, 3, 12)` and `(2, 1, 5)` would make them usefully distinct so the difference is harder to overlook?

PR #298, opened 2020-12-13
Raise ConfigError for missing newsfragment directory

### LINKED ISSUE #85: newsfragments directory not created
When running `towncrier --draft` first time, it crashes with:
```
Loading template...
Finding news fragments...
…
  File "/tmp/myvenv/lib/python2.7/site-packages/towncrier/_builder.py", line 46, in find_fragments
    files = os.listdir(section_dir)
OSError: [Errno 2] No such file or directory: '…/PyInstaller/newsfragments'
```

towncrier should not crash with a traceback, but catch the exception and print a nice error message without traceback. Or it should create this directory.

altendky:
Fixes #85


src/towncrier/newsfragments/85.bugfix.rst
```diff
@@ -0,0 +1 @@
+A missing newsfragment directory is treated as if it is empty rather than raising an exception.
```
adiroiban, line 1:
```suggestion
A configuration error is triggered when the newsfragment files couldn't be discovered. 
```

adiroiban (changes requested):
Thanks for working on fixing this bug.

I am -1 on this.

I think that is correct to not have towcrier create directories.

But I think that the error should not be ignored.

I am running towncrier draft as part of by continuous testing.
I want to see an error. For example if I have a typo in the path.

So instead of ignoring, just raise a ConfigError with a message that the news fragment couldn't be retrieved. It might be a missing path, it might be an OS permission error.

Just show a message like: "Failed to list the news fragment files. OS_ERROR_HERE"

And people should know what is wrong from the OS ERROR message.

adiroiban:
My view now 
image

src/towncrier/test/test_build.py
```diff
@@ -12,13 +12,14 @@
...
 
-def setup_simple_project():
+def setup_simple_project(create_newsfragments_directory=True):
     with open("pyproject.toml", "w") as f:
         f.write("[tool.towncrier]\n" 'package = "foo"\n')
     os.mkdir("foo")
     with open("foo/__init__.py", "w") as f:
         f.write('__version__ = "1.2.3"\n')
-    os.mkdir("foo/newsfragments")
+    if create_newsfragments_directory:
```
adiroiban, line 21: Is it worth updating this setup method when the new functionality is only used by a single test?

Also, the test `test_no_newsfragments` is easier to read when ` os.mkdir("foo/newsfragments")` is called inside the test.

src/towncrier/test/test_build.py
```diff
@@ -82,6 +83,35 @@ def test_command(self):
...
+        """
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            setup_simple_project(create_newsfragments_directory=True)
+
+            result = runner.invoke(_main, ["--draft", "--date", "01-01-2001"])
+
+        # This should fail
+        self.assertEqual(result.exit_code, 0)
```
adiroiban, line 112: I am not sure what is going on here.

I guess that this is just a TODO due to some strange code in Click, and in the change we will have a non-zero exit code here.

```suggestion
        # For now, exit code is zero, but it should be non-zero.
        # See: LINK TO ISSUE
        self.assertEqual(result.exit_code, 0)
```

adiroiban (approved):
For the title of PR, this PR is ok.

I would prefer to also see a non-zero exit code implemented in this PR :)

But this can  be merged, assuming that we have a follow up issue for the non-zero exit code.
 
Thanks!

src/towncrier/test/test_build.py
altendky: Let's disregard my comment, I'll remove it.  Do you want the pre-existing behavior of creating output with 'no significant changes' as the text and a zero exit code?  Or, are we just confusing each other starting with my bad comment I left laying around.

altendky: Looks like the comment may have been a copy/paste error from the test after this.

altendky:
Kick CI to watch try to observe codecov responses a bit.

src/towncrier/test/test_build.py
```diff
@@ -82,6 +82,35 @@ def test_command(self):
...
+        An empty newsfragment directory acts as if there are no changes.
+        """
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            setup_simple_project()
+
+            result = runner.invoke(_main, ["--draft", "--date", "01-01-2001"])
+
+        self.assertEqual(result.exit_code, 0)
```
adiroiban, line 111: Very minor comment  to match the previous tests... and to have self.assert(EXPECTED, ACTUAL) as the order of arguments.

```suggestion
        self.assertEqual(0, result.exit_code)
```

adiroiban (approved):
Merge it :) Thanks!

Issue #164, opened 2019-11-03
Setting title_format = false does not work


vytas7:
Hello,
I would like to disable title generation as our file already has a title heading.

`README.rst` mentions this can be achieved by setting `title_format` to `false` in the `pyproject.toml` config:
```toml
    title_format = "{name} {version} ({project_date})"  # or false if template includes title
```

However, that doesn't quite work for me:
```
...

  File "<... my virtual env ...>/lib/python3.7/site-packages/towncrier/__init__.py", line 122, in __main
    top_line = config["title_format"].format(
AttributeError: 'bool' object has no attribute 'format'
```

Setting `title_format` to an empty string seems to work (?) as a workaround, although a lot of empty space is emitted suggesting it is sort of using that string to format a zero-length heading.

altendky:
@vytas7, I think this has been fixed though there seemed to be no tests for it.  I added a couple in #299 (and snuck in one less newline which seemed more consistent).  Would you be willing to take a look at that and/or try out the `master` branch to see if your issue is still present?

vytas7:
Thanks for looking into this @altendky !
Checking master and your feature branch.

vytas7:
`towncrier==19.9.0rc1` already works as expected (`false` is honoured).
Now it's just a matter of getting a stable release out :slightly_smiling_face: 

(FWIW, using the branch from your PR did not render any difference in our case.)

PR #299, opened 2020-12-14
Add title_format tests and drop a line

altendky:


src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
             """
             ).lstrip(),
         )
+
+    def test_title_format_non_empty(self):
```
adiroiban, line 675:
```suggestion
    def test_title_format_custom(self):
```

adiroiban (approved):
Looks good.

I am not sure what is the scope of this PR and why we need it.

I would argue that we don't need to support `false` so don't add extra test for it... but at the same time we don't need to add extra logic for it.

I would say that only having`test_title_format_non_empty` is enough for this PR.

Thanks!

adiroiban (changes requested):
I see this in the docs

`title_format = "{name} {version} ({project_date})"  # or false if template includes title`

And now I am confused...

I think that we should change README as part of this PR to make it clear how title_format should be used and what is the purpose of setting it to false.

altendky:
Explanation was certainly lacking.  This followed from https://github.com/twisted/towncrier/issues/164 where it was reported that `title_format = false` in the configuration file was not working properly.  The code had been fixed by 19.9.0rc1 but I didn't see any tests so I figured it was worth adding a test in response to the ticket rather than just closing the ticket.

My understanding is that if you specify a template which includes the title then you should specify `title_format = false` to avoid doubling up on titles (your template plus the default title format).  So, switch this from misc to a doc news fragment, provide an explanation in the readme, and keep both tests?

After we agree on that I'll handle the rest of the details.

adiroiban:
> My understanding is that if you specify a template which includes the title then you should specify `title_format = false` to avoid doubling up on titles (your template plus the default title format). So, switch this from misc to a doc news fragment, provide an explanation in the readme, and keep both tests?

OK. I guess that we should have a separate section to explain each option... and not show them as an example with a brief comment.

So, my understanding is:

```
title_format = "{name} {version} ({project_date})"

Use this option to define format the title for each release.
Available placeholders are:
* name - name of the project
* version - current version of the project
* project_date - date when the version was/is released

Leave it out or set it to empty string to use the default towncrier format.
Set it to `false`  to not have townrier generate a tiltte. For the case in which your custom template is defining its own title.
```

And yes. the docstring for each test could be verbose and explain each tests case and what functionality is tested.

adiroiban (approved):
With better documentation this can be merged.

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
...
+            )
+
+        expected_output = dedent("""\
+            Loading template...
+            Finding news fragments...
+            Rendering news fragments...
+            Draft only -- nothing has been written.
+            What is seen below is what would be written.
+
+            abc FooBarBaz 7.8.9 (01-01-2001)
```
altendky, line 716:
```suggestion
            [{01-01-2001}] CUSTOM RELEASE for FooBarBaz version 7.8.9
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
...
+            - Extends levitation (#124)
+
+        """)
+
+        self.assertEqual(0, result.exit_code)
+        self.assertEqual(expected_output, result.output)
+
+    def test_title_format_false(self):
+        """
+        Setting the title format to false disables the explicit title.
```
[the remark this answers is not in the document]
altendky: I think this was addressed with the explanation of a `false` value.

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
...
+            )
+
+        expected_output = dedent("""\
+            Loading template...
+            Finding news fragments...
+            Rendering news fragments...
+            Draft only -- nothing has been written.
+            What is seen below is what would be written.
+
+            FooBarBaz 7.8.9 (01-01-2001)
```
altendky, line 771:
```suggestion
            FooBarBaz 7.8.9 (20-01-2001)
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
...
+
+            result = runner.invoke(
+                _main,
+                [
+                    "--name",
+                    "FooBarBaz",
+                    "--version",
+                    "7.8.9",
+                    "--date",
+                    "01-01-2001",
```
altendky, line 704:
```suggestion
                    "20-01-2001",
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
...
+            with open("pyproject.toml", "w") as f:
+                f.write(dedent("""\
+                    [tool.towncrier]
+                    package = "foo"
+                    title_format = false
+                """))
+            os.mkdir("foo")
+            os.mkdir("foo/newsfragments")
+            with open("foo/newsfragments/123.feature", "w") as f:
+                f.write("Adds levitation")
```
altendky, line 746:
```suggestion
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,113 @@ def test_bullet_points_false(self):
...
+            Draft only -- nothing has been written.
+            What is seen below is what would be written.
+
+            FooBarBaz 7.8.9 (01-01-2001)
+            ============================
+
+            Features
+            --------
+
+            - Adds levitation (#123)
```
altendky, line 777:
```suggestion
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,111 @@ def test_bullet_points_false(self):
...
+            )
+
+        expected_output = dedent("""\
+            Loading template...
+            Finding news fragments...
+            Rendering news fragments...
+            Draft only -- nothing has been written.
+            What is seen below is what would be written.
+
+            FooBarBaz 7.8.9 (20-01-2001)
```
vytas7, line 770: I missed this when reporting the fix worked for me... This is well incorrect, the title should not appear here as per `title_format = false` above?

PR #300, opened 2020-12-14
Fix handling of sys.stdout.encoding being None

### LINKED ISSUE #175: Failure to run in GitHub Actions
I'm trying to run  `python -m towncrier.check --pyproject towncrier.pyproject.toml` from GitHub Actions steps, and towncrier fails in this manner:

```
  File "/opt/hostedtoolcache/Python/2.7.17/x64/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/opt/hostedtoolcache/Python/2.7.17/x64/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/towncrier/check.py", line 98, in <module>
    _main()
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 764, in __call__
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 956, in invoke
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 555, in invoke
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/towncrier/check.py", line 31, in _main
    return __main(compare_with, directory, pyproject)
  File "/home/runner/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/towncrier/check.py", line 44, in __main
    .decode(getattr(sys.stdout, "encoding", "utf8"))
TypeError: decode() argument 1 must be string, not None
```

That `sys.stdout.encoding` is set to `None` seems to be a GitHub Actions quirk, because this doesn't happen in TravisCI or CircleCI.

It happens on all three GA setups (Windows, macOS, Ubuntu) I tried.

More here: https://github.com/sajith/tahoe-lafs/runs/462170050#step:9:22.

altendky:
Fixes #175.


src/towncrier/check.py
```diff
@@ -31,12 +31,17 @@ def __main(comparewith, directory, config):
 
     base_directory, config = load_config_from_options(directory, config)
 
+    # handled like this in case the attribute is present but None
+    encoding = getattr(sys.stdout, "encoding", None)
+    if encoding is None:
+        encoding = "utf8"
+
```
adiroiban, line 38: The comment is confusing...  I use getattr when the attribute might not be present.

Maybe

```suggestion
    encoding = sys.stdout.encoding
    if not encoding:
        # On Python 2.7 or piped standard output.
        encoding = "utf-8"

```

src/towncrier/test/test_check.py
```diff
@@ -121,3 +122,24 @@ def test_fragment_missing(self):
...
+            with open(fragment_path, "w") as f:
+                f.write("Adds gravity back")
+
+            call(["git", "add", fragment_path])
+            call(["git", "commit", "-m", "add a newsfragment"])
+
+            check_call(
+                [sys.executable, '-m', 'towncrier.check', "--compare-with", "master"],
+                stdout=PIPE,
+            )
```
adiroiban, line 145: I don't like how this assertion is done.

From Python docs

Note
Do not use stdout=PIPE or stderr=PIPE with this function. The child process will block if it generates enough output to a pipe to fill up the OS pipe buffer as the pipes are not being read from. 

```
proc = subprocess.Popen([sys.executable, '-m', 'towncrier.check', "--compare-with", "master"], stdout=PIPE, stder=PIPE)
try:
    outs, errs = proc.communicate(timeout=15)
except TimeoutExpired:
    proc.kill()
    outs, errs = proc.communicate()
```

and then check that err is empty and the out has the expecte output and no sign of error.
Not sure how to check for no errors in stdout... but maybe in the future we can have towncrier sending all errors to stderr and then checking for an empty stderr is enough.

adiroiban (approved):
Look good. Only a medium comment regarding the test.
I think that the test can be updated and then merged without another review.

Thanks!

src/towncrier/check.py
altendky: Several tests fail in 2.7 if you drop the `getattr()` bit.  So, we need to choose an encoding in that case.  Unless the `.decode()` below shouldn't always happen?  But the extra bit handled here is that if `.encoding` is present but is `None`, we need to pick an encoding.

That said, the comment confused you, and also me for a minute.  So, it needs adjusted.

```suggestion
    # Use UTF-8 both when sys.stdout does not have .encoding (Python 2.7) and
    # when the attribute is present but not None (explicitly piped output and
    # also some CI such as GitHub Actions).
    encoding = getattr(sys.stdout, "encoding", None)
    if encoding is None:
        encoding = "utf8"

```

adiroiban: Looks good :)


> # when the attribute is present but not None (explicitly piped output and

# when the attribute is present but **set to** None (explicitly piped output and

src/towncrier/test/test_check.py
altendky: Implemented without the timeout since py2 doesn't have timeout.  Could go for [`subprocess32`](https://pypi.org/project/subprocess32/) or something if it's particularly important  (sure, it's obviously nice).

adiroiban: it's ok without the timeout

altendky:
Kick again to see if codecov works better

PR #301, opened 2020-12-14
Correct to start_string consistently

### LINKED ISSUE #277: start_string vs start_line
We have here some inconsitency between the namings `start_string` vs `start_line`
https://github.com/twisted/towncrier/blob/5c431028a3b699c74b162014e907272cbea8ac81/src/towncrier/_settings.py#L150

If you specify `start_line` in the `pyproject.toml`, it will not be recognized correctly.
We have here some inconsistency in the code base itself, & also in the documentation.

Should be easy to solve:
1. Decide for which option to go: `start_string` or `start_line`
2. Unify the codebase
3. Adapt the docs accordingly

altendky:
Fixes #277.


altendky:
At the moment I'm planning on overriding this coverage complaint on a line touched only to make a variable name more consistent.  Perhaps at some point I'll attack the missing coverage directly.

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,58 @@ def test_bullet_points_false(self):
             """
             ).lstrip(),
         )
+
+    def test_start_string(self):
+        """
+        Rendered content added after specified start string.
```
adiroiban, line 677:
```suggestion
       The `start_string` configuration is used to detect the starting point for inserting the generated release notes. 
       A newline is automatically added to the configured value.
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,58 @@ def test_bullet_points_false(self):
...
+        """
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            with open("pyproject.toml", "w") as f:
+                f.write(dedent("""\
+                    [tool.towncrier]
+                    start_string="a different start line"
```
adiroiban, line 685:
```suggestion
                    start_string="Release notes start marker"
```

src/towncrier/test/test_write.py
```diff
@@ -226,3 +227,46 @@ def test_append_at_top_with_hint(self):
             output = f.read()
 
         self.assertEqual(expected_output, output)
+
+    def test_multiple_file_no_start_string(self):
+        """
+        Multiple file output with no start string works.
```
adiroiban, line 233: Maybe move this to test_build to have better coverage  / end to end testing.

```suggestion
        When no `start_string` is defined, the generated content is added at the start of the file.
```

src/towncrier/test/test_build.py
```diff
@@ -671,3 +671,58 @@ def test_bullet_points_false(self):
...
+                """))
+            os.mkdir("newsfragments")
+            with open("newsfragments/123.feature", "w") as f:
+                f.write("Adds levitation")
```
adiroiban, line 689: I see this pattern many times

I think it can be extracted into a helper ... and the helper can also auto call os.mkdir if not already exists.

```suggestion
            self.createFragment('123.feature', "Adds levitation")
```

README.rst
```diff
@@ -137,7 +137,7 @@ Towncrier has the following global options, which can be specified in the toml f
     version = "1.2.3"  # project version if maintained separately
     name = "arbitrary project name"
     template = "path/to/template.rst"
-    start_line = "start of generated content"
+    start_string = "start of generated content"
```
adiroiban, line 140: Maybe a better name and documentation

```suggestion
    start_marker = "Text used to detect where to add the generated content in the middle of a file. Generated content added after this text. Newline auto added."
```

src/towncrier/newsfragments/277.doc.rst
```diff
@@ -0,0 +1 @@
+``start_line`` corrected to ``start_string`` to match the long standing implementation.
```
adiroiban, line 1:
```suggestion
``start_line`` renamed to ``start_string`` to match the long standing implementation.
```

adiroiban (approved):
Looks good.
Only minor comments.

If we are doing the rename, my suggestion is to rename it to "start_marker" as I think it's a better name.

The `start_string` name tells us that it's the start of something and that the type is string.
I don't think that we need to include the type name in the configuration value name.

altendky:
> If we are doing the rename, my suggestion is to rename it to "start_marker" as I think it's a better name.

The actual name read out of the configuration file is not being changed in this PR.  The docs are being corrected and the code is being changed to match.  I agree with the sentiment that the name isn't great but wasn't shooting for breaking changes here.  If you agree with this path I'll skip those comments and address the rest.

src/towncrier/_writer.py
```diff
@@ -35,11 +35,12 @@ def append_to_newsfile(
...
 
         f.write(top_line.encode("utf8"))
         f.write(content.encode("utf8"))
-        if existing_content[0]:
-            f.write(b"\n\n")
-        f.write(existing_content[0].lstrip().encode("utf8"))
+        if existing_content:
+            if existing_content[0]:
+                f.write(b"\n\n")
+            f.write(existing_content[0].lstrip().encode("utf8"))
```
altendky, line 46: This is probably better fixed by setting `existing_content = [u"", u""]` above when `single_file` is `False`.

```suggestion
        if existing_content[0]:
            f.write(b"\n\n")
        f.write(existing_content[0].lstrip().encode("utf8"))
```

src/towncrier/_writer.py
```diff
@@ -24,7 +24,7 @@ def append_to_newsfile(
         else:
             with io.open(news_file, "r", encoding="utf8") as f:
                 existing_content = f.read()
-        existing_content = existing_content.split(start_line, 1)
+        existing_content = existing_content.split(start_string, 1)
     else:
         existing_content = [u""]
```
altendky, line 29:
```suggestion
        existing_content = [u"", u""]
```

altendky: Hmm, or maybe not since the other code can also result in either a single or double element list depending...  Time to think a bit more.

adiroiban:
>  I agree with the sentiment that the name isn't great but wasn't shooting for breaking changes here. If you agree with this path I'll skip those comments and address the rest.

Yes. This can be merged with the current name (via the approved action) :) Thanks!

src/towncrier/newsfragments/277.doc.rst
altendky:
```suggestion
``start_line`` corrected to ``start_string`` in the readme to match the long standing implementation.
```

PR #302, opened 2020-12-14
Clean no config exit

### LINKED ISSUE #84: 'NoneType' object has no attribute '__getitem__'
Running towncrier without any config file crashes:
```
$ bin/pip install towncrier
$ bin/towncrier --draft
  File "/tmp/myvenv/lib/python2.7/site-packages/click/core.py", line 722, in __call__
  File "/tmp/mvenv/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/tmp/mvenv/lib/python2.7/site-packages/click/core.py", line 895, in invoke
  File "/tmp/mvenv/lib/python2.7/site-packages/click/core.py", line 535, in invoke
  File "/tmp/mvenv/lib/python2.7/site-packages/towncrier/__init__.py", line 38, in _main
    return __main(draft, directory, project_version, project_date, answer_yes)
  File "/tmp/mvenv/lib/python2.7/site-packages/towncrier/__init__.py", line 50, in __main
    if config['template'] is None:
TypeError: 'NoneType' object has no attribute '__getitem__'
```

altendky:
Fixes #84.


src/towncrier/build.py
```diff
@@ -72,9 +71,9 @@ def _main(
     except Exception as e:
         if isinstance(e, ConfigError):
             print(e)
+            sys.exit(1)
         else:
-            traceback.print_exc(file=sys.stderr)
-        raise
+            raise
```
adiroiban, line 76: I think that the ConfigError should also be sent to stderr... as it is still an error. Just that is an error that doesn't require a traceback.

Maybe use this to always exit (raise SystemExit) in `_main` ... but maybe your fix work better with Click.... so is ok.
 
```python
    if isinstance(e, ConfigError):
        print(e, file=sys.stderr)
    else:
        traceback.print_exc(file=sys.stderr)
    sys.exit(1)
```

src/towncrier/newsfragments/84.bugfix.rst
```diff
@@ -0,0 +1 @@
+Stop showing tracebacks for known configuration errors.
```
adiroiban, line 1:
```suggestion
Configuration errors found during command line execution now trigger a message to stderr and no longer show a traceback.
```

adiroiban (approved):
All good. Only minor comments.

Please consider printing the config error to stderr (and any otther error message).

But I think that this can be fixed and then merged without any review.

src/towncrier/newsfragments/84.bugfix.rst
altendky: Agreed, this is a much better description.  Thanks.

src/towncrier/build.py
altendky: Seems I forgot to go back and convert this to `except ConfigError as e:`...  But, to your comments, sure, I'll put the error on `stderr`.  For the traceback, I figured that an uncaught exception ends up on `stderr` anyways.  That was the idea with the bare `raise` (or after adjusting to `except ConfigError`, nothing)  Though, the explicit `sys.exit(1)` does make the exit code more...  explicit.  I don't know if that's good for unexpected exceptions or not.

As to Click, it does offer its own version of this mechanism.  I started to switch to that briefly before seeing that this was already here.  I figured I would leave this focused and consider a full application switch over to Click exceptions separately (if at all).  So, I don't think Click needs to be significantly included in consideration how this should be written.

src/towncrier/build.py
```diff
@@ -69,12 +68,9 @@ def _main(
...
         )
-    except Exception as e:
-        if isinstance(e, ConfigError):
-            print(e)
-        else:
-            traceback.print_exc(file=sys.stderr)
-        raise
+    except ConfigError as e:
+        print(e, file=sys.stderr)
+        sys.exit(1)
```
altendky, line 73: This seems like the simplest change that 1) avoids the traceback for `ConfigError`s and 2) doesn't double up the traceback on regular exceptions.  Though, admittedly, 2) is actual scope creep for this PR.  `:]`  I tried your suggestion which resulted in test failures due to getting `SystemExit` rather than `ValueError` exceptions, for example.  If you think that is in fact better, I can switch back and update the tests.

adiroiban:
This can be merged  as it is. Thanks. It is already an improvement over the main branch.

I prefer SystemExit as for me it's an "expected error" / business as usual type of event. I don't see it as an error, it's just an instruction to end the execution ASAP :)

A ValueError is an "unexepected error" that most probably signals a bug.

But as mention, this PR can be merged. Thanks!

altendky:
The 'expected' `ConfigError` is being turned into a `SystemExit`.  The 'unexpected' `ValueError` in some test _also_ got turned into `SystemExit` using the code you suggested.  As the PR is now (about to merge), the `ValueError` is retained.

Well, I guess before merging I have to figure out what happened to codecov...

PR #303, opened 2020-12-15
Use common code between draft and not for combining top line and rendered

### LINKED ISSUE #105: Preview different from final output
As mentioned in #97 I get different output from preview to final. Now I was able to reproduce it.

The template:
```
{% for section in sections %}
{% set underline = "-" %}
{% if section %}
{{section}}
{{ underline * section|length }}{% set underline = "~" %}

{% endif %}
{% if sections[section] %}
{% for category, val in definitions.items() if category in sections[section] %}
{% if definitions[category]['name'] %}

{{ definitions[category]['name'] }}
{{ underline * definitions[category]['name']|length }}
{% endif %}

{% if definitions[category]['showcontent'] %}
{% for text, values in sections[section][category]|dictsort(by='value') %}
- {{ text }}

{% endfor %}
{% else %}
- {{ sections[section][category]['']|sort|join(', ') }}


{% endif %}
{% if sections[section][category]|length == 0 %}

No significant changes.


{% else %}
{% endif %}
{% endfor %}
{% else %}

No significant changes.


{% endif %}
{% endfor %}
```

The pyproject.toml:
```toml
[tool.towncrier]
package = "project"
filename = "src/project/docs/changes.rst"
directory = "changes/"
title_format = "{version} - {project_date}"
template = "changes/_template.rst"

  [[tool.towncrier.type]]
  directory = "change"
  name = ""
  showcontent = true

  [[tool.towncrier.type]]
  directory = "migration"
  name = "Contains migration"
  showcontent = true

  [[tool.towncrier.type]]
  directory = "solr"
  name = "Requires Solr update"
  showcontent = true
```

The draft looks like this:
```
% towncrier --draft
Loading template...
Finding news fragments...
Rendering news fragments...
Draft only -- nothing has been written.
What is seen below is what would be written.

10.0 - 2018-01-01
=============


- bla

%
```

The changes file afterwards looks like this:
```
10.0 - 2018-01-01
=============

- bla


9.0 - ...
=====
```

Notice the difference in newlines below the equal signs and after the changelog entry. The goal is to have one after the equal signs and two after the changes before the previous version.

altendky:
Fixes #105.


fschulze:
I can confirm this fixes the issue.

adiroiban (dismissed):
Looks good.

----

I would say that `top_line` should be moved outside of `append_to_newsfile.

Also, what is `single_file` in the context of `append_to_newsfile` ? :)
If single_file is False it will make existing_content a list and compare it with top_line 
Doesn't look ok :)

My suggestion is to extract the duplicate version detection code outside of `append_to_newsfile`. Have a dedicated name for this method and add all the checks there :)

altendky:
> I would say that `top_line` should be moved outside of `append_to_newsfile.

Perhaps I am misreading this, but isn't that what I did here?  Are you confirming that is good, or suggesting something else should happen?

> Also, what is `single_file` in the context of `append_to_newsfile` ? :)
> If single_file is False it will make existing_content a list and compare it with top_line
> Doesn't look ok :)

I think single file is the standard usage where towncrier tries to inject new news into the right place in an existing file (generally) where as not-single-file is if you make a separate output file each time you process news fragments.  Maybe this is more common in bigger projects where you might have a large output file for each release and then if you want a comprehensive all-history output you could join each individual release file together.  Maybe?  So, if it is not a single file, you don't need to do all the reading of the existing content and injecting of new content, just write the new content over anything that may or may not be there.

I'm not super excited about this code design, but I think it works?  Well, other than the fix over at https://github.com/twisted/towncrier/pull/301#discussion_r546397440 which you made me rethink the fix for (and I already rethought again...).

> My suggestion is to extract the duplicate version detection code outside of `append_to_newsfile`. Have a dedicated name for this method and add all the checks there :)

What version detection code is there in `append_to_newsfile()`?  I feel like I may have missed something here.

adiroiban:
> Perhaps I am misreading this, but isn't that what I did here? Are you confirming that is good, or suggesting something else sh

I was suggesting this removal :)

```diff
- def append_to_newsfile( directory, filename, start_line, top_line, content, single_file=True
+ def append_to_newsfile( directory, filename, start_line, content, single_file=True
```

adiroiban:
> What version detection code is there in `append_to_newsfile()`? I feel like I may have missed something here.

This is the code in which `append_to_newsfile` is not only appending, but also doing a sanity check to make sure we don't have duplicated notes

```python
    if single_file:
        if not os.path.exists(news_file):
            existing_content = u""
        else:
            with io.open(news_file, "r", encoding="utf8") as f:
                existing_content = f.read()
        existing_content = existing_content.split(start_line, 1)
    else:
        existing_content = [u""]

    if top_line and top_line in existing_content:
```

src/towncrier/templates/default.rst
```diff
@@ -1,13 +1,12 @@
-{% if top_line %}
-{{ top_line }}
-{{ top_underline * ((top_line)|length)}}
-{% elif versiondata.name %}
+{% if render_title %}
+{% if versiondata.name %}
```
adiroiban, line 2: Would this work?

```suggestion
{% if render_title and versiondata.name %}
```

altendky:
The catchup merge brought in some problems.  I reworked it a bit.  I didn't get `top_line` completely out of `append_to_newsfile()`, but I did remove one of the uses.  Hopefully this structure for handling the top line/title completely outside the template in the default case is a bit cleaner.

Sorry, but I feel like this is enough of a change to deserve a re-review.  I'm also not convinced this is a well focused PR, so 'go break this down into more pieces' is a perfectly reasonable response.

src/towncrier/test/test_build.py
```diff
@@ -740,12 +740,25 @@ def test_title_format_false(self):
...
-            # Towncrier ignores .rst extension
-            with open("foo/newsfragments/124.feature.rst", "w") as f:
-                f.write("Extends levitation")
+            with open("template.rst", "w") as f:
+                f.write(dedent("""\
+                    Here's a hardcoded title added by the template
+                    ==============================================
+                    {% for section in sections %}
+                    {% set underline = "-" %}
+                    {% for category, val in definitions.items() if category in sections[section] %}
```
adiroiban, line 753: it looks like all these jinja directives from the template are not used in this test.

I don't understand why you would like to have `title_format = false` **and** a custom template.
With a custom template you can just don't include the title part without having to use False.

src/towncrier/test/test_build.py
```diff
@@ -836,3 +844,65 @@ def test_start_string(self):
         """)
 
         self.assertEqual(expected_output, output)
+
+    def test_with_topline_and_template_and_draft(self):
+        """
+        Spacing is proper when drafting with a topline and a template.
```
adiroiban, line 850: I am not sure what the difference between `test_title_format_custom` and this test.

adiroiban (commented):
Thanks for the update.

Just a comment for now... but I am kind of -1 or the PR as it is now.

I am confused by the tests :)

src/towncrier/test/test_build.py
altendky: It at least differs in that this specifies a custom template while `test_title_format_custom` leaves the default template in use.  When writing this test it ended up a bit confusing to me too how all the bits were interacting but it did seem to recreate the issue reported at #105.

I feel like the existing implementation code had evolved and needed cleanup but as it was it offered corners where specifying a title format \_with\_ a custom template \_and\_ draft (and maybe something about the title-less feature section?) could result in some unique bug.

src/towncrier/test/test_build.py
altendky: Yes, this template should be able to have all these normal template bits removed.  Change incoming.

Does it matter why you would want that scenario if it is documented and released functionality?  Or is this question more about making sure we have a common understanding of the feature than questioning if the situation should be tested?  I haven't really thought through what I think the configuration design ought to be as I'm targeting a bugfix+feature release without introducing breaking changes (nor deprecations).

None the less, false is exactly for the case where you specify a template that provides its own title.  If you don't specify false then you will (should?  i think?...  `:[`) get the default title before whatever your template does.  I do feel funny about the existing non-empty string or empty string/unspecified or false possibilities for `title_format`.  I guess I see the use cases along the lines of 'default title format and template' (unspecified or empty string), 'the default template is fine but i want to change just the title format' (non-empty string), and 'gosh darn it let me just write everything in my custom template' (false).

Hopefully this helps...  this is definitely an area that strikes me as simply confusing.

altendky (commented):
Let me know if the explanation of the tests helps, or maybe it just reinforces the -1.  I could understand that.  The issue reported in #105, as I recall, was a weird corner case and perhaps I should approach it with a differently formed PR.

src/towncrier/test/test_build.py
adiroiban: > Does it matter why you would want that scenario if it is documented and released functionality? Or is this question more about making sure we have a common understanding of the feature than questioning if the situation should be tested? 

It's more about understanding the feature and its valid use cases.

I imagine someone reading this test 1 year in the future, looking at a bug and will say: Why you need this feature with a custom template.

So I prefer to have a test in which the feature is tested like in real file, and not using a combination of configuration option that  are possible, but will never be used.

At the same time, I think that this test is wrong. We should have a full test for the default template itself.

This should also be a test for the `{% if render_title %}` part of the default template

src/towncrier/test/test_build.py
adiroiban: Understood. But I think that this test can be removed.
Now that we use the exact same output for draft and final,#105 should be fixed.

I would prefer not to have this tests as it's long and confusing.

The test only checks the output of draft.

A good test for #105 would had generated the output with and without draft and compare them to make sure they are exactly the same.

But I don't think that we need extra test for that.

adiroiban (changes requested):
Sorry for the late reply.

My changes for this :

* `test_title_format_false` should use the default template, to cover the logic from default template.
*

src/towncrier/build.py
```diff
@@ -141,30 +141,49 @@ def __main(
         top_line = config["title_format"].format(
             name=project_name, version=project_version, project_date=project_date
         )
+        render_title_with_fragments = False
+        render_title_separately = True
+    elif config["title_format"] == False:  # noqa: E712
```
webknjaz, line 146: Use the identity check for booleans and None:
```suggestion
    elif config["title_format"] is False:
```

glyph:
@altendky @adiroiban Is this PR still alive? I have lost the plot on the controversy over this change, and I'm not sure how to review it taking that history into account.

adiroiban (approved):
Having this blocked was my fault for commenting to much in terms of automated tests.
The tests are very verbose and might break with future changes that are not related to the draft functionality.

And there was no other towncrier developer to step in and approve it.

Since we have limited resources for towncrier development, let's merge this as the issue if fixed.

We can worry about code cleanup and reducing the test side-effects later :)

adiroiban:
Thanks @glyph for the follow up.
I think that if anything is no OK with trunk now, is better to fix it in a separate PR.

webknjaz:
@altendky @adiroiban any chance to release this soonish? AFAIK, this fixes poor UX in https://github.com/sphinx-contrib/sphinxcontrib-towncrier that's been annoying since about v21.

PR #307, opened 2020-12-19
Try to get codecov responding quickly

altendky:
Draft for:
- [ ] This is presently an exploratory PR...  which doesn't seem to have helped with prompt codecov result reporting.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,7 @@ include pyproject.toml
 include tox.ini
 recursive-include src *.rst
 
+exclude codecov.yml
 exclude bin
 exclude src/towncrier/newsfragments
 
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+codecov:
+  notify:
+    # The reference documentation suggests that this is the default setting:
+    # https://docs.codecov.io/docs/codecovyml-reference#codecovnotifywait_for_ci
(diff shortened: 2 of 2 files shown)


hynek:
may i recommend https://hynek.me/articles/ditch-codecov-python/ instead to simplify the whole thing?

adiroiban:
we can give it a try as a toy task. and sprinkle it with diff-cover that is piped to a github comment :) 

But codecov.io is much more than a 100% pass/fail status check.
There is an HTML view of the coverage, inline annotations, flapping lines , diff coverage.

I think that codecov.io for towncrier is ok.

I know that codecov.io can be a pain, especially for big projects .


---------

I think the main issue with codecov.io is that being a free ride is hard to fund its development.
I like the codecov.io project and I hope it can solve these reliability issues and be a success :)

hynek:
The problem is less about what it does and more about how it's flaky. Latest example: my PR that didn't touch code supposedly lowered coverage in files I didn't touch…

adiroiban:
I anyone wants to implement coverate.py + GitHub action reporting, I would be more than happy to access such a PR...

I can live with a bit of flakiness for towncrier.

I can imagine that for NASA space missions codecov.io flaky behaviour is not acceptable ...but I think we are fine.

hynek:
Very much stale so closing :)

Issue #269, opened 2020-06-09
New release post 19.2.0 


hmaarrfk:
Recent updates have allowed for the use of a towncrier.toml file that doesn't conflict with pyproject.toml which triggers special pip behavior.

It would be nice to be able to use this.

A 20.XX release would be greatly appreciated!

Thanks!

rohanpm:
I would also appreciate a new release for the sake of the `template = "towncrier:<bundled>"` feature added in 3d01a194a68f48d.

hmaarrfk:
Sorry for bumping this request, but any news on this?

altendky:
Aside from a release, I have asked a couple times in IRC about the review etc policies here.  I'll check on the maillist now too.

saroad2:
Hey, I'm joining the request in this PR.
I would love to see a new release of *Towncrier* being released. Really want to start integrating it into my CI/CD  process.

If help is needed in review/implementing new features, let me know and I'll give a hand!

altendky:
For awhile I had good momentum so I had planned to just finish out the rest of the stuff I was interested in before release, but the progress has slowed down.  I'll look back at doing a release soon without waiting for those other bits.

altendky:
Welp, we're failing to get it done quickly but we are still doing stuff over in #313.

hmaarrfk:
Thanks, i periodically check to see if i should delete my hacked version. Glad to see that you are simplifying the scope!

altendky:
Well, kinda...  `:|`  Now we've fallen into the 'define and automate the release process before releasing instead of after' hole.  Lemme see where everything's at...

hmaarrfk:
i guess i read too quickly ^_^

hmaarrfk:
I'm going to say, that since you have CIs for testing, releasing a pure python package manually is not the end of the world. You could even distribute it as source....

altendky:
Agreed.  It is a team effort here with multiple opinions.  But, I did make a work around for one of the hang ups and that got approved and merged.  I need to catch up https://github.com/twisted/towncrier/pull/315.  After that, the checklist in https://github.com/twisted/towncrier/issues/313 looks pretty free flowing.

Regardless, this is certainly frustrating for you and others.  I'm sorry for the hassle.  With a little luck this will be less painful in the future.  (and maybe next time I'll actually get the bits I wanted fixed and added included in the release `:]`)

altendky:
[...]

PR #310, opened 2021-01-09
Merge in 19.09-rc1

altendky:
Hopefully `hawkowl/release-19.09` is the right place to merge from to include the 19.09rc1 release.


altendky:
I think that the newsfragment check failing is fine, maybe good, since this is an attempt to merge in a release branch.  Albeit a bit late...

adiroiban:
Thanks for the docs update :) Nice update.

Why not release it as 19.09  .

Also, I think that a tag should be created for 19.09rc1 or just create tag and release 19.09  https://github.com/twisted/towncrier/tags  based on 19.09rc1 code.

Were there any changes / bugs found during 19.09 to require a rc2?

At the same time, I think that for towncrier we should keep things simple as just release normal versions.
if a bug is found in 21.1.0 we can always release 21.1.0 to fix it.

altendky:
To be clear, I didn't do any 'doc updates' here.  This is just dealing with the fact that the 19.9 release branch never got merged back to `master`.

As of now, https://github.com/twisted/towncrier/commits/hawkowl/release-19.09 has 1f396eab7acd6eece8cb3294a9f1d7921fde76cc and 280e5b3bbcd84812b6a4a5f159a8194db757b309 beyond what appears to likely be the commit released as 19.9.0-rc1 which I think is 101d60ea150a8c73ca479f97c2707a3f07e0f61d.  If you agree, I'll push a tag for 19.9.0-rc1 there.  Or should it be 19.9.0rc1 as written in `NEWS.rst`?  `:|`

I was planning on just walking away from 19.9.0 and leaving it with only the RC.  It's been 15 months and there have been lots of fixes since.  I guess I'd be game for doing a 19.9.0 release but I'm not sure what the value is at this point?  If checking for bugs introduced in rc1 to include in the release, I would have to go back and identify when all the bug reports were introduced etc and potentially cherry pick fixes back to that release.  It seems like a lot of work to get right and I don't clearly see the value at this point.  If I'm missing something, do let me know.

I've expressed my thoughts elsewhere on keeping RCs around, but we don't have release automation here (yet?) so I'm game for not bothering with RCs, at least for now.

altendky:
I guess I ought to be explicit about the context here.  I was starting to look at doing a release roughly in the state it is now to address https://github.com/twisted/towncrier/issues/269#issuecomment-756785977.  There are three remaining [bug-tagged issues](https://github.com/twisted/towncrier/issues?q=is%3Aissue+is%3Aopen+label%3Abug) right now.  I didn't get all the bits done that I wanted to, neither the list of things I personally want nor the goal of addressing all open PRs and bug reports, but the request for the release reminded me that the momentum had dissipated and so it seems relevant to get a release out with what has been done.

I'm not saying I won't be working more on finishing out the other bits, but reality ...  is.  And I have been spending time on other projects for a couple weeks.

adiroiban:
I see that release 19.09 Updated 16 months ago by hawkowl .
This looks like an exceptional situation.

Exceptional situation might need exceptional actions :)
-----

> Hopefully hawkowl/release-19.09 is the right place to merge from to include the 19.09rc1 release.

My confusion is from the PR description. 
I had no idea what that want to say :)
Your later comments clarify... and I think that PR description should be updated to include that info.

Also, the PR description should include the commit that you think should be tagged.

-------

> what appears to likely be the commit released as 19.9.0-rc1 which I think is 101d60e

If you think that `101d60e` is the actual release, I would just to just tag it as 19.9 and make it the "final release" to have a closure on this :)

------


This looks like a release pull request to me.

So I would say, make this PR the release for 19.9.

Tag just 19.9  . No need to tag the release candidate. 

------

Later we can decide which versioning convention to use.

For example, I prefer consistent 19.9.0 over 19.09. That is no zero padding and patch number is explicit set to 0.

altendky:
I don't think this PR is useful for any releases since it already has `master` merged in to it ready for re-integration merging.  That was, in fact, it's entire purpose.  See #313 for further discussion of releases.

adiroiban:
See #313

PR #311, opened 2021-01-09
less leading spaces in 173.feature.rst

altendky:
This resulted in an extra space of indent on the hanging lines which caused misformatting when rendered.


altendky:
@adiroiban, what do you think.  Is it a bug that this isn't handled properly?  Or just a formatting error in the newsfragment.

adiroiban (approved):
Looks good. Thanks.

Maybe update towncrier to strip leading / trailing spaces with creating the final document.

I think that lines are already re-wrapped ... so we can also strip the line before re-wrapping.

PR #314, opened 2021-01-17
Add build-system to pyproject.toml for pep517 (tox-wheel)

altendky:
Stalled on:
- [ ] https://github.com/tox-dev/tox/issues/1850

PR #315, opened 2021-01-17
twisted-alike (so similar) automatic publishing on gha

altendky:
Stalled on:
- [x] ~~#314~~ #323 (probably) is needed as the Twisted setup this comes from uses PEP 517 builds.
- [x] If going with #323 then the wheel and sdist are already built and should just be checked and uploaded.


adiroiban:
This looks good.
For validating this PR my suggestion is to first push something like 11.1dev1 tag to check that the publish is blocked due to tag mismatch.

Then update the version to 21.3.dev1 and push that tag to make sure publishing to public repo works.

Then we can merge this and delete the dev1 release from PyPi.

Thanks!

altendky:
[Working tagged build](https://github.com/twisted/towncrier/actions/runs/666386455) ([PyPI](https://pypi.org/project/towncrier/21.3.0.dev2/))
[Failing build due to incorrect tag](https://github.com/twisted/towncrier/actions/runs/666410345)
[Same commit as above tagged properly](https://github.com/twisted/towncrier/actions/runs/666432988) ([PyPI](https://pypi.org/project/towncrier/21.3.0.dev4/))

.github/workflows/ci.yml
```diff
@@ -3,6 +3,7 @@ name: CI
 on:
   push:
     branches: [ master ]
+    tags: [ "**" ]
```
adiroiban, line 6: why double ? just asking :)

src/towncrier/_version.py
```diff
@@ -7,5 +7,5 @@
 
 from incremental import Version
 
-__version__ = Version("towncrier", 19, 2, 0)
+__version__ = Version("towncrier", 21, 3, 0, dev=1)
```
adiroiban, line 10: Maybe use dev4 as this was the last released ?

```suggestion
__version__ = Version("towncrier", 21, 3, 0, dev=4)
```

adiroiban (approved):
Wonderful work. Thanks! Looking forward to the public release :)

src/towncrier/_version.py
altendky: Huh...  I definitely didn't mean to do this, lemme find where I lost that.  Thanks.

.github/workflows/ci.yml
altendky: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
> `*`: Matches zero or more characters, but does not match the / character. For example, Octo* matches Octocat.
> `**`: Matches zero or more of any character.

So it will match even if the tag ends up having a `/` in it.  Not something we need, just my default at this point when saying 'anything'.

src/towncrier/_version.py
altendky: Yeah, merge and fork and branch and PR and everything more confusion.  Sorry, thanks.  There was something else missing so this is a good catch.

https://github.com/twisted/towncrier/commit/149c27f5c45b030ed803accb5c2ce5539dd6e425

Issue #294, opened 2020-12-06
Stop assuming the default primary branch name in tests


altendky:
When I installed a new version of Git and let it set the default to `main` this broke some tests in towncrier that assume the main branch will be `master`.  We need to either detect the primary branch name or just force it in the tests.  Forcing seems simpler and `main` seems a good option.

https://github.com/twisted/towncrier/pull/287#discussion_r537056933

PR #316, opened 2021-01-17
Catch up tox envs to include py38,py39

altendky:


altendky:
So I guess technically we can't test on 3.9 on Windows due to Twisted still not being released with 3.9 support.  Maybe that justifies not including it in the environment list?

src/towncrier/newsfragments/316.misc.rst
```diff
@@ -0,0 +1 @@
+
```
adiroiban, line 1: Maybe add a note that 3.8 and 3.9 are "officially" supported.

Like:

CPython 3.8 and 3.9 are now part of our automated test matrix and are officially supported.

adiroiban (approved):
All good. Thanks!

Twisted is only needed for trial...so as log trial works on py3.9 we don't have to worry about official Twisted release.

src/towncrier/newsfragments/316.misc.rst
altendky: I would have thought we already had a newsfragment from when the matrix was expanded and this was forgotten...  but I just looked and didn't see one, I'll add one.

altendky:
Yes, only Trial is needed but Twisted can't even be installed...  so we can't run our tests in 3.9 on Windows.  Well, not with released Twisted and we keep almost having a released Twisted that will work...  almost...

adiroiban:
Well... we have 3.9 on Linux.

I am not sure that we need to run towncrier on every OS  with every python.

I would say that running on Linux with every python and then on macOS and Windows only on latest Python.

But the tests are fast... so no big deal.

While checking the tests on Linux with py3.9 I have observed this output

https://github.com/twisted/towncrier/runs/1716924240#step:6:30

```
test_no_confirmation ... hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /tmp/tmpia63lpf2/.git/
[master (root-commit) 167cb11] Initial Commit
 4 files changed, 5 insertions(+)
 create mode 100644 foo/__init__.py
 create mode 100644 foo/newsfragments/123.feature
 create mode 100644 foo/newsfragments/124.feature.rst
 create mode 100644 pyproject.toml
```

I think that it needs some updates as the output of all the tests should be captured and checked....but this is low priority

altendky:
We could bet that Linux developers carefully keep in mind oddities that differ on Windows across Python versions...  but that sounds like a bad bet.  `:]`

Yes, git is hinting something and I ran into that when testing in Windows as I had installed a git update and set it to default to main and some tests had assumed master.  There is https://github.com/twisted/towncrier/issues/294 on this topic.  You want the tests to test the output of git for messages like this?  Checking all `stdout` sounds super fragile but maybe I'm missing something.

PR #317, opened 2021-01-17
Release instructions

altendky:


CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- If working on the first release candidate from this branch, create a PR named in the form ``Release 19.9``.
+
+- Request a review and address raised concerns until receiving an approval.
+
+- Tag that commit such as ``19.9rc1`` and push the tag to the primary repository.
+
+  - This will result in another build which will publish to PyPI.
+  - Confirm the presence of the release on PyPI.
+  - # TODO: use the GitHub release and tagging feature?
```
altendky, line 54: I haven't used this much so I don't know any nuances.  Should this just be used for all tagging of release candidates and final releases?  Will we have to manually add the sdist and wheel artifacts?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- If not processing the final release, `dismiss the approving review <https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/dismissing-a-pull-request-review>`_.
+
+  - The review process will be reused for any subsequent release candidates as well as the final release so it must be cleared at each stage.
+
+- # TODO: notify people some how some way, the maillist?  hmmm...
```
altendky, line 60: The Twisted maillist that is...  I dunno.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- If ready for a final release, remove the release candidate indicator from the version.
+
+  - Edit ``src/towncrier/_version.py`` such as ``__version__ = Version('towncrier', 19, 9)`` to remove the release candidate indication.
+
+  - Return to the towncrier build step and continue.
+
+- Once the final release has been made, leave the approving review intact and merge the PR.
+
+# TODO: should we have an indicator on the version for interrim development or just leave the last release version intact?
```
altendky, line 76: Probably ought to just leave this for a separate activity I guess.  It isn't needed to do a release and isn't presently being done at all.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
+Releasing
+=========
+
+Summary
+-------
+
+- Create a release branch in the main repository, not in a fork.
+- Prepare a commit for a release candidate, get an approving review, and tag the commit which will upload artifacts to PyPI.
```
adiroiban, line 8:
```suggestion
- Create a commit for a release candidate inside the release branch (see next for details
- Create a Pull Request for the release. The same branch and PR will be used for both the release candidate and  the final release.
- Get an approving review
- Once approved, tag the last commit in the branch as a release candidate.
- Push the tag and this trigger the upload of artifacts to PyPI.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+- Notify people and give time for feedback.
```
adiroiban, line 9: Where to notify? I have no idea where to notify. Maybe IRC #twisted-dev  ?

How much time? -- Maybe 1 week.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+- Address concerns raised via new issues for later handling or PRs against the release branch.
```
adiroiban, line 10: My take on this.

```suggestion
- Address concerns raised via new issues and fix them in dedicated branches that will be first merged in the main branch.
- Once merged in the main branch, cheery-pick the merge commit in the release branch.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+- Prepare a commit for a final release, get an approving review, and tag the commit which will upload artifacts to PyPI.
```
adiroiban, line 11:
```suggestion
- Prepare a commit for a final release and push it to the release branch
- Get an approving review for the final release.
- Tag the commit using the final release version. This will trigger the upload of artifacts to PyPI.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+  - towncrier uses `CalVer <https://calver.org/>`_ of the form ``YY.MM.micro`` with the micro version just incrementing.
+  - Normalize the version according to `PEP 440 <https://www.python.org/dev/peps/pep-0440/#normalization>`_.
+
+    - This requires that ``towncrier[dev]`` extra is installed.
+    - ``venv/bin/python admin/canonicalize_version.py 19.09.00-rc1``
+    - Outputs ``19.9rc1`` which is the form to be used.
+
+- Create a release branch with a name of the form ``release-19.9`` starting from the ``master`` branch.
+
+  - This new branch will contain all tagged release candidate commits as well as the final tagged release commit.
```
adiroiban, line 34:
```suggestion
  - On the new release branch you will commit all tagged release candidate commits as well as the final tagged release commit.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- Update the version to the release candidate with the first being ``rc1`` (as opposed to 0).
+
+  - In ``src/towncrier/_version.py`` the version is set using ``incremental`` such as ``__version__ = Version('towncrier', 19, 9, release_candidate=1)``
+
+- Run ``venv/bin/towncrier build --yes`` to build the the newsfragments into the release notes document and remove the newsfragment files.
```
adiroiban, line 40:
```suggestion
- Run ``venv/bin/towncrier build --yes`` to build the the newsfragments into the release notes document and to automatically remove the newsfragment files.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- Commit and push to the primary repository, not a fork.
+
+  - It is important to not use a fork so that pushed tags end up in the primary repository.
```
adiroiban, line 44: I think that we can keep it simple. One reason is for tags, the other reason is that forks don't have access to PYPI secrets.


```suggestion
- Commit and push to the primary repository, not a fork.
```

CONTRIBUTING.rst
adiroiban: I prefer using git CLI to tag.
For small projects I don't think that we need to GItHub release GUI.

CONTRIBUTING.rst
adiroiban: mailing list can work and IRC for quick feedback...it's very low trafic

CONTRIBUTING.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- If not processing the final release, `dismiss the approving review <https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/dismissing-a-pull-request-review>`_.
+
+  - The review process will be reused for any subsequent release candidates as well as the final release so it must be cleared at each stage.
+
+- # TODO: notify people some how some way, the maillist?  hmmm...
+
+- If another release candidate is required:
+
+  - Submit PRs against the release branch to integrate the needed changes.
```
adiroiban, line 64: my idea is to fix master first...and then the release branch... but this is also ok.

adiroiban (commented):
just initial comments

CONTRIBUTING.rst
adiroiban: Good question. I think that we should have a placeholder like. "No changes since last release."

For my project, I do a stupid think and replace the release candidate in the release notes with the final note.
But this erases the project history and I think it's bad practice.

In my fork of towncrier, If I run towncrier with a new version and no news fragments I get an em

But we can have something like this for the default template

```
{% set underline = "^" %}
{% set section = sections[''] %}
{% set empty_release = True %}

{% for category, val in definitions.iteritems()%}
{% if definitions[category]['showcontent'] and section[category]|length > 0 %}
{{ definitions[category]['name'] }}
{{ underline * definitions[category]['name']|length }}

{% for text, values in section[category]|dictsort(by='value') %}
* {{ text }} {{ values|sort|join(', ') }}
{% set empty_release = False %}
{% endfor %}


{% endif %}
{% endfor %}
{% if empty_release %}
No changes since the last release.
{% endif %}
```

To produce something like

```

Version 4.8.0, 2021-02-01
-------------------------


No changes since the last release.


Version 4.8.0rc1, 2020-11-19
-------------------------


New Features
^^^^^^^^^^^^

* The embedded OpenSSL libraries used on Windows, macOS, and generic Linux were
  updated to version 1.1.1h. [#5496]
```

adiroiban (changes requested):
Looks like a very good start. Thanks. I left my comments.
You are the boss so feel free to do it as you like :)

CONTRIBUTING.rst
altendky:
```suggestion
- Notify the `Twisted-Python maillist <https://twistedmatrix.com/cgi-bin/mailman/listinfo/twisted-python>`_ of the release to allow for feedback if a pre-release and just for notification if a final release.
```

CONTRIBUTING.rst
altendky: I guess as long as the cherry pick from master is done as a PR against the release branch it doesn't really matter.  My point is to encourage that the actual release branch stays reasonably clean to alleviate confused tossing of changes there and flailing around trying to fix something.

```suggestion
  - Submit PRs against the release branch to integrate the needed changes.  Any PRs could be cherry picks from the ``master`` branch if already resolved there, or direct PRs against the release branch that will be merged back into master at the completion of the release.
```

CONTRIBUTING.rst
altendky: Explaining why is detrimental here?  hmm...

```suggestion
- Commit and push to the primary repository, not a fork.

  - It is important to not use a fork so that pushed tags end up in the primary repository, server provided secrets for publishing to PyPI are available, and maybe more.
```

CONTRIBUTING.rst
altendky: I Think the summary section is sufficiently short and well labeled to not need to reference ahead for details in the middle.  Also, other small adjustments.

```suggestion
- Create a commit for a release candidate inside the release branch.
- Create a Pull Request for the release. The same branch and PR will be used for both the release candidates and  the final release.
- Get an approving review.
- Once approved, tag the last commit in the branch as a release candidate.
- Push the tag.  This will trigger a build including the upload of artifacts to PyPI.
```

CONTRIBUTING.rst
altendky:
```suggestion
- Notify the Twisted maillist allow a week for feedback before continuing.
```

CONTRIBUTING.rst
altendky:
```suggestion
- Make sure that tickets exist for all raised concerns.  These tickets should be marked using the ``blocking`` label.
- Address each issue by either concluding that it is not really a release blocker and removing the label or by fixing via a PR to the release branch.
```

CONTRIBUTING.rst
altendky:
```suggestion
- Prepare a commit for a final release and push it to the release branch.
- Get an approving review for the final release.
- Tag the commit using the final release version. This will trigger the upload of artifacts to PyPI.
```

CONTRIBUTING.rst
altendky: I was referring to the project version, unrelated to generating release notes.  Rather, if I checkout and install an arbitrary commit of towncrier I will get one claiming to be the last release version since after doing a release the version is just left alone.

Despite having numerous changes since the release of 19.2, the latest commit on `master` still claims to be 19.2.

https://github.com/twisted/towncrier/blame/00f19b4d24ad5dfd57124f7c93f9ba2107408f8c/src/towncrier/_version.py#L10

adiroiban:
> To keep master up to date, perhaps the release branch should be merged back into master not only at the end but also after each pre-release?

I would say it's not worth the trouble.
But it's your call :)


> I was referring to the project version, unrelated to generating release notes. Rather, if I checkout and install an arbitrary commit of towncrier I will get one claiming to be the last release version since after doing a release the version is just left alone.

Sorry for missing that.

Yes. I think that after the public release there should another step to update the release to +1 and + .dev0

so if 21.1.0 was release, after the release the release branch is updated with 21.1.1dev0 and merged into master.

For a 2.1.2 release... it will be updated to 2.1.3dev0

altendky:
this is getting painful...  not complaining about your reviews, just the detail and growth of the process and a minimum of triplicate reviews and dismissing reviews and...  the review dismissals make me wonder if it would be simpler to treat each sub-release via a separate PR against the release branch.  sounds complicated, also sounds simpler since you wouldn't dismiss reviews.  anyways.  :|

CONTRIBUTING.rst
```diff
@@ -0,0 +1,88 @@
...
+
+  - Return to the step where the version is updated and increment the release candidate number.
+
+- If ready for a final release, remove the release candidate indicator from the version.
+
+  - Edit ``src/towncrier/_version.py`` such as ``__version__ = Version('towncrier', 19, 9)`` to remove the release candidate indication.
+
+  - Return to the towncrier build step and continue.
+
+- If the final release has been completed, continue below.
```
adiroiban, line 80: Maybe keep it simple.

I don't think that we need to go into the `edit src/towncrier/_version.py` as by this point people should know how do update the version.

```suggestion
- If the final release has been completed, mark current version as development `Version('towncrier', 19, 9, 1, dev=0)`` and merge it without requesting another review.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,88 @@
...
+
+- Increment the patch version by one and set to a development version.
+
+  - In ``src/towncrier/_version.py`` the version is set using ``incremental`` such as ``__version__ = Version('towncrier', 19, 9, 1, dev=0)``
+
+- Request a review and address raised concerns until receiving an approval.
+
+- With an approving review, merge.
```
altendky, line 88:
```suggestion
- Merge without waiting for an approving review.
```

adiroiban (approved):
All good. Thanks!

I think that having a single PR is easier. I guess that we can give it a try and see if things get ugly.. .then the process can be changed with a PR for each release and separate PR for each release candidates.

I think the title of  this PR can be updated and this PR merged .

Looking forward to the next release where we can practice what is documented here.

Many thanks again for putting this together.

altendky:
Thanks for seeing it through with the reviews.

PR #319, opened 2021-02-05
Prepare for tox v4

altendky:
Alphas are being pre-released so may as well try them out somewhere.

Draft for:
- [ ] This is never to be merged.  Rather, it is for testing the prereleases with the `tox4` command.  On release the command will return to just `tox`.
- [ ] ~~tox 4 being released~~
- [ ] ~~Removing the `--pre` for tox now that we know it (at least the alpha 5) works~~
  - Just specified 4 specifically (allowing alpha5+)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,30 +24,39 @@ jobs:
           - name: CPython 2.7
             tox: py27
             action: 2.7
+            tox_command: tox
           - name: CPython 3.5
             tox: py35
             action: 3.5
+            tox_command: tox
           - name: CPython 3.6
             tox: py36
             action: 3.6
+            tox_command: tox4
           - name: CPython 3.7
             tox: py37
             action: 3.7
+            tox_command: tox4
           - name: CPython 3.8
             tox: py38
             action: 3.8
+            tox_command: tox4
           - name: CPython 3.9
             tox: py39
             action: 3.9
+            tox_command: tox4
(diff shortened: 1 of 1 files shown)


altendky:
I expect various people won't really like this.  If you are reviewing and you just don't like it, don't worry too much.  Just feedback on how you think this situation ought to be handled is a start I would appreciate.

The basic issue is that despite the non-CI bit of the project requiring no changes to work with tox v4, the CI bit does need to reference the command `tox4` when it gets that version and `tox` when it doesn't (so long as we continue to support Python <3.6 since tox v4 is py3.6+).  Once you are referring to one command or the other you also need to be sure you are installing the version that will provide that command.  I guess the version could be handled as `'tox~=3.0; python_version < "3.6"' 'tox~=4.0a5; python_version >= "3.6"'`.  I'm not sure that is better though since the other selection mechanism has to stick around to identify which command to run and doing both bits the same way seems simpler.

altendky:
https://discuss.python.org/t/tool-tox-4-on-the-horizon/6537/3
> While it becomes stable it installs under the tox4 console entry and uses the tox4 working directory (not to break v3).

Ok, so apparently this is mostly pointless and probably shouldn't be merged.  Sorry for the noise.  I'll keep this around and run it every now and again to keep checking against tox 4.  But, at some point the command will return to just `tox` and none of this will be needed.

hynek:
We've switched to nox for now, so closing.

PR #322, opened 2021-03-03
Pep517 more explore

altendky:


altendky:
Went with #323, for now at least.

PR #323, opened 2021-03-03
Add build-system to pyproject.toml for PEP 517 (-m build)

altendky:
Like https://github.com/twisted/towncrier/pull/314 but with a prebuild step using [build](https://pypi.org/project/build/) instead of [tox-wheel](https://pypi.org/project/tox-wheel/) integrated into the tox run.  It has been expressed that tox-wheel has the added benefit of being the same thing in CI and locally.  I do like this but something about that setup is broken with PyPy on Windows, and this works...  Also, using tox-wheel in CI is a bit odd since every tox environment in every job would be rebuilding the wheel rather than building it once and testing that to-be-published artifact everywhere.  So, I think some drawbacks both ways.  Also, not being the same as Twisted is a downside.  Then again, Twisted isn't [testing on PyPy on macOS or Windows](https://github.com/twisted/twisted/pull/1520).

Note that the code in `tox_build.sh` can't be put in `tox.ini` since it uses wildcards and [tox doesn't support those](https://github.com/tox-dev/tox/issues/1571) (but yay, tox 4.1 maybe will).


adiroiban (approved):
It looks good. Thanks.

I guess thare are many ways to do it.

As long as the Ci works, I am happy with any solution pushed by @altendky :)

Thanks!

tox_build.sh
```diff
@@ -0,0 +1,6 @@
+# build the sdist
+python -m build --sdist --outdir ${toxinidir}/dist/ ${toxinidir}
+tar -xvf ${toxinidir}/dist/*
+cd *
+# build the wheel from the sdist
```
gaborbernat, line 5: Why do we need to build the wheel from the sdist, and not build the wheel directly? Feels like you don't need this indirection, you're only testing the correctness of setuptools backend at best.

altendky: It makes sure that the sdist was in the path of what is tested.  Trio does it.  Made enough sense to me to do it too.  I don't know what errors I might be able to have that could make a wheel work but an sdist be broken.  It may well be 'excessive' but it do you see harm that it could do?  The overall goal is to make sure that the actual files to be published are what is tested.  Not just that 'some wheel built in some situation is tested'.  Likewise for sdist.

graingert: https://twistedmatrix.com/trac/ticket/10103 would have been caught if I'd have used this strategy

gaborbernat: If you go down this path why not also test that direct wheel build works? I'm sure they'll be some edge case in some situation when src to sdist works, sdist to wheel also, but src to wheel does not 🤷‍♂️

gaborbernat: And it's not some wheel in some situation. The builder ensures it's a pep 517/518 situation.

altendky: @gaborbernat, I agree that this doesn't cover every scenario.  I also acknowledge that my interest in doing this is in part based on my lack of detailed knowledge of all the build rules.  Then again, if you can easily add a check to compensate for needing to always know and think about such details, there's some value there.

The sequential build slightly slows down CI and it is a tad more complicated.  I should have put a more explanatory comment as to my intent and the (low) level of knowledge of the details I'm 'concerned' about.  Do you have other concerns?

For my education, does `python -m build` fully isolate from any outside influence?  I would think that environment variables could be different, files present in the directory could be different, etc.

gaborbernat: Your build backend is the one responsible to ensure it generates the same build when called from src/sdist/dirty src/etc 👍🏻 I'm just saying that you could also delegate/trust this job to your build the backend. Nothing wrong with doing this, just felt arbitrary to pick out two edge cases.

altendky: They are arbitrary in terms of testing all possible build paths.  I don't think they are arbitrary in terms of testing all to-be-published-to-PyPI artifacts.  There will be an sdist and a wheel.  This flow tests both exact files I will publish to PyPI.  Albeit one (sdist) via the other (wheel).  As @graingert pointed out, this would even have caught a recent bug in the Twisted ecosystem.  (though I am taking their word for it, I haven't dug into that scenario personally)

If we expand from 'files to be published to PyPI' to 'ways to install' then `env/bin/pip install git+https://github.com/twisted/towncrier` is certainly also valid and not presently tested.  I suppose `--editable` would be another. 
 There may be more that we ought to support as well.  Also some that we shouldn't support.  But, 'ways to install' is a bigger scope than I was targeting.

Without shutting it down, thanks for the conversation.  `:]`

gaborbernat: How do you generate your packages to upload to pypi?

altendky: This code that we are discussing.  They are built, archived in GitHub Actions.  Downloaded for tests and checks.  If all is still well, they are downloaded and then published to PyPI.  At least that is the presently-being-developed workflow.  This PR does the first bit, #315 is adding the automated PyPI upload.

gaborbernat: Perhaps you want a wheel-from-sdist option for ``build`` project so you don't have to do this dance 🤷🏻

altendky: Maybe.  We'll see what everyone else thinks I guess.  https://github.com/pypa/build/issues/257

PR #324, opened 2021-03-03
Test on 3.9 in Windows

altendky:
Now that Twisted 21.2.0 is out with 3.9 support, test it fully.


adiroiban (approved):
Hooray. Thanks! This can be merged :)

PR #327, opened 2021-03-15
Remove basepython = python3.6 for checks

altendky:


altendky:
I don't really know why #323 was only ok until it got merged...  Maybe the underlying Linux image changed and lost 3.6?  But, I don't really have a strong need to have these versions pinned locally nor to duplicate them from the GHA config where they are 3.8.

adiroiban:
Looks good. Looking forward for the final review :)

altendky:
Supposedly GHA is back online...

adiroiban (approved):
Thanks. Ready to merge!

PR #328, opened 2021-03-18
testing automatic publishing

altendky:


altendky:
Merged back to #315 now that the fixes and testing are finished.

PR #329, opened 2021-03-19
use incremental to canonicalize versions for release

altendky:


adiroiban (approved):
I preferred to use some pip / stdlib / package version so that we don't have yet another version standard.

But as you are the maintainer of this library, I don't see any critical issue with this change.

I don't think that it's worth spending more time arguing about which version scheme is better.
We only have one life :)

Thanks!

altendky:
I mentioned to @adiroiban that I am not particular attached to incremental.  I just figured the branch etc names ought to match what will show up on PyPI and `pip list` etc.  If we change that then yeah, I might tend back towards PEP440 normalization everywhere.

Issue #309, opened 2020-12-23
Support explicitly selected 'change' and 'release' checks


altendky:
This is maybe a piece of, or duplicate of, or related to, or...  issues #152 and #246 and also PRs #174 and #296.  This issue is being created as a clear place to discuss addition of an explicit control between two types of checks and what those checks would do.  The other issues either describe a problem and a general request.  The PRs (one built from the other) implement weak implicit detection and checks.

Given that the subcommand interface is only available presently in 19.9.0rc1 I think it is ok to change that without deprecation etc (I don't know if the option-based CLI was even deprecated).  @energizah suggested that we consider names relating to what the checks do which seems at the very least worth considering.  I will be referring to them as the change check and the release check for now without implying any decision against the suggestion.  @adiroiban suggested the existing check be documented in the readme first.  I think we should know what it does but I'm not sure it needs to go in the readme if we plan to change it or it's surroundings before the next release?  We should be sure to remember the `-m towncrier` interface as well while considering the design of the explicit check-type selection (I say this because I had kind of forgotten about it).

Checks:
- Existing
  - Run `git diff` against the reference https://github.com/twisted/towncrier/blob/70646d98ce0e31920d6295aa5cb7dcbb7c913e6f/src/towncrier/check.py#L44
  - If no files changed, `SUCCEED` https://github.com/twisted/towncrier/blob/70646d98ce0e31920d6295aa5cb7dcbb7c913e6f/src/towncrier/check.py#L55
  - Identify present fragments that differ from the reference (new or modified or renamed or ...?) https://github.com/twisted/towncrier/blob/70646d98ce0e31920d6295aa5cb7dcbb7c913e6f/src/towncrier/check.py#L89
  - If no fragments differ, `FAIL`
  - If one or more fragments differ, `SUCCEED`
- Change
  - Should it change from `Existing`?
- Release
  - No fragments present
  - Release notes modified
    - Maybe check more carefully?  Presence of expected top line/title given the package version?

This is just an initial quick list, I started into this a bit later than I ought have.

Just to make sure everyone sees this: @adiroiban, @energizah, @freakboy3742, @tjanez, @vcalvert.  Obviously feel free to unsubscribe if you don't want to follow this new issue.

adiroiban:
Thanks @altendky for putting this together

---------

I am -1 on naming based on what it does (functional name)... the check might do many things.

I am +1 for naming it based on when you need to use the check (semantic name) - on a non-release branch vs on the release branch.

---------


[...]

PR #330, opened 2021-03-20
Fixup release instructions

altendky:
Just whatever I catch while doing the releases.

Draft for:
- [ ] Tagging on a branch isn't great when you then squash-merge it (which is the default)
- [ ] Tagging on the branch also results in a double build https://github.com/twisted/towncrier/pull/332#issuecomment-812777833
- [ ] The newsfragment creation and consumption to keep the build happy is silly.  Should note this until #309 (I think?) addresses this.


adiroiban:
Any plan to continue working on this, or we can merge the current changes?

I did some minor changes to CONTRIBUTING.rst while doing the last release.

altendky:
Go for it.  Maybe the checklist in the OP would make sense to add as notes in the instructions until they are addressed?

adiroiban:
> The newsfragment creation and consumption to keep the build happy is silly.

I hope this will be fixed once #337 is merged.

----------

>  Tagging on a branch isn't great when you then squash-merge it (which is the default)

I am not sure what is the issue here.

You tag the code that is released.
It's only for reference.

--------

> Tagging on the branch also results in a double build

I am not sure what is the issue there.

When you create a tag, we run all the checks to make sure we don't release something for which the tests are broken.

This is the strict release process.
I am fine with it.

I would also be fine, that when a tag is created, to only run the release task and don't re-run the tests.

The idea is that for a PR you see the test green and then you manually create the tag/release.

adiroiban:
> I find it a bit weird when a project's last release (or all of them for that matter) do not exist in the history of the main branch.

True. But but this is what you get when play god and rewrite the history :)

The changes were made into small commits, but with a squash you pretend that everything was a single big commit :)

I got used to that.

In practice, if a need to do a backport for an older release, I always create a new branch from tag and for me it's not really an issue.

So.. is this PR still a draft?

Can we merge it? Thanks!

adiroiban (approved):
Ready to land. Thanks!

Issue #313, opened 2021-01-10
Release planning (19.9.0rc1, 19.9.0, and 21.3.0)


altendky:
19.9.0rc1 has been release to PyPI but was not tagged or merged back to `master`.  No 19.9.0 release steps have been taken yet.  Subsequent develop is (maybe?) ready for a ~~21.1.0~~ 21.3.0 release.  So, there are some various existing bits to cleanup and new bits to create.

101d60ea150a8c73ca479f97c2707a3f07e0f61d seems to be the commit released as https://pypi.org/project/towncrier/19.9.0rc1/ and doesn't have either the 19.9.0-rc1 release notes (fixed in https://github.com/twisted/towncrier/commit/280e5b3bbcd84812b6a4a5f159a8194db757b309) nor an (empty) 19.9.0 release notes entry (should be added if releasing 19.9.0).  It also has a formatting error in the readme fixed in 1f396eab7acd6eece8cb3294a9f1d7921fde76cc.  Both linked commits are in the [`hawkowl/release-19.09`](https://github.com/twisted/towncrier/compare/hawkowl/release-19.09) branch but are after the seemingly 19.9.0-rc1 released commit https://github.com/twisted/towncrier/commit/101d60ea150a8c73ca479f97c2707a3f07e0f61d.  A 19.9.0 release ought to include all of those plus an empty release notes section for 19.9.0 itself.

Here are my proposed actions:
- [x] Tag https://github.com/twisted/towncrier/commit/101d60ea150a8c73ca479f97c2707a3f07e0f61d as `19.9.0rc1` to give reference to anyone looking for what https://pypi.org/project/towncrier/19.9.0rc1/ is.
- [x] Close (https://github.com/twisted/towncrier/pull/310) since it has a catch up merge of `master` into [`hawkowl/release-19.09`](https://github.com/twisted/towncrier/compare/hawkowl/release-19.09) so it is not fit for any release.
- [x] Document the release process (#317)
- [x] Implement any release automation associated with the new process (#315)
- [x] Fix release instructions to use incremental to define the textual form of the version (#329)
[...]

PR #331, opened 2021-03-20
Release 19.9.0

altendky:


altendky:
Proposed release artifacts (GitHub requires not-`.whl` so here is a `.zip` of the two).
[towncrier-19.9.0-artifacts.zip](https://github.com/twisted/towncrier/files/6176531/towncrier-19.9.0-artifacts.zip)

Being a special case here we're holding off on publishing until the next release is ready.  Merge this first then publish it with the next release.  See plan and discussion in https://github.com/twisted/towncrier/issues/313.

adiroiban (approved):
Looks good. Thanks!

Please merge and create the tag.

As you mentioned, the PyPi zip will be uploaded after the 21.X.X release.

PR #337, opened 2021-04-06
Skip the checks if the news file was changed

### LINKED ISSUE #152: Running check after building the changelog

### LINKED ISSUE #154: Please document what `towncrier.check` check for and how to use it
The README does not state anything about `towncrier.check`. [NEWS.rst](https://github.com/hawkowl/towncrier/blob/master/NEWS.rst#towncrier-1860-2018-07-05) says:
> Features:
> * `python -m towncrier.check`, which will check a Git branch for the presence of added newsfiles […]

Please add a few lines of documentation:
What exactly does `towncrier.check` check for? What is its intended usage?

### LINKED ISSUE #246: Chicken-and-egg problem with `--check` option

hexagonrecursion:
Closes #152
Closes #246
Closes #154

I wanted to use `towncrier check` as part of pip CI, but this should be fixed first.
This PR makes `towncrier check` succeed for a branch that changes the news file even if it does not add news fragments.


hexagonrecursion:
Why are you testing on python2.7 if the minimum version you officially support is python3.5?
https://github.com/twisted/towncrier#quick-start
> towncrier, as a command line tool, works on Python 3.5+ only.

altendky:
> Why are you testing on python2.7 if the minimum version you officially support is python3.5?
> [twisted/towncrier#quick-start](https://github.com/twisted/towncrier#quick-start)
> 
> > towncrier, as a command line tool, works on Python 3.5+ only.

Looks like the readme got updated at some point without updating the rest.  It still works on 2.7 and I was planning on keeping that around until it becomes problematic.  Correction submitted as https://github.com/twisted/towncrier/pull/338.  Not sure if there should be a 21.3.0post1 to address that on the PyPI page.

altendky:
I think this PR relates to https://github.com/twisted/towncrier/issues/309.  I haven't been thinking about that for awhile.  `:[`

altendky:
FWIW, I do like pathlib.  I'm not going to drop 2.7 just to get it, but there is https://pypi.org/project/pathlib2/ that I might be game for.  Probably simplest though to just use os.path for now though.  Though until we figure out #309, I suspect this PR might get hung up.  Though I hate to have that be the case.

adiroiban:
Thanks Andrey for the nice PR.

I like the idea of using the NEWS file update as a hint that you are on a release branch.
I don't see any issues with that.

---------------

I would go even further and fail in there are news fragment **and** NEWS  file is updated at the same time.... but this can be a separate PR.

----------

I think that this also fixes #154 

Yes. It's related to #309 .

src/towncrier/test/test_check.py
```diff
@@ -148,3 +172,52 @@ def test_none_stdout_encoding_works(self):
 
         self.assertEqual(0, proc.returncode)
         self.assertEqual(b"", stderr)
+
+    def test_first_release(self):
```
adiroiban, line 176: It would be nice to have a docstring explaining what is the purpose of this test.
Together with a docsting from test_second_release they should help understand the differences between these 2 tests.

We don't have tests for testing code and in order to be able to validate that the test is correct we need a docstring explaining the purpose of each test.

src/towncrier/test/test_check.py
```diff
@@ -148,3 +172,52 @@ def test_none_stdout_encoding_works(self):
...
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            # Arrange
+            write("towncrier.toml", "[tool.towncrier]")
+            write("newsfragments/.gitignore", "!.gitignore")
+            initial_commit()
```
adiroiban, line 183: This looks like duplicate code between tests.
Why not reuse the existing `create_project`


```suggestion
            create_project('towncrier.toml')
```

src/towncrier/test/test_check.py
```diff
@@ -148,3 +172,52 @@ def test_none_stdout_encoding_works(self):
...
+            commit("Prepare a release")
+
+            # Act
+            result = runner.invoke(_main, ["--compare-with", "main"])
+
+            # Assert
+            self.assertEqual(0, result.exit_code, (result, result.output))
+            self.assertIn("Checks SKIPPED: news file changes detected", result.output)
+
+    def test_second_release(self):
```
adiroiban, line 199: I don't understand this test. What code path / feature is tested here that is not tested in the first test?

src/towncrier/test/test_check.py
```diff
@@ -148,3 +172,52 @@ def test_none_stdout_encoding_works(self):
...
         self.assertEqual(0, proc.returncode)
         self.assertEqual(b"", stderr)
+
+    def test_first_release(self):
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            # Arrange
+            write("towncrier.toml", "[tool.towncrier]")
+            write("newsfragments/.gitignore", "!.gitignore")
```
adiroiban, line 182: Why we need this ignore in this test?

src/towncrier/test/test_check.py
```diff
@@ -30,6 +31,29 @@ def create_project(pyproject_path):
...
+    if dir:
+        try:
+            os.makedirs(dir)
+        except OSError:
+            pass
+    with open(path, "w") as f:
+        f.write(contents)
+
+
+def initial_commit():
```
adiroiban, line 50: A comment would be nice. It's  not clear why existing `create_project` is not good enough for testing.
If you need a different branch name, I think that you can extend create_project and add an optional argument.

README.rst
```diff
@@ -163,3 +163,20 @@ Furthermore, you can add your own fragment types using:
     directory = "deprecation"
     name = "Deprecations"
     showcontent = true
+
+
+Automatic pull request checks
+-----------------------------
+
+To check if a feature branch is missing a news fragment, run::
```
adiroiban, line 171:
```suggestion
To check if a feature branch adds at least one news fragment, run::
```

README.rst
```diff
@@ -163,3 +163,20 @@ Furthermore, you can add your own fragment types using:
...
+    towncrier check
+
+By default this compares the current branch against ``origin/master``. You can use ``--compare-with`` if the trunk is named differently::
+
+    towncrier check --compare-with origin/main
+
+The check succeeds if at least one of the following is true:
+
+- The current branch adds at least one news fragment
+- The current branch changes the news file (e.g. the PR is a new release)
```
adiroiban, line 182: Maybe something like this.
My fear is that in the future we might modify the check and will forget to update the docs.
So instead of an exhaustive list of checks, maybe just mention that check will do nothing when NEWS is modified.

```suggestion
The check is automatically skipped when the main news file is modified inside the branch as this signals a release branch that is expected to not have news fragments.
```

adiroiban (commented):
Thanks for the PR and the documentation.

I left an initial set of comments.

I would be awesome to all see `check` fails if there are both NEWS changes and newsgragment files present... but not required for this PR.

I am waiting for @altendky to see how he thinks about this PR

src/towncrier/test/test_check.py
hexagonrecursion: Done

src/towncrier/test/test_check.py
hexagonrecursion: I dislike `create_project()`.
* It encapsulates a poorly scoped amount of code. The semantics are unclear. It does too many unrelated things. Changing it like you did in #296 affects every test that uses it which is risky.
* It is inflexible: you either use it or you don't. I prefer my approach: a set of (mostly) orthogonal building blocks that every test can use to arrange things precisely the way that particular test needs to.
* It is a little to implicit for my taste. I actually like that every file the test creates is explicitly named in the test itself.

src/towncrier/test/test_check.py
hexagonrecursion: Good point. We do not.

README.rst
hexagonrecursion: Good catch!

src/towncrier/test/test_check.py
hexagonrecursion: See https://github.com/twisted/towncrier/pull/337/files/9f705991bea4447da623247db2d4ad3a1f831932#r608653671

src/towncrier/test/test_check.py
hexagonrecursion: > Changing it like you did in #296 affects every test that uses it which is risky.

Just to be clear: I'm not criticizing you, I'm criticizing a specific decision you made because I think I know how to solve that specific problem better.

adiroiban: @hexagonrecursion  don't worry :) If you think that some code is not ok. please say it loud.
I will not take it as an offense... I know that I am capable of writing bad code :)

Your comments are valid.

Regarding the risky change. As long as the tests are not failing, I think that any change should be done if we thing that it improves the readability and usability.
We should be confident that the test suite is solid and don't code with fear :)

If you think that `create_project` is bad pattern, it's best to add a docstring comment to it and let other developers know what is the current best practice.

But I think that all test helpers should have short docstrings describing when and why they should be used :)

src/towncrier/test/test_check.py
```diff
@@ -148,3 +172,54 @@ def test_none_stdout_encoding_works(self):
...
+            self.assertIn("Checks SKIPPED: news file changes detected", result.output)
+
+    def test_second_release(self):
+        """The checks should be skipped on a branch that modifies the news file."""
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
+            # Arrange
+            write("towncrier.toml", "[tool.towncrier]")
+            write("newsfragments/.gitignore", "!.gitignore")
```
adiroiban, line 207: Do we need this ignore here?

adiroiban:
A brief summary of an IRC chat with Kyle.

My preference is to have a separate `towncrier check-release`that is explicitly triggered for a release branch.
It will check that NEWS is updated and no fragments are left in the branch.
In this way, the normal `towncrier check` will fail if you are not on a release branch and you accidentally modify the news file without create a news fragment.

I understand that having a modified NEWS file means 99.99% that you are on a release branch and this is why I think that this behavior is also. 

At the same time, by just using `check` you get a simple solution that just works and a modified NEWS file is a good heuristic for a release branch (99.99%).

Later we can add `towncrier check --strict` if required.

I am waiting for @altendky to weight his opinion on this :)

src/towncrier/test/test_check.py
hexagonrecursion: > I know that I am capable of writing bad code

Same here. I also prefer to err on the side of not offending anybody.

> If you think that create_project is bad pattern, it's best to add a docstring comment to it and let other developers know what is the current best practice.

If you are onboard with my argument I'll submit a followup PR to refactor the rest of the file and delete `create_project()`

> But I think that all test helpers should have short docstrings describing when and why they should be used :)

:+1:

hexagonrecursion:
codecov complains that a line in the `write()` helper is not covered. Should I write a test for the helper?

adiroiban:
I would say `no`... and just add a pragma coverage comment, and enable them in .coveragerc

altendky:
@hexagonrecursion, my apologies for not getting back to you on this sooner.  I am with Adi on this one in terms of preferring an explicit selection of the mode of the check.  That said, I think we can start with the implicit command and add the explicit later whenever those of us that want it find the time to do so.  We already have `towncrier check`.  In a comprehensive solution we could have subcommands `towncrier check auto`, `release`, and `non-release` (or whatever names, I don't mean to suggest these are the best names).

I am not convinced that 'the newsfile has been edited' is a great heuristic.  There have been a number of times where I have edited a newsfile to fix it.  If we can make this a provisional feature then I'll be less picky, but I'm not sure that is plausible as the default behavior for an existing command.  That leads to either it being configurable or a dedicated command.  It doesn't seem like it should be too problematic to introduce this as a new feature while leaving the existing behavior intact.

My gut, at the moment, says to add two subcommands (existing behavior as default, other subcommand being the new behavior).  We can add the explicit release check command separately, or here, no real matter.  Then, when another heuristic is developed, we can add a configuration setting for which one to use with news-file-modified being the default.  I think that this will be a small amount of added work, will retain backwards compatibility, and will enable future expansion.  Does anyone agree?

To avoid dumping more work on you, @hexagonrecursion, I'll volunteer to rework the commands either before or after merging this, if you prefer.

I hope this comes across as appreciation for your work and an effort to figure out how to make sure we put it to the best use for all.  When I released 21.3.0 it was actually missing one of...  I think two (?) things that I had set out to include initially.  This was one of them.  So, I'm looking forward to getting this figured out.  Thank you.  `:]`

adiroiban:
doing a release for towncrier itself using `tox -e check-newsfragment` is a pain.


so, even if this doesn't solve all the problems, I think that it should be updated to pass all tests and then merged.

hexagonrecursion:
> Regarding the risky change. As long as the tests are not failing, I think that any change should be done if we thing that it improves the readability and usability.
> We should be confident that the test suite is solid and don't code with fear :)

The tests are our safety net when we refactor code, but we have **no** such luxury when we refactor the tests themselves. Changing `create_project()` might lead to tests no longer testing what they intended to test and possibly passing when they should fail.

We **should** refactor tests, but the refactoring should usually be limited to one of two types:
1. True refactoring - the implementation is changed, but the observable behavior stays the same
2. Refactoring of a specific test that makes sure to preserve **intention** of the test

A "refactoring" that changes the setup used in many tests requires an audit of every affected test to make sure it still tests the behavior it intended to test

src/towncrier/test/test_check.py
```diff
@@ -190,51 +196,68 @@ def test_none_stdout_encoding_works(self):
...
+            call(["git", "checkout", "main"])
+            call(["git", "merge", "otherbranch", "-m", "Sync release in main branch."])
+
+            # We have a new feature branch that has a news fragment that
+            # will be merged to the main branch.
+            call(["git", "checkout", "-b", "new-feature-branch"])
+            write("foo/newsfragments/456.feature", "Foo the bar")
+            commit("A feature in the second release.")
+            call(["git", "checkout", "main"])
+            call(["git", "merge", "new-feature-branch", "-m", "Sync release in main branch."])
```
hexagonrecursion, line 252: > "Sync release in main branch."

should probable say "Merge new-feature-branch" or "Merge: Foo the bar" because we are merging a feature branch rather than a release branch

hexagonrecursion:
> Fix the test.

There is nothing in the commit message explaining (I) what was broken (II) how did your change fix it

src/towncrier/test/test_check.py
```diff
@@ -190,51 +196,68 @@ def test_none_stdout_encoding_works(self):
...
 
-        If the checkes are not skipped in this case, towncrier check would fail
-        for the first release that has a changelog"""
+        If the checks are not skipped in this case, towncrier check would fail
+        for the first release that has a changelog.
+        """
         runner = CliRunner()
 
         with runner.isolated_filesystem():
-            # Arrange
```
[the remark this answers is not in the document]
adiroiban: I will add `Arrange` back. Thanks!

adiroiban:
> The tests are our safety net when we refactor code, but we have no such luxury when we refactor the tests themselves.

I think that we should take into consideration that this is a hobby project,
for which people are contributing in their free time.

Also, we are talking here about `towncrier` which is a tool to combine a set of text files into a single one.
It's also a tool that is used only by the developers and not triggered by end users.
Is also not a tool distributed to end users.
I don't think that is OK to apply the same set of development requirements for `towncrier` as a software that will drive a robot on Mars.

I think that we are OK to go with a more relaxed approach.

I am not saying that we should ignore all the best practices  :)

> There is nothing in the commit message explaining (I) what was broken (II) how did your change fix it. Also: the refactoring should have been a separate commit before or after the fix to make it more obvious which changes are a part of the fix

True. I will try to separate the changes for the next PR.

src/towncrier/test/test_check.py
```diff
@@ -190,51 +196,68 @@ def test_none_stdout_encoding_works(self):
...
-            write("towncrier.toml", "[tool.towncrier]")
-            write("newsfragments/123.feature", "Foo the bar")
-            initial_commit()
+            create_project()
+            # Before any release, the NEWS file might no exist.
+            self.assertNotIn('NEWS.rst', os.listdir())
 
-            call(["git", "checkout", "-b", "next-resease"])
```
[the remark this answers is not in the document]
adiroiban: The intention is to try to reuse the `create_project` so that the tests will use the same same set of conventions when managing the news files. 

So, to increase consistency.

I am not happy with `create_project`, but the idea is that when you work on these tests, you understand what `create_project` does once, and then you should have a good idea of what to expect in each test.

adiroiban:
Hi @hexagonrecursion Many thanks for the review and sorry for the mess.

If you think that the last refactoring is that bad, I am happy to have it reverted and have the tests fixed in another way.

Cheers

hexagonrecursion:
It could better, but it is OK. I am *not* against merging it.

adiroiban:
Many thanks for the excellent review and your support for improving the quality of the project.

I think that we can at any time have a new PR to improve the current testing... but time is short and I think that for towncrier "done is better than perfect".

I did the last release of towncrier, and without this feature, the release process was a bit of a PITA :) 

Happy to see this merge and hope it will improve the next release... and I plan to release soon :)

PR #338, opened 2021-04-06
Update readme to list support of 2.7 and 3.5+

altendky:


adiroiban (approved):
Thanks. Looks good!

Happy to see support for 2.7

PR #341, opened 2021-04-14
Provide a --version option

### LINKED ISSUE #339: Consider a --version option
# Situation

The `towncrier` script doesn't know `--version`.

# Proposed solution

```
$ towncrier --version
[... outputs the current version ...]
```

Should be fairly easy. :wink:

tomschr:
This PR fixes #339 and adds a `--version` option. This makes it possible to run the command like this:

```
$ towncrier --version
towncrier, version 21.3.1.dev0
```

### DIFF
--- a/src/towncrier/_shell.py
+++ b/src/towncrier/_shell.py
@@ -11,9 +11,11 @@ from click_default_group import DefaultGroup
 from .build import _main as _build_cmd
 from .check import _main as _check_cmd
 from .create import _main as _create_cmd
+from ._version import __version__
 
 
 @click.group(cls=DefaultGroup, default="build", default_if_no_args=True)
+@click.version_option(__version__.public())
 def cli():
     pass
 
--- a/src/towncrier/newsfragments/339.feature.rst
+++ b/src/towncrier/newsfragments/339.feature.rst
@@ -1,0 +1,1 @@
+towncrier --version` was added to the command line interface to show the product version.
--- a/src/towncrier/test/test_project.py
+++ b/src/towncrier/test/test_project.py
@@ -116,6 +116,13 @@ class InvocationTests(TestCase):
             os.makedirs("news")
             out = check_output([sys.executable, "-m", "towncrier", "--help"])
             self.assertIn(b"[OPTIONS] COMMAND [ARGS]...", out)
-            self.assertIn(b"--help  Show this message and exit.", out)
+            self.assertRegex(out, br".*--help\s+Show this message and exit.")
         finally:
             os.chdir(orig_dir)
+
+    def test_version(self):
+        """
+        `--version` command line option is available to show the current production version.
+        """
+        out = check_output(["towncrier", "--version"])
+        self.assertTrue(out.startswith(b"towncrier, version 2"))


tomschr:
If I'm not mistaken, the failures looks like they are not directly related to my changes. I could hunt down two issues.

The first one was about some space issues:

```diff
@@ -116,6 +116,13 @@ class InvocationTests(TestCase):
             os.makedirs("news")
             out = check_output([sys.executable, "-m", "towncrier", "--help"])
             self.assertIn(b"[OPTIONS] COMMAND [ARGS]...", out)
-            self.assertIn(b"--help  Show this message and exit.", out)
+            self.assertIn(b"--help     Show this message and exit.", out)
```

To my understanding, if I introduce just a simple option (`--version` in this case) it should NOT influence this test. If it does, the test is broken. :wink:

Shouldn't we write the above test in such a way that spaces aren't really matter? Isn't it possible to match against some regex like `--help\s+Show this message and exit.`?

The other issue was in `towncrier.test.test_build.TestCli.test_command`. I have no idea why it fails when I introduce the option `--version`.

Hope someone with more knowledge than I could look into this. Thanks! :+1:

adiroiban:
> If I'm not mistaken, the failures looks like they are not directly related to my changes. I could hunt down two issues.

I don't know. The tests were executed 7 days ago and they were green  https://github.com/twisted/towncrier/runs/2288463008

I am manually triggering a re-run on the main trunk branch. If that run is green, then the issue is with this PR.


> Shouldn't we write the above test in such a way that spaces donn't really matter? Isn't it possible to match against some regex like `--help\s+Show this message and exit.`?

Yes. One option is to use [assertRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex)

adiroiban:
It looks like the failures are from this branch as trunk is green  https://github.com/twisted/towncrier/runs/2341993285

@tomschr are the test green when you execute them on your local dev system? Are the trunk tests also green on your local dev?

Trunk is green on my local Ubuntu

```
(build3) ✔  ~/dev/towncrier [master|✔ ] 
$ trial towncrier

...

PASSED (successes=68)

```

I am checking your branch now

adiroiban:
The issue is `TypeError('cli() takes 0 positional arguments but 1 was given')` ... found on your branch via `$ trial -b towncrier`

src/towncrier/_shell.py
```diff
@@ -11,9 +11,12 @@
 from .build import _main as _build_cmd
 from .check import _main as _check_cmd
 from .create import _main as _create_cmd
+from ._version import __version__
 
 
 @click.group(cls=DefaultGroup, default="build", default_if_no_args=True)
+@click.version_option(__version__.public())
+@click.pass_context
```
adiroiban, line 19: Is this needed? I think that this is why the test fails.

I see the tests are green even if this is removed

adiroiban (changes requested):
Thanks for the PR.

I think that it looks good.... but it needs some cleanup.

Do we need `@click.pass_context`?
I have removed it and  `test_version` was still green.

Would be nice to use `assertRegex` instead of `assertIn` but I am also ok with your quick fix for `assertIn` :)

Thanks!

src/towncrier/_shell.py
tomschr: Ahh, when I remove this line it works! :+1:

tomschr:
Hi @adiroiban,
thanks for the feedback, much appreciated! :+1: I've added your suggestions and on my machine it works and all is green. I've used the `assertRegex` part.

Should I squash the commits or are you doing it? :slightly_smiling_face: 

Thanks again!

adiroiban:
You can squash the commit and don't forget to add a news fragment inside `src/towncrier/newsfragments/` for towncrier itself :)t

It can be `src/towncrier/newsfragments/339.feature` with content like

```
`towncrier --version` was added to the command line interface for showing the product version.
```

tomschr:
Seems this time I was successful. :smile: Adi, anything else I should add, correct etc.?

src/towncrier/newsfragments/339.feature.rst
```diff
@@ -0,0 +1 @@
+`towncrier --version` was added to the command line interface for showing the product version.
```
adiroiban, line 1:
```suggestion
towncrier --version` was added to the command line interface to show the product version.
```

src/towncrier/test/test_project.py
```diff
@@ -116,6 +116,13 @@ def test_dash_m(self):
...
             out = check_output([sys.executable, "-m", "towncrier", "--help"])
             self.assertIn(b"[OPTIONS] COMMAND [ARGS]...", out)
-            self.assertIn(b"--help  Show this message and exit.", out)
+            self.assertRegex(out, br".*--help\s+Show this message and exit.")
         finally:
             os.chdir(orig_dir)
+
+    def test_version(self):
+        """
+        Invokes `python -m towncrier --version`
```
adiroiban, line 125: Maybe something like this

```suggestion
        `--version` command line option is available to show the current production version.
```

https://jml.io/pages/test-docstrings.html

src/towncrier/test/test_project.py
```diff
@@ -116,6 +116,13 @@ def test_dash_m(self):
...
+        """
+        out = check_output([sys.executable, "-m", "towncrier", "--version"])
+        self.assertIn(b"version", out)
```
adiroiban, line 128: Maybe something like this

```suggestion
        self.assertTrue(out.startswith(b", version 2"))
```

The problem with the current test is that if I use `--help` instead of version, the test is still green

```diff
diff --git a/src/towncrier/test/test_project.py b/src/towncrier/test/test_project.py
index 271b33d..9f3f0ef 100644
--- a/src/towncrier/test/test_project.py
+++ b/src/towncrier/test/test_project.py
@@ -124,5 +124,5 @@ class InvocationTests(TestCase):
         """
         Invokes `python -m towncrier --version`
         """
-        out = check_output([sys.executable, "-m", "towncrier", "--version"])
+        out = check_output([sys.executable, "-m", "towncrier", "--help"])
```

adiroiban (approved):
Look good. Only a few minor comments :) Thanks!

src/towncrier/test/test_project.py
tomschr: Oh you are absolutely right! Good point! :+1:

tomschr: What about this:

```python
self.assertRegex(out, br"^towncrier..version \d\d.*$")
```

I've tested it with `--version` (succeeds) and `--help` (fails as expected). Maybe we don't need to test exactly what version it prints, but just that it's a specific format (that starts with two numbers).
That way it would be flexible enough for any version change.

tomschr: If you prefer to really match the version, we could use this alternative:

```python
def test_version(self):
        """
        `--version` command line option is available to show the current production version.
        """
        from .._version import __version__
        version = __version__.public()
        pattern = rf"^[\w\.]+..version {version}$".encode()
        out = check_output([sys.executable, "-m", "towncrier", "--version"])
        self.assertRegex(out, pattern)
```

It behaves as expected with `--version` and `--help`.

adiroiban: i think that the current is OK. We are using click standard code...so we just need a way to check that it is there.

altendky:
Thanks to you both.  `:]`

seberg:
I think this PR broke our CI at NumPy, because `--version` is a valid option to overwrite the "version" identifier in the release notes.  (Still using the dev version, although I think there was probably a new enough release by now)

This is mainly to let you know that this breaks (maybe add a release note).  I can fix it by using `towncrier build` to make it explicit that we are building.

adiroiban:
Thanks @seberg for the report. Much appreciated and it's an important side effect.

I guess that in order to not break backward compatibility we can add `towncrier version` 

But maybe we should should just go with towncrier subcommands and in this case we need to update the release note and inform about this backward incompatible change.

altendky:
The default subcommands are nice for backwards compatibility, but maybe when we add them the implicit usage should be deprecated and later removed.  I'm not used to using default subcommands so definitely thanks for filling that gap in my thinking and reporting this @seberg.

Issue #349, opened 2021-05-28
Add support for date objects and not strings in jinja templates


kinkerl:
Hi,
i would like to format the dates which are used in the jinja2 templates. The dates are currently strings and just providing them as date objects to the template rendering would allow users to customize the date output. 

I will start to work on a PR for this, maybe we can get it into a state to merge it upstream at some point. 

Cheers

PR #350, opened 2021-05-28
initial work to use date type objects for the date and not strings

kinkerl:
Hi,
this is my first change to support better date objects during template rendering. 
This PR aims to solve #349.

Cheers

### DIFF
--- a/src/towncrier/templates/default.rst
+++ b/src/towncrier/templates/default.rst
@@ -3,10 +3,10 @@
 {{ top_underline * ((top_line)|length)}}
 {% elif versiondata.name %}
 {{ versiondata.name }} {{ versiondata.version }} ({{ versiondata.date }})
-{{ top_underline * ((versiondata.name + versiondata.version + versiondata.date)|length + 4)}}
+{{ top_underline * ((versiondata.name + versiondata.version)|length + 4)}}{{ top_underline * (versiondata.date|title|length)}}
 {% else %}
 {{ versiondata.version }} ({{ versiondata.date }})
-{{ top_underline * ((versiondata.version + versiondata.date)|length + 3)}}
+{{ top_underline * (versiondata.version|length + 3)}}{{ top_underline * (versiondata.date|title|length)}}
 {% endif %}
 {% for section, _ in sections.items() %}
 {% set underline = underlines[0] %}{% if section %}{{section}}
--- a/src/towncrier/templates/single-file-no-bullets.rst
+++ b/src/towncrier/templates/single-file-no-bullets.rst
@@ -1,9 +1,9 @@
 {% if versiondata.name %}
 {{ versiondata.name }} {{ versiondata.version }} ({{ versiondata.date }})
-{{ top_underline * ((versiondata.name + versiondata.version + versiondata.date)|length + 4)}}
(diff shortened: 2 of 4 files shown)


src/towncrier/newsfragments/349.enhancement
```diff
@@ -0,0 +1 @@
+Changed type of `versiondata.date` in the main template from `string` to `date`. This allows users to customize the date rendering in the template with `strftime`, like so: `{{versiondata.date.strftime("%B")}}`.
```
adiroiban, line 1: Minor comment.

I think is best when this text is written in the style of a public release note and not as a commit message :)

```suggestion
The `versiondata.date` is not `date` instance. This allows users to customize the date rendering in the template with `strftime`, like so: `{{versiondata.date.strftime("%B")}}`.
```

adiroiban:
Hi @kinkerl 

Sorry for the delay.
I am trying to do emergency-mode maintenance for this project.

I think the feature looks good.

I think that jinja defaults to `str` and default `str` for `date` is `isoformat`

So we should be ok for backward compatibility.

```python
>>> from datetime import date
>>> ca  = date.today()
>>> str(ca)
'2022-02-13'
>>> ca.isoformat()
'2022-02-13'
>>> ca.strftime("%B")
```

I few questions/comments for this:

* can you add an automated test ... this should make it clear that the date type is part of the public API and should prevent any future regressions.
* It would be nice to have some soft of docs for this feature...but I guess that this is not required right now as we don't have any other "template design" documentation.

kinkerl:
hey @adiroiban 

I made the changes and added first tests. The tests could be adapted once I figure out how to actually run them properly locally. 

I am not happy about the required template changes. Any input (see my comments)?

adiroiban (changes requested):
Thanks for the update and for moving this forward.

This looks good...but I have a few comments regarding the extend of these changes.

It is not clear why the default template needs to be changed.

The change might be needed and there might a good reason behind that change. but i don't see any info about this in the description of this PR.

Also, I think that we should focus the test for datetime to only check the datetime rendering and don't worry about the other rendered objects... we have plenty of tests for the full template rendering.

adiroiban:
It looks like the automated tests are failing.

We  need a green CI run before this can be merged.

I hope that by reverting the default template format and simplifying the new test, this PR will be green.

Thanks!

jaraco:
@kinkerl Just checking in on this issue. Do you have plans to revisit the concerns above?

kinkerl:
i totally lost track of this but i can check them out again next week!

src/towncrier/build.py
```diff
@@ -135,7 +132,9 @@ def __main(
                 project_name = ""
 
     if project_date is None:
-        project_date = _get_date().strip()
+        project_date = date.today()
+    else:
+        project_date = parse(project_date).date()
```
adiroiban, line 137: Just a minor comment

Can we just support ISO date format and then use stdlib `datetime.strptime` to parse the date ?

Not a big deal... just a shame that we dont't have "smart" parse function in stdlib.

At the same time, I think that we should encourage standard date representation, and only support ISO format.
... and it this way, we don't need to add yet another dependency

PR #354, opened 2021-07-02
swap toml with tomli

graingert:
pip is moving from toml to tomli https://github.com/pypa/pip/pull/10035 towncrier should consider following suit

* pip: https://github.com/pypa/pip/pull/10035
* build: https://github.com/pypa/build/issues/308
* black: https://github.com/psf/black/pull/2301
* setuptools_scm https://github.com/pypa/setuptools_scm/pull/586


src/towncrier/_settings.py
```diff
@@ -69,8 +69,8 @@ def load_config(directory):
 
 def load_config_from_file(directory, config_file):
 
-    with open(config_file, "r") as conffile:
-        config = toml.load(conffile)
+    with open(config_file, "r", encoding="utf8") as conffile:
```
graingert, line 72: >  A TOML file must be a valid UTF-8 encoded Unicode document.

twm:
It looks like you'll need to drop Python 3.5 support too.

altendky:
What's the benefit of this switch to justify dropping <3.6?  FWIW, here's my take on supporting old versions.  https://github.com/pytest-dev/pytest-twisted/tree/5dc4efc5d335da0172fec02e48076aacef4bf75d#python-2-support-plans

graingert:
> What's the benefit of this switch to justify dropping <3.6? FWIW, here's my take on supporting old versions. https://github.com/pytest-dev/pytest-twisted/tree/5dc4efc5d335da0172fec02e48076aacef4bf75d#python-2-support-plans

toml doesn't support toml v1 syntax

hukkin:
As graingert mentioned, this will avoid issues like [this Black issue](https://github.com/psf/black/issues/2280), i.e. crashing hard if another tool uses TOML v1 syntax in the shared pyproject.toml.

Note that by setting `python_requires='>=3.6'` metadata in setup.py (currently not set), many legacy Python users will not notice Python 2 being dropped, as pip is smart enough to not install unsupported versions.

setup.py
```diff
@@ -41,8 +40,9 @@
         "Programming Language :: Python :: Implementation :: PyPy",
     ],
     use_incremental=True,
-    install_requires=["click", "click-default-group", "incremental", "jinja2", "setuptools", "toml"],
+    install_requires=["click", "click-default-group", "incremental", "jinja2", "setuptools", "tomli"],
     extras_require={"dev": ["packaging"]},
+    requires_python=">=3.5",
```
hukkin, line 45:
```suggestion
    python_requires=">=3.6",
```

altendky:
Well, yeah, that'll be a relevant reason...  `:|`  Thanks for the explanation.  After getting the last release out I've more or less disappeared so I've got no real say in process here at the moment, but is there any chance we could get another release out with whatever bugfixes before dropping 2.7 etc?

graingert:
> Well, yeah, that'll be a relevant reason... `:|` Thanks for the explanation. After getting the last release out I've more or less disappeared so I've got no real say in process here at the moment, but is there any chance we could get another release out with whatever bugfixes before dropping 2.7 etc?

Yeah would be good to get another release

graingert:
Could make this a conditional dep and drag 2.7 support on for a little while longer

setup.py
```diff
@@ -41,7 +41,15 @@
...
     use_incremental=True,
-    install_requires=["click", "click-default-group", "incremental", "jinja2", "setuptools", "toml"],
+    install_requires=[
+        "click",
+        "click-default-group",
+        "incremental",
+        "jinja2",
+        "setuptools",
+        "toml; python_version < '3.6'",
+        "tomli; python_version > '3.5'",
```
graingert, line 51:
```suggestion
        "tomli; python_version > '3.6'",
```

setup.py
```diff
@@ -41,7 +41,15 @@
...
     use_incremental=True,
-    install_requires=["click", "click-default-group", "incremental", "jinja2", "setuptools", "toml"],
+    install_requires=[
+        "click",
+        "click-default-group",
+        "incremental",
+        "jinja2",
+        "setuptools",
+        "toml; python_version < '3.6'",
+        "tomli; python_version > '3.6'",
```
graingert, line 51:
```suggestion
        "tomli; python_version >= '3.6'",
```

src/towncrier/_settings.py
```diff
@@ -1,9 +1,15 @@
 # Copyright (c) Amber Brown, 2015
 # See LICENSE for details.
 
+import io
 import os
+import sys
 import pkg_resources
-import toml
+
+if sys.version_info > (3, 5):
```
graingert, line 9:
```suggestion
if sys.version_info >= (3, 6):
```

src/towncrier/_settings.py
graingert:
```suggestion
    with io.open(config_file, "r", encoding="utf8") as conffile:
```

graingert:
@altendky the conditional import isn't so bad py2.7 can survive another day

altendky (approved):
I appreciate the accommodation.

PR #357, opened 2021-07-22
Handle deletion of uncommitted news fragments

abompard:
Before this commit, all the news fragments needed to be committed into git, or the fragments removal after building the news file would crash.

In my workflow, I add missing fragments before building the news file because I'm extracting author names from the git log, and towncrier crashes at the end of the build process.

### DIFF
--- a/src/towncrier/_git.py
+++ b/src/towncrier/_git.py
@@ -5,12 +5,27 @@ from __future__ import annotations
 
 import os
 
-from subprocess import STDOUT, call, check_output
+from subprocess import STDOUT, CalledProcessError, call, check_output
 
 
 def remove_files(fragment_filenames: list[str]) -> None:
-    if fragment_filenames:
-        call(["git", "rm", "--quiet"] + fragment_filenames)
+    if not fragment_filenames:
+        return
+
+    # Filter out files that are unknown to git
+    try:
+        git_fragments = check_output(
+            ["git", "ls-files"] + fragment_filenames, encoding="utf-8"
+        ).split("\n")
+    except CalledProcessError:
+        # we may not be in a git repository
+        git_fragments = []
+
+    git_fragments = [os.path.abspath(f) for f in git_fragments if os.path.isfile(f)]
+    call(["git", "rm", "--quiet", "--force"] + git_fragments)
+    unknown_fragments = set(fragment_filenames) - set(git_fragments)
+    for unknown_fragment in unknown_fragments:
+        os.remove(unknown_fragment)
 
 
 def stage_newsfile(directory: str, filename: str) -> None:
--- a/src/towncrier/newsfragments/357.feature.rst
+++ b/src/towncrier/newsfragments/357.feature.rst
@@ -1,0 +1,1 @@
+``towncrier build`` now handles removing news fragments which are not part of the git repository. For example, uncommitted or unstaged files.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -6,6 +6,7 @@ import tempfile
 
 from datetime import date
 from pathlib import Path
+from subprocess import call
 from textwrap import dedent
 from unittest.mock import patch
 
@@ -747,8 +748,8 @@ class TestCli(TestCase):
                     "--yes",
                 ],
             )
-            # not git repository, manually remove fragment file
-            Path(f"newsfragments/{fragment_file}").unlink()
+            # Fragment files unknown to git are removed even without a git repo
+            assert not Path(f"newsfragments/{fragment_file}").exists()
             return result
 
         results = []
@@ -845,8 +846,8 @@ class TestCli(TestCase):
                 ],
                 catch_exceptions=False,
             )
-            # not git repository, manually remove fragment file
-            Path(f"newsfragments/{fragment_file}").unlink()
+            # Fragment files unknown to git are removed even without a git repo
+            assert not Path(f"newsfragments/{fragment_file}").exists()
             return result
 
         results = []
@@ -1530,3 +1531,55 @@ class TestCli(TestCase):
 
         self.assertEqual(0, result.exit_code, result.output)
         self.assertEqual(expected_output, result.output)
+
+    @with_git_project()
+    def test_uncommitted_files(self, runner, commit):
+        """
+        At build time, it will delete any fragment file regardless of its stage,
+        included files that are not part of the git reporsitory,
+        or are just staged or modified.
+        """
+        # 123 is committed, 124 is modified, 125 is just added, 126 is unknown
+
+        with open("foo/newsfragments/123.feature", "w") as f:
+            f.write("Adds levitation. File committed.")
+        with open("foo/newsfragments/124.feature", "w") as f:
+            f.write("Extends levitation. File modified in Git.")
+
+        commit()
+
+        with open("foo/newsfragments/125.feature", "w") as f:
+            f.write("Baz levitation. Staged file.")
+        with open("foo/newsfragments/126.feature", "w") as f:
+            f.write("Fix (literal) crash. File unknown to Git.")
+
+        with open("foo/newsfragments/124.feature", "a") as f:
+            f.write(" Extended for an hour.")
+        call(["git", "add", "foo/newsfragments/125.feature"])
+
+        result = runner.invoke(_main, ["--date", "01-01-2001", "--yes"])
+
+        self.assertEqual(0, result.exit_code)
+        for fragment in ("123", "124", "125", "126"):
+            self.assertFalse(os.path.isfile(f"foo/newsfragments/{fragment}.feature"))
+
+        path = "NEWS.rst"
+        self.assertTrue(os.path.isfile(path))
+        news_contents = open(path).read()
+        self.assertEqual(
+            news_contents,
+            dedent(
+                """\
+                Foo 1.2.3 (01-01-2001)
+                ======================
+
+                Features
+                --------
+
+                - Adds levitation. File committed. (#123)
+                - Extends levitation. File modified in Git. Extended for an hour. (#124)
+                - Baz levitation. Staged file. (#125)
+                - Fix (literal) crash. File unknown to Git. (#126)
+                """
+            ),
+        )
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -28,8 +28,9 @@ If there are no news fragments (including an empty fragments directory or a
 non-existent directory), a notice of "no significant changes" will be added to
 the news file.
 
-By default, the processed news fragments are removed using ``git``, which will
-also remove the fragments directory if now empty.
+By default, the processed news fragments are removed. For any fragments
+committed in your git repository, git rm will be used (which will also remove
+the fragments directory if now empty).
 
 .. option:: --draft
 


src/towncrier/newsfragments/357.feature.rst
```diff
@@ -0,0 +1 @@
+Handle deletion of uncommitted news fragments
```
adiroiban, line 1: Maybe something like this.

```suggestion
It can now handle news fragments which are not part of the git repository. For example uncommited or unstagged files.
```

src/towncrier/test/test_build.py
```diff
@@ -836,3 +836,81 @@ def test_start_string(self):
         """)
 
         self.assertEqual(expected_output, output)
+
+    def test_uncommitted_files(self):
+        runner = CliRunner()
```
adiroiban, line 841: It would be nice to have a dostring describing what feature this test tries to cover.

```suggestion

    def test_uncommitted_files(self):
        """
        At build time, it will delete any fragment file regardless of its stage,
        included files that are not part of the git reporsitory,
        or are just staged or modified.
        """
        runner = CliRunner()
```

src/towncrier/test/test_build.py
```diff
@@ -836,3 +836,81 @@ def test_start_string(self):
...
+                    --------
+
+                    - Adds levitation (#123)
+                    - Extends levitation for an hour (#124)
+                    - Baz levitation (#125)
+                    - Fix (literal) crash (#126)
+                    """
+                ),
+            )
+            self.assertEqual(
```
adiroiban, line 898: do we need this assertion?

I am ok with it... but it can cause future failure when we update part of the output.

src/towncrier/_git.py
```diff
@@ -16,11 +28,23 @@ def remove_files(fragment_filenames, answer_yes):
     else:
         click.echo("I want to remove the following files:")
 
-    for filename in fragment_filenames:
+    for filename in sorted(fragment_filenames):
         click.echo(filename)
 
+    # Filter out files that are unknown to git
+    try:
+        known_fragments = run(
```
adiroiban, line 36: Maybe rename to something like this to hint that these are the files handled by git.

```suggestion
        git_fragments = run(
```

And shouldn't this be moved inside `if answer_yes` , to hint that this code is only needed if we need to delete the files.

And maybe we can resolve the path to absolute so that we don't have 2 variables.
git should be able to handle absolute paths.

src/towncrier/_git.py
```diff
@@ -1,12 +1,24 @@
...
-
 import os
+import sys
+from subprocess import call, check_output, CalledProcessError
+
 import click
 
 
+def run(args, **kwargs):
+    # Use UTF-8 both when sys.stdout does not have .encoding (Python 2.7) and
```
adiroiban, line 12: Maybe have the comment as docstring ? 

And this is not really related to the git code.

Maybe keep is as the private `check._run` and update the private `_run` to always return decoded output.

Make sense?

src/towncrier/check.py
```diff
@@ -8,15 +8,11 @@
...
 import click
 
-from subprocess import CalledProcessError, check_output, STDOUT
+from subprocess import CalledProcessError, STDOUT
 
 from ._settings import load_config_from_options
 from ._builder import find_fragments
-
-
-def _run(args, **kwargs):
```
adiroiban, line 17: I would prefer to keep the private `_run` and refactor it to always return decoded output and add a docstring that explains what it does.

adiroiban (changes requested):
Many thanks for the changes.

They look good.

I left a few minor comments regarding the documentation/docstrings of these changes.

I would like to keep the `_run` helper private and outside of the git code :)

Sorry for the late review.

I hope you will still have time to drive this PR.

Let me know if you are no longer interested in working on it and I will try to take over and finalize this PR.

Cheers

src/towncrier/test/test_build.py
adiroiban: And I think that this assertion is cause failures on Windows...as the path has a different separator.

jaraco:
It looks like this issue has gotten a little stale, and since we haven't heard back from the contributor, I'm going to close it for now. If you'd like to revive the effort, feel free to ping this PR and we can re-open, or start a new one. Thanks.

abompard:
Hey folks! Sorry I had totally forgotten about this PR. I'll try to find some time to finalize it, if it's still relevant.

src/towncrier/_git.py
abompard: Yeah, but I need to call this function from the `remove_files` function, and it feels a bit weird to import a private method from the `check` module (which is a CLI module).
How would you do it?

abompard:
Could you please reopen @jaraco ?

src/towncrier/_git.py
```diff
@@ -1,12 +1,27 @@
...
+import sys
+from subprocess import call, check_output, CalledProcessError
+
 import click
 
 
+def run(args, **kwargs):
+    """Run a command.
+
+    Use UTF-8 both when sys.stdout does not have .encoding (Python 2.7) and
```
adiroiban: do we still support python 2.7 ?

I think that this comment, and maybe the implementation should be updated for py3 only.

adiroiban:
Thanks for the update.

Can you please fix the conflicts and update the docstring for `run` to describe it's current behaviour

We no longer support python 2.7

abompard:
I'm more than happy to remove Python 2.7 remnants :-D
Alright, it's updated and tests should pass now.

src/towncrier/_git.py
```diff
@@ -5,12 +5,26 @@
...
-        call(["git", "rm", "--quiet"] + fragment_filenames)
+    if not fragment_filenames:
+        return
+
+    # Filter out files that are unknown to git
+    try:
+        git_fragments = check_output(
+            ["git", "ls-files"] + fragment_filenames, encoding="utf-8"
+        ).split("\n")
+        git_fragments = [os.path.abspath(f) for f in git_fragments if os.path.isfile(f)]
```
adiroiban: Minor comment

can we have this outside of try/except ?

Are we expecting this line to raise `CallerProcessError` ?

src/towncrier/_git.py
```diff
@@ -5,12 +5,26 @@
...
+    except CalledProcessError:
```
adiroiban, line 20: can we have a comment describing in which case we exepct to have this error raised ?

src/towncrier/test/test_build.py
```diff
@@ -747,8 +748,8 @@ def do_build_once_with(version, fragment_file, fragment):
                     "--yes",
                 ],
             )
-            # not git repository, manually remove fragment file
-            Path(f"newsfragments/{fragment_file}").unlink()
+            # fragment files unknown to git are removed even without a git repo
+            # Path(f"newsfragments/{fragment_file}").unlink()
```
adiroiban: I think that this needs a cleanup.

We should not merge commented code.

I think that we should repace this line with an assertion to check that the file is already removed.

src/towncrier/test/test_build.py
```diff
@@ -845,8 +846,8 @@ def do_build_once_with(version, fragment_file, fragment):
                 ],
                 catch_exceptions=False,
             )
-            # not git repository, manually remove fragment file
-            Path(f"newsfragments/{fragment_file}").unlink()
+            # fragment files unknown to git are removed even without a git repo
+            # Path(f"newsfragments/{fragment_file}").unlink()
```
adiroiban: this needs cleanup

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1531,55 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        At build time, it will delete any fragment file regardless of its stage,
+        included files that are not part of the git reporsitory,
+        or are just staged or modified.
+        """
+        # 123 is committed, 124 is modified, 125 is just added, 126 is unknown
+
+        with open("foo/newsfragments/123.feature", "w") as f:
+            f.write("Adds levitation")
+        with open("foo/newsfragments/124.feature", "w") as f:
+            f.write("Extends levitation")
```
adiroiban: Maybe add more info to know the purpose of each file.

```suggestion
            f.write("Adds levitation. File committed.")
        with open("foo/newsfragments/124.feature", "w") as f:
            f.write("Extends levitation. File modified in git.")
```

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1531,55 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+
+        commit()
+
+        with open("foo/newsfragments/125.feature", "w") as f:
+            f.write("Baz levitation")
```
adiroiban:
```suggestion
        with open("foo/newsfragments/125.feature", "w") as f:
            f.write("Baz levitation. Stagged file.")
```

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1531,55 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        with open("foo/newsfragments/126.feature", "w") as f:
+            f.write("Fix (literal) crash")
```
adiroiban:
```suggestion
        with open("foo/newsfragments/126.feature", "w") as f:
            f.write("Fix (literal) crash. File outside of git.")
```

adiroiban (changes requested):
Thanks for the changes.

I think that we are very close to being able to merge this.

But this needs more cleanup.

The commented lines in the test file are a blocker.

Thanks again.

abompard:
I've made the changes, thanks a lot for your patience. FYI I do appreciate the in-depth reviews, even for minor style changes. I'm frequently on the other end of those ;-)
So thanks.

src/towncrier/newsfragments/357.feature.rst
```diff
@@ -0,0 +1 @@
+It can now handle news fragments which are not part of the git repository. For example uncommited or unstagged files.
```
SmileyChris:
```suggestion
``towncrier build`` now handles removing news fragments which are not part of the git repository. For example, uncommitted or unstaged files.
```

SmileyChris (changes requested):
Currently, the command line docs say:
> By default, the processed news fragments are removed using git, which will also remove the fragments directory if now empty.

This should be updated. Maybe something like
> By default, the processed news fragments are removed. For any fragments committed in your git repository, ``git rm`` will be used (which will also remove the fragments directory if now empty).

adiroiban:
Thanks for the update. I will let Chris do the final review and approval for this PR.

@abompard Note that we merge with squash.
There is not much use in rewriting the history for the PR branch.
Rewriting/amending the commits on the PR make it harder to do a followup for a review.

Cheers

abompard:
> @abompard Note that we merge with squash. There is not much use in rewriting the history for the PR branch. Rewriting/amending the commits on the PR make it harder to do a followup for a review.

Ah, sorry about that, I didn't know.

FYI, I've had this same issue many times (trying to review PRs that rebase/rewrite) and I ended up writing a small tool to help me with the workflow: https://pypi.org/project/git-pr-branch/. The main idea is that it will download the different revisions of a PR in different branches that you can diff, even if it's always 1 commit that's being rebased.
Maybe you'll find it useful for your workflow as well.

adiroiban:
No problem.
I am looking after `towncrier` as my 4th job ... so I don't have much time for it.
This is why, using the embedded GitHub web tools makes it easier to review.

For most of the reviews, I don't checkout the code on my computer.
I am just doing a code review and hope the automated tests are doing their job.

adiroiban:
needs-review

adiroiban (approved):
This looks good to me.
I will wait a bit more for Chris to review this.

If there is a big delay, we can have this merged.

There is no rush, as there is no relesed planed anytime soon...
but we might want do to a new release soon :)

adiroiban:
I think that the comments from Chris were address.
I have merged this.

Thanks for your help!

PR #359, opened 2021-08-19
Fix deprecation warning from Tomli

hukkin:
- Pass a binary file object to Tomli to fix a deprecation warning
- ~~Turn DeprecationWarnings in the test suite to errors~~


src/towncrier/_settings.py
```diff
@@ -74,9 +75,12 @@ def load_config(directory):
...
 
 def load_config_from_file(directory, config_file):
-
-    with io.open(config_file, "r", encoding="utf8") as conffile:
-        config = _toml_load(conffile)
+    if tomli:
+        with open(config_file, "rb") as conffile:
+            config = tomli.load(conffile)
+    else:
+        with io.open(config_file, "r", encoding="utf8") as conffile:
```
graingert, line 82:
```suggestion
        with io.open(config_file, "r", encoding="utf8", newlines="") as conffile:
```

src/towncrier/_settings.py
```diff
@@ -74,9 +75,12 @@ def load_config(directory):
 
 
 def load_config_from_file(directory, config_file):
-
-    with io.open(config_file, "r", encoding="utf8") as conffile:
-        config = _toml_load(conffile)
+    if tomli:
+        with open(config_file, "rb") as conffile:
```
graingert, line 79:
```suggestion
        with io.open(config_file, "rb") as conffile:
```

src/towncrier/_settings.py
hukkin: Thanks!

src/towncrier/_settings.py
hukkin: This should be unnecessary in Python 3.6+ (which is always true in this branch)?

graingert:
hmm, does this need something to enable deprecation warnings in tests?

src/towncrier/_settings.py
graingert: yeah it's unnecessary, but I'm going to pyupgrade all these away in one go

src/towncrier/_settings.py
```diff
@@ -74,9 +75,12 @@ def load_config(directory):
...
+            config = tomli.load(conffile)
+    else:
+        with io.open(config_file, "r", encoding="utf8", newlines="") as conffile:
```
hukkin, line 82:
```suggestion
        with io.open(config_file, "r", encoding="utf8", newline="") as conffile:
```

graingert:
oops

hukkin:
> hmm, does this need something to enable deprecation warnings as failures in tests?

This doesnt seem to use a use a widely used test runner so I'm a bit clueless how to implement this :smile:

src/towncrier/_settings.py
hukkin: Haha, ok ill commit this if you want more work for pyupgrade? :smile:

graingert:
> This doesnt seem to use a use a widely used test runner 
😬

hukkin:
Ah it's twisted's own test runner :smile: Sorry for my ignorance, didn't mean to undermine it. What I meant to say is I personally have ever done this in pytest :smile: 

I think prefixing the test runner invocation in `tox.ini` with `python -W error::DeprecationWarning` should probably do the trick. Should I add it?

hukkin:
I tested that the `python -W error::DeprecationWarning -m` addition actually makes the test suite fail in master branch and it does. So commited that change here too.

graingert:
https://github.com/twisted/towncrier/runs/3376146214#step:7:60 ah that's an upstream twisted bug.
ok best leave it off for now and we'll have to do it later

hukkin:
:+1: reverted that change

graingert:
this doesn't need a newsfragment because tomli support hasn't been released yet

graingert (approved):

adiroiban:
I guess that this can be merged :)

PR #362, opened 2021-10-28
feat(settings): Add support to tables in toml settings

### LINKED ISSUE #361: "TypeError: string indices must be integers" when trying to use custom newsfragments types.
# Situation
I was trying to use custom newsfragments types since there is a mention about it in the [docs](https://towncrier.readthedocs.io/en/actual-freaking-docs/quickstart.html#creating-news-fragments) and I was curious about it.

I included pyproject.toml: 
```
[tool.towncrier.type.feat]
directory="feat"
name="Feature"
showcontent=true

[tool.towncrier.type.fix]
directory="fix"
name="BugFix"
showcontent=true

[tool.towncrier.type.misc]
directory="misc"
name="Misc"
showcontent=false

[tool.towncrier.type.docs]
directory="docs"
name="Docs"
showcontent=true

[tool.towncrier.type.foo]
directory="Foo"
name="Misc"
showcontent=false

```

And it (`towncrier check`) resulted in the exception "TypeError: string indices must be integers".

https://github.com/twisted/towncrier/blob/0e6a4793bb786053c491fde52ea2667c73807f75/src/towncrier/_settings.py#L104-L105

As is, the loop is iterating over the keys of the dictionary, `x` is a string.

# Proposed solutions

It think that the loop was expected to iterate over the values instead, something like:
```python
 for x in config["type"].values(): 
     types[x["directory"]] = {"name": x["name"], "showcontent": x["showcontent"]} 
```

 It would be interesting too to make  "directory"  optional in the config. Something like:

```python
newtypes_config = config["type]
for newtype_name, newtype_params in newtypes_config.items():
     type_directory = newtype_params.get("directory", newtype_name)
     types[directory] = {"name": newtype_params["name"],  "showcontent": newtype_params["showcontent"]} 
```

Thank you very much!

danieljmv01:
~~Correct loop so that it iterates over the values and not the keys of the custom fragment types if provided.~~ 

~~Resolves #361~~

* Iterate through values if types is a mapping instead of an array.

Resolves: twisted/#361

### DIFF
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -199,3 +199,83 @@ package = "foobar"
         self.assertEqual(
             str(e.exception), "Towncrier does not have a template named 'foo'."
         )
+
+    def test_custom_types_as_tables_array(self):
+        """
+        Custom fragment categories can be defined inside
+        the toml config file using an array of tables
+        (a table name in double brackets).
+        """
+        toml_content = (
+            "[tool.towncrier]\n"
+            "package = \"foobar\"\n"
+            "[[tool.towncrier.type]]\n"
+            "directory=\"foo\"\n"
+            "name=\"Foo\"\n"
+            "showcontent=true\n"
+            "[[tool.towncrier.type]]\n"
+            "directory=\"spam\"\n"
+            "name=\"Spam\"\n"
+            "showcontent=true\n"
+        )
+        expected = {
+            "foo": {
+                "name": "Foo",
+                "showcontent": True,
+            },
+            "spam": {
+                "name": "Spam",
+                "showcontent": True,
+            },
+        }
+
+        config = self._get_actual_config(
+            toml_content,
+        )
+        actual = config["types"]
+        self.assertDictEqual(expected, actual)
+
+    def test_custom_types_as_tables(self):
+        """
+        Custom fragment categories can be defined inside
+        the toml config file using tables.
+        """
+        test_project_path = self.mktemp()
+        os.makedirs(test_project_path)
+        toml_content = (
+            "[tool.towncrier]\n"
+            "package = \"foobar\"\n"
+            "[tool.towncrier.type.foo]\n"
+            "directory=\"bar\"\n"
+            "name=\"Bazz\"\n"
+            "showcontent=true\n"
+            "[tool.towncrier.type.spam]\n"
+            "directory=\"spam\"\n"
+            "name=\"Spam\"\n"
+            "showcontent=true\n"
+        )
+        expected = {
+            "bar": {
+                "name": "Bazz",
+                "showcontent": True,
+            },
+            "spam": {
+                "name": "Spam",
+                "showcontent": True,
+            },
+        }
+
+        config = self._get_actual_config(
+            toml_content,
+        )
+        actual = config["types"]
+        self.assertDictEqual(expected, actual)
+
+    def _get_actual_config(self, toml_content: str):
+        test_project_path = self.mktemp()
+        os.makedirs(test_project_path)
+        toml_path = os.path.join(test_project_path, "pyproject.toml")
+        with open(toml_path, "w") as f:
+            f.write(toml_content)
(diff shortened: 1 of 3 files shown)


adiroiban (commented):
Looks good. Thanks.

Can you add a test for custom types?

In this way we will make sure that this feature will always work and it will not be broken by future work.

But if you don't have time for a test, that is ok.

We can merge it as it is.

Thanks again!

danieljmv01:
Yes, I think I can do it tomorrow. Thanks!

src/towncrier/test/test_settings.py
```diff
@@ -30,6 +30,61 @@ def test_base(self):
         self.assertEqual(config["filename"], "NEWS.rst")
         self.assertEqual(config["underlines"], ["=", "-", "~"])
 
+    def test_custom_types(self):
```
adiroiban, line 33: can we these as 2 tests with separate docstring.

I am not sure that is the targeted difference between readme_format and alternative_format.

For the docstring, my suggetion is something like this

```
Custom fragment categories can be defined inside the toml config file using implicit file extension.
The file extension is the same as directory name.
```


```
Custom fragment categories can be defined inside the toml config file using custom file extension.
The file extension is define by the section name.
```

Does these docstrings make sense?
Is this what it is tested?

You can refactor the shared code in a helper function...but I think that it helps to have 2 separate test functions with 2 docstrings as they are covering 2 separate uses cases.

Thanks!

adiroiban:
and since this is a new feature, can you also update the configuration ? Thanks!

src/towncrier/test/test_settings.py
danieljmv01: Done.

danieljmv01:
> and since this is a new feature, 

Initially I didn't know that I was passing the configuration in an unexpected way, and I considered this a bug.  I tried to just fix it, without adding new features. 
But, since this is not a bug, but a new feature itself, I would like to stop using (or make it optional) the "directory" key if passing tables (it will be mandatory if using a table array). 
Also,  I would like to make optional  including "showcontent" key (e.g., assume that type has "showcontent": True if not passed).  
Do you consider that doing this is ok or do you prefer those changes in an additional PR?

> can you also update the configuration ? Thanks!

I am not sure about what do you refer. The configuration docs?

Thanks!

adiroiban:
Is this PR still valid or can be closed as invalid?

My understanding is that you found a way to use the existing towncrier functionality to implement your requirements.

I guess that we ended up with this PR since the existing documentation was not clear enough.

So maybe the documentation can be improved.

adiroiban:
@danieljmv01 . I have created #367 to solve the docs issue.

I prefer to have the code and docs in the same repo.

I agree that things are very confusing with the current setup.

So, if you have time, I am more than happy to review and approve a PR which moves the doc files into the main branch.

adiroiban:
@danieljmv01 can we close this PR? is there any value in having it merged?

Thanks

danieljmv01:
@adiroiban as is right now: no I don't see any value added. 

I wanted to keep changes as minimal as possible, and ended not adding value. What I would like is to give users a simpler, clearer and more intuitive and clear way to configure custom fragment types. I want to support this:

code-block-1:
```
    [tool.towncrier.type.feat]
    name = "Features"
    showcontent = true
```

Since the title of each table would describe the custom fragment type (see code-block-1)  the field `directory` is not necessary in this  way of configuration.

In the alternative (existing) way to configure custom types you have to include the `directory` field, that then will be use as key when transforming the array  of dicts to a dict. 

 I would rather wait until #368 is merged to rebase and update this, so I can document this  in this PR.   After rebasing this branch, I will update the code to support omitting the `directory` field.  If you still consider that it adds no value, then I'm fine with closing this PR.

adiroiban:
Ok.No hurry from my side. 
let me know when this PR is ready from a new review.

danieljmv01:
@adiroiban  I pushed to another branch  since this is not a bug. 

I have created #370, I will close this one.

PR #368, opened 2022-02-11
docs(Configuration): Move docs to the main branch and add info about custom fragment types.

danieljmv01:
* Describe how to add custom fragment types.

Resolves: twisted/#367


adiroiban:
Thanks for the update.

There is on issue...towncrier uses towncrier and we have a CI check :)

> CI / Check Newsfragment - Failed
> No new newsfragments found on this branch.

But I don't see that this is documented.

It looks like you need to create a file inside "src/towncrier/newsfragments/367.doc" with a sentence describing that the documentation was moved into the main branch.

------------

 I think the current CONTRIBUTING.rst should be moved into the docs and renamed "docs/release-process.rst"

And we can have a separate docs/CONTRIBUTING.rst file in which we add a few info for contributors... like requiting to have all tests green, to support all the python verison and to add a release note fragment.

But this can be done in a separate PR.

For this PR is enough to add the src/towncrier/newsfragments/367.doc

Thanks

adiroiban:
and I now see https://github.com/twisted/towncrier/pull/89 

So it looks like the docs are in the current branch as the PR was not merged.

@danieljmv01 you can review the feedback from that PR.

But at this time, I think is best to just merge what we have and then we can fix thinks in other PRs.

adiroiban:
let me know when you think this PR is ready for review and merge. Thanks

danieljmv01:
Hi @adiroiban, I have included the newsfragment and rebased, so I think that it is ready for review. Thanks!

PR #370, opened 2022-02-15
feat(settings): Add support to configure fragment types via tables

danieljmv01:
* Add support to toml tables for custom fragments,
  with default values.

Resolves: twisted/#369
Signed-off-by: danieljmv01 <danieljmv01@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -156,15 +156,21 @@ If ``title_format`` is unspecified or an empty string, the default format will b
 If set to ``false``, no title will be created.
 This can be useful if the specified template creates the title itself.
 
-Furthermore, you can add your own fragment types using:
+Furthermore, you can customize each of your own fragment types using:
 
 .. code-block:: toml
 
     [tool.towncrier]
-    [[tool.towncrier.type]]
-    directory = "deprecation"
-    name = "Deprecations"
-    showcontent = true
+	# To add custom fragment types, with default setting, just add an empty section.
+    [tool.towncrier.feat]
+    [tool.towncrier.fix]
+
+	# Custom fragment types can have custom attributes
+	# that are used when rendering the result based on the template.
+    [tool.towncrier.chore]
+        name = "Other Tasks"
+        showcontent = false
+
 
 
 Automatic pull request checks
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -25,18 +25,59 @@ Top level keys
 - ``issue_format`` -- A format string for rendering the issue/ticket number in newsfiles. ``#{issue}`` by default.
 - ``underlines`` -- The characters used for underlining headers. ``["=", "-", "~"]`` by default.
 
-
 Custom fragment types
 ---------------------
+``Towncrier`` allows defining custom fragment types. Custom fragment types
+will be used instead ``towncrier`` default ones, they are not combined.
 
-Custom fragment types can be including in the
-pyproject.toml.
-Each custom type (``[[tool.towncrier.type]]``) has the following keys:
+Users can configure each of their own custom fragment types by adding tables to
+the pyproject.toml named ``[tool.towncrier.type.<a custom fragment type>]``.
+
+These tables may include the following optional keys:
+
+ * ``name``: The description of the fragment type, as it must be included
+   in the news file. If omitted, it defaults to  its  fragment type,
+   but capitalized.
+ * ``showcontent``: Whether if the fragment contents should be included in the
+   news file. If omitted, it defaults to ``true``
+
+
+
+For example, if you want your custom fragment types to be
+``["feat", "fix", "chore",]`` and you want all
+of them to use the default configuration except ``"chore"`` you can do it as
+follows:
+
+
+.. code-block:: toml
+
+
+    [tool.towncrier]
+
+
+    [tool.towncrier.fragment.feat]
+    [tool.towncrier.fragment.fix]
+
+    [tool.towncrier.fragment.chore]
+        name = "Other Tasks"
+        showcontent = false
+
+DEPRECATED: Defining custom fragment types with an array of toml tables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Users can create their own custom fragment types by adding an array of
+tables to the pyproject.toml named ``[[tool.towncrier.type]]``.
+
+If still using this way to configure custom fragment types,
+please notice that ``fragment_types`` must be empty or not provided.
+
+Each custom type (``[[tool.towncrier.type]]``) has the following
+mandatory keys:
  * ``directory``: The type of the fragment.
  * ``name``: The description of the fragment type, as it must be included
    in the news file.
  * ``showcontent``: Whether if the fragment contents should be included in the
    news file.
+
 
 
 For example:
@@ -48,3 +89,8 @@ For example:
     directory = "deprecation"
     name = "Deprecations"
     showcontent = true
+
+    [[tool.towncrier.type]]
+    directory = "chore"
+    name = "Other Tasks"
+    showcontent = false
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -1,10 +1,12 @@
 # Copyright (c) Amber Brown, 2015
 # See LICENSE for details.
 
-from twisted.trial.unittest import TestCase
-
+import collections as clt
 import os
 from textwrap import dedent
+import textwrap
+
+from twisted.trial.unittest import TestCase
 
 from .._settings import load_config, ConfigError
 
@@ -172,7 +174,7 @@ package = "foobar"
             str(e.exception),
             "The template file '{}' does not exist.".format(
                 os.path.normpath(os.path.join(temp, "foo.rst")),
-            )
+            ),
         )
 
     def test_missing_template_in_towncrier(self):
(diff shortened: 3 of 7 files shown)


docs/configuration.rst
```diff
@@ -28,10 +28,49 @@ Top level keys
...
+   news file. If omitted, it defaults to ``true``
+
+
+For example:
+
+.. code-block:: toml
+
+
+        [tool.towncrier]
+        [tool.towncrier.type.chores]
```
adiroiban, line 58: can we have this like `[tool.towncrier.chores]`  with the mention that `type` is a reserved word.

I think this would look more "natural"

I don't like at all the old syntax like the one before...maybe it should be deprecated and suggestion new users to use the new syntax ? 

```
[[tool.towncrier.type]]
directory = "chore"
```

Thanks!

danieljmv01: Sure, no problem. 

I would like to confirm something. If using named tables like `[tool.towncrier.chores]` or `tool.towncrier.feat]` 
they will be mixed with other fields like `package`, `package_dir`, and so on. 

If mappings  are reserved to custom types, then I can get anything that behaves like a mapping and parse the fragment type from it.

An alternative that I like is adding a new field to the [tool.towncrier] table (for example, `custom_fragment_types`) that contains a list with all the custom types. This could be interesting since with default values, you would not need to add a named table per each fragment type, but for the ones that you want to configure. For example:
```
[tool.towncrier]
project="bar"

custom_fragment_types = [
  "feat",
  "refactor",  
  "fix", 
  "doc", 
  "chore"
]

[tool.towncrier.chores]
       name = "Other Tasks"
        showcontent = False
```

What do you think @adiroiban ?

docs/configuration.rst
```diff
@@ -28,10 +28,49 @@ Top level keys
...
+        name = "Other Tasks"
+        showcontent = False
+
+        [tool.towncrier.type.deprecations]
+        # This fragment type will be named "Deprecations"
+        # and it will show it contents.
+
+
+Defining custom fragment types with an array of toml tables
```
danieljmv01, line 67: @adiroiban  do you want to indicate this section of the docs as deprecated?

adiroiban: Yes... or not document it at all :) or only add a very brief info with the mention that this is no longer recommend.

docs/configuration.rst
adiroiban: I don't know if this should be named `custom_fragment_types` ... maybe just `available_types` or `fragment_types`  ?

I have not used TOML to much so I don't know what are the best practices for TOML.

We can also keep `[tool.towncrier.type.chores]` as maybe for TOML users this is normal.

----------

I don't care that much about this.

I set up towncrier for my project in 2017 and never had to change it since then :)

danieljmv01: I have remove `type` from the table name for the new way to configure fragment types (the existing one still uses `[[tool.towncrier.type]]`). So we can have `[tool.towncrier.chores]` or `[tool.towncrier.feat]`. 

But I can change  it to something like `[tool.towncrier.fragment.chores]` if you prefer. That would prevent conflicts between custom fragment types and other fields of `[tool.towncrier]`.

README.rst
```diff
@@ -149,22 +149,32 @@ Towncrier has the following global options, which can be specified in the toml f
     underlines = "=-~"
     wrap = false  # Wrap text to 79 characters
     all_bullets = true  # make all fragments bullet points
+    fragment_types = [] # A list of strings with custom fragment types. Leave it empty if you want to use the default ones.
```
adiroiban, line 152: I  think that due to lack the of the docs documentation, we ended up with documentation split between README and docs/

It would be nice to have fewer docs in the README.
README should just be a quick introduction to all the pointers for the project - docs, download, contributing... etc


```suggestion
    fragment_types = [] # A list strings to overwrite the default ones. If you want to reuse the default types you need to include them here. Leave it empty if you want to use the default ones.
```

adiroiban (commented):
Many thanks for the changes.

> So we can have [tool.towncrier.fragment.chores] or [tool.towncrier.fragment.feat].

I am a bit (just a bit) worried about possible future conflicts.

So, as you think is best.

We can play safe and go with the more verbose ` [tool.towncrier.fragment.chores]` as long as this is documented, it should be fine.

adiroiban:
I am ok with any change that makes you happy :) as long as it has tests to prevent any future regressions.

My only concern is that we should try to get it right, so that in the future we will not have to also deprecate this method :)

So, maybe play it safe and go with `[tool.towncrier.fragment.chores]`
It's bit more verbose, but you only define this once.
And it should be more explicit.

Thanks again

docs/configuration.rst
```diff
@@ -24,21 +24,63 @@ Top level keys
...
+   news file. If omitted, it defaults to ``true``
+
+
+For example:
+
+.. code-block:: toml
+
+
+    [tool.towncrier]
+    fragment_types = [
```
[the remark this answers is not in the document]
danieljmv01: Do you refer to moving all the fields in `[tool.towncrier]` to `[tool.towncrier.settings]` like this?:

```toml
[tool.towncrier]
[tool.towncrier.settings]
package="foo"
newsfile="news.rst"
fragment_types = []
```
The Json equivalent:
```json
{
  "tool": {
    "towncrier": {
      "settings": {
        "package": "foo",
        "newsfile": "news.rst",
        "fragment_types": []
      }
    }
  }
}

```

I don't think it solves anything. It just adds a nesting level.
---------------------------------------------------------------------------------------------------------------------

The conflicts that I mention was about  using `[tool.towncrier.chores]` :
```toml
[tool.towncrier]
package="foo"
newsfile="news.rst"
fragment_types = [
     "chores",
     "newsfile", # Same as the  field of [tool.towncier]
]
[tool.towncrier.chores]
name= "foo"

```
```json
{
  "tool": {
    "towncrier": {
      "package": "foo",
      "newsfile": "news.rst",
      "fragment_types": [
        "chore",
        "package"
      ],
      "chore": {
        "name": "foo",
      }
    }
  }
}
```
`"package"` was not configured, but another field is named like that so `self.config.get("package", {})` would return a str where a dict was expected. 
 If we  use `[tool.towncrier.fragment.chores]` (I can use other thing instead of `fragment` as  long as it is not used, `fragment_options` for example) this problem can't happen:
```toml
[tool.towncrier]
package="foo"
newsfile="news.rst"
fragment_types = [
     "chores",
     "newsfile", # Same as the  field of [tool.towncier]
]
[tool.towncrier.fragment.chores]
name= "foo"

```
```json
{
  "tool": {
    "towncrier": {
      "package": "foo",
      "newsfile": "news.rst",
      "fragment_types": [
        "chore",
        "package"
      ],
      "fragment": {
        "chore": {
          "name": "foo",
        }
      }
    }
  }
}
```
Since towncrier uses `"fragment"` to store all the options now, `self.config.get("fragment", {}).get("package", {})` (sorry for the trainwreck) now will return an empty dict.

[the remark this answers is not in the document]
adiroiban: > I don't think it solves anything. It just adds a nesting level.

I was thinking that this allows having `towncrier.chore` fragment or  `towncrier.package`

The only "reserved" fragmet would be "settings".

-------

But we can play it safe and go with 

```
[tool.towncrier.fragment.chores]
name= "foo"
```

adiroiban:
So, maybe we should go with something like this

```
[tool.towncrier]
package="foo"
newsfile="news.rst"

[tool.towncrier.fragment.chores]
name= "foo"
```

It's more verbose, but we don't risk any colision.
I am not very happy about the extra `fragment_types` list.

I think that if we place everything inside `tool.towncrier.fragment.*` they can be automatically discovered.

danieljmv01:
> So, maybe we should go with something like this
> 
> ```
> [tool.towncrier]
> package="foo"
> newsfile="news.rst"
> 
> [tool.towncrier.fragment.chores]
> name= "foo"
> ```
> 
> It's more verbose, but we don't risk any colision. I am not very happy about the extra `fragment_types` list.
> 
> I think that if we place everything inside `tool.towncrier.fragment.*` they can be automatically discovered.


Ok, cool. I will use the keys of "tool.towncrier.fragment" mapping to get the fragments instead of using   "fragment_types" array then.

README.rst
```diff
@@ -149,22 +149,26 @@ Towncrier has the following global options, which can be specified in the toml f
     underlines = "=-~"
     wrap = false  # Wrap text to 79 characters
     all_bullets = true  # make all fragments bullet points
+    fragment_types = [] # A list of strings of strings with custom fragment types. Leave it empty if you want to use the default ones.
```
adiroiban, line 152: Is this still needed?

README.rst
```diff
@@ -149,22 +149,26 @@ Towncrier has the following global options, which can be specified in the toml f
...
 
 .. code-block:: toml
 
     [tool.towncrier]
-    [[tool.towncrier.type]]
-    directory = "deprecation"
-    name = "Deprecations"
-    showcontent = true
+    [tool.towncrier.feat]
+    [tool.towncrier.fix]
```
adiroiban, line 166:
```suggestion
    # To add custom fragment types, with default setting, just add an empty section.
    [tool.towncrier.feat]
    [tool.towncrier.fix]
```

README.rst
```diff
@@ -149,22 +149,26 @@ Towncrier has the following global options, which can be specified in the toml f
...
+
+    [tool.towncrier.chore]
+        name = "Other Tasks"
+        showcontent = false
```
adiroiban, line 170:
```suggestion
    # Custom fragment types can have custom attributes
    # that are used when rendering the result based on the template.
    [tool.towncrier.chore]
        name = "Other Tasks"
        showcontent = false
```

src/towncrier/test/test_settings.py
```diff
@@ -199,3 +201,70 @@ def test_missing_template_in_towncrier(self):
         self.assertEqual(
             str(e.exception), "Towncrier does not have a template named 'foo'."
         )
+
+    def test_custom_types_as_tables_array(self):
+        """
+        Custom fragment categories can be defined inside
+        the toml config file using an array of tables
+        (a table name in double brackets).
+        """
```
adiroiban, line 210:
```suggestion
    def test_custom_types_as_tables_array_deprecated(self):
        """
        Custom fragment categories can be defined inside
        the toml config file using an array of tables
        (a table name in double brackets).
        
        This functionality is considered deprecated, but we continue
        to support it to keep backward compatibility.
        """
```

src/towncrier/test/test_settings.py
```diff
@@ -199,3 +201,70 @@ def test_missing_template_in_towncrier(self):
...
+            (u"fix", {"name": u"Fix", "showcontent": True, }),
+        ]
+
+        expected = clt.OrderedDict(expected)
+        config = self._get_actual_config(toml_content,)
+        actual = config["types"]
+        self.assertDictEqual(expected, actual)
+
+    def _get_actual_config(self, toml_content):
+        test_project_path = self.mktemp()
```
adiroiban, line 264: I think there is no need to make it private, as this is testing code.

I think that if a helper method can be used by multiple tests, it can have a public name.

We can still use private names in testing code, for code that should only be used by a single test.

```suggestion
    def load_config(self, toml_content):
        """
        Load the configuration based on a temporary
        file having `toml_content` as content.
        """
        test_project_path = self.mktemp()
```

or make `def config_from_text` or `def load_content` 

any name is ok, but it's important to have a docsting describing the purpose / intent of this helper method.

adiroiban (changes requested):
Looks good. Many thanks.

I left a few minor comments.

Other than that, I think that this is ready to merge.

adiroiban (approved):
All good. Thanks.

let's merge this.

We need to remove the duplication of docs from README and docs/ , but that can be don in a separate PR.

Thanks!

PR #374, opened 2022-03-27
Create message in command line

saroad2:
Description
=======

When using `git` one can set a commit message using the `-m` flag. For example:
```
git commit -m "This is my message"
```

Taking inspiration from this, now *Towncrier* can also get a message from user using a `-m` flag:

```
towncrier create 123.feature -m "This is a message that describes the news fragment"
```
This feature plays nicely with the `--edit` flag.

### DIFF
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -94,7 +94,7 @@ def _get_news_content_from_user(message):
     if message is not None:
         initial_content += (
             "\n"
-            f"{message}\n"
+            "{message}\n".format(message=message)
         )
     content = click.edit(initial_content)
     if content is None:
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -22,12 +22,17 @@ from ._settings import load_config_from_options
     default=False,
     help="Open an editor for writing the newsfragment content.",
 )  # TODO: default should be true
+@click.option(
+    "-m", "--message",
+    type=str,
+    help="Set message in new fragment."
+)
 @click.argument("filename")
-def _main(ctx, directory, config, filename, edit):
-    return __main(ctx, directory, config, filename, edit)
+def _main(ctx, directory, config, filename, edit, message):
+    return __main(ctx, directory, config, filename, edit, message)
 
 
-def __main(ctx, directory, config, filename, edit):
+def __main(ctx, directory, config, filename, edit, message):
     """
     The main entry point.
     """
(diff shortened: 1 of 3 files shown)


src/towncrier/test/test_create.py
```diff
@@ -93,6 +98,26 @@ def test_edit_abort(self):
                 self.assertEqual([], os.listdir("foo/newsfragments"))
                 self.assertEqual(1, result.exit_code)
 
+    def test_message(self):
+        """Set fragment message in command line"""
```
adiroiban, line 102: Maybe add more info about the targeted test case / functionality.

Without documentation, it's hard otherwise to know what is the intended behaviour.

```suggestion
    def test_message(self):
        """
        When creating a new fragment the content can be passed as a
        command line argument.
        The text editor is not invoked.
        """
```

src/towncrier/test/test_create.py
```diff
@@ -93,6 +98,26 @@ def test_edit_abort(self):
...
+        message = "This is a message"
+        self._test_success(content=[message], additional_args=["-m", message])
+
+    def test_message_and_edit(self):
+        """Set fragment message in command line and edit"""
```
adiroiban, line 107: maybe.
```suggestion
    def test_message_and_edit(self):
        """
        When creating a new message, a initial content can be passed via
        the command line and continue modifying the content by invoking the
        text editor.
        """
```

src/towncrier/newsfragments/374.feature.rst
```diff
@@ -0,0 +1,2 @@
+
+Add -m flag to create endpoint
```
adiroiban, line 2: This is the end user text that helps advertise the new functionality.

I think that is best to write it as a full sentence.

```suggestion
The `towncrier create` command line now has a new `-m TEXT` argument that is used to define the content of the newly created fragment.
```

src/towncrier/create.py
```diff
@@ -22,12 +22,17 @@
     default=False,
     help="Open an editor for writing the newsfragment content.",
 )  # TODO: default should be true
+@click.option(
+    "-m", "--message",
+    type=str,
+    help="Set message in new fragment."
+)
```
adiroiban, line 29: I am OK with `message`.

Just a minor suggest to consider, but feel free to discard it.

```suggestion
@click.option(
    "-c", "--content",
    type=str,
    help="Sets the content of the new fragment."
)
```

adiroiban (approved):
Many thanks for the change.
Looks good. Thanks.

Only a few minor comments to consider.

Especially improving the test docstring as we don't have separate documentation to specify what is the expect behaviour and what to consider a feature and what a bug :)

I think that is ok to not add extra narrative documentation, as long as the tests explains the intended behaviour.

saroad2:
Fixed after code review.

The reduction in coverage is because the portion of untested lines got 0.05% up.
However, every ability presented in this PR was tested.

adiroiban:
thanks for this perfect PR.
Looks like codecov.io went wild :(

Merged with the power of administrators :)

PR #376, opened 2022-03-28
add CPython 3.10, replace PyPy 3.6 with 3.8, use CPython 3.9 for checks

altendky:


altendky:
@saroad2, ready for review.

I set Python 3.9 for checks because their purpose is to run reliably, not test the latest Python version.  As such I think it is reasonable to keep them one minor version behind.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = cov-erase, flake8, {pypy27,pypy36,pypy37,py27,py35,py36,py37,py38,py39}-{tests,flake8,check-manifest,check-newsfragment}, cov-report
+envlist = cov-erase, flake8, {pypy27,pypy37,pypy38,py27,py35,py36,py37,py38,py39,py310}-{tests,flake8,check-manifest,check-newsfragment}, cov-report
```
saroad2, line 2: I must say that I don't quite know why he separation between `cov-erase` and `cov-report`.

In my projects I just join the testing and coverage to one *Tox* task. Here's an [example](https://github.com/saroad2/statue/blob/main/tox.ini)

saroad2 (approved):

saroad2:
> I set Python 3.9 for checks because their purpose is to run reliably, not test the latest Python version. As such I think it is reasonable to keep them one minor version behind.

I agree with this approach. The smoke tests should run on 3.9 and 3.10 should be checked later.

altendky:
I personally like to get all the feedback rather than failing fast.  Then you have all the relevant information.

It looks like the `cov-erase` bit was introduced in https://github.com/twisted/towncrier/commit/4068ee4a7e9aaca2f2fafb44665df5e4a1ec7aab#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449 which was before I started working with towncrier.  I'm not sure how this ends up working with parallel runs.  I guess you would have to have tox not run in parallel and just run one env at a time.  With your setup that would result in each individual version erasing, testing, and reporting coverage whereas this setup erases, tests all versions, reports the overall coverage from all versions together.  I think...  I haven't played with it though.

@adiroiban, go ahead and merge this if you want.  You've been more active so I don't want to jump in and merge this without an "official" approval, or you deciding to accept saroad2's review.

PR #377, opened 2022-04-07
Get rid of setup.cfg

hynek:
The only option present in setup.cfg has been available in setup.py for a fews years (if not always) already. :)


hynek:
Yeah that coverage decrease ain't my fault. :D

adiroiban (approved):
It this makes people happy we can merge it.

And I think that this works for towncier.

---------

Just asking ?

What is wrong with having a setup.py

I am using the setup.py for `python setup.py bdist_wheel` and then publish in a private repo
... if you are paranoid and don't trust PyPi :)

hynek:
> What is wrong with having a setup.py

This is removing setup.**cfg** that has only one line in it, not setup.py. :)

hynek:
for the sake of completion: you can – [and should](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) – build without invoking setup.py. The cleanest way is using https://pypa-build.readthedocs.io/

PR #380, opened 2022-04-09
Drop compat cruft

hynek:
So this removes all the obvious, machine-removable pre-3.7 compat cruft. There's some OrderedDicts that could be replace by normal ones, but that's a bit more delicate because insert order is preserved, but not on creation.

---

The version in docs/conf.py should probably not be hard-coded? But I'm not sure how docs are built since https://towncrier.readthedocs.io/ has only `latest`.

---

To ensure modern code, I've added a pre-commit tox env and a very simple pre-commit config. If you're OK with that, I would also add flake8, black, and isort to it and talk to Anthony to get us free per-commit.ci (much faster than GHA, fixes problems in-PR)..


hynek:
I'm confident the global coverage changes are again due to removal of lines.

I think it would be a good idea to switch off the global check (and keep the patch check) until towncrier has 100% code coverage. Otherwise there's gonna be a lot of red/failed PRs which is somewhat demoralizing.

adiroiban:
The code coverage reports are only "informational" they are not blocking a PR.

---------

The RTD build is brand new. This is why it only has latest.
There was no version release with support for RTD. This is why it only has latest.

I agree that the version should not be hardcoded. PR welcomed :)

-------

I know some people would like to see py2.7 and python 3.5 support in Towncrier.

But as mentioned, I am only the maintenance maintainer:)

So I will approve it.

The reasoning is that in 10 years we might want to do these changes anyway.
And since we have the resources to the them now, let's do it.

Many thanks for the PR.

adiroiban (approved):
For a glorious py3 future.

src/towncrier/_builder.py
```diff
@@ -276,11 +275,11 @@ def get_indent(text):
...
         # indentation depends on whether or not this is a bullet point.
         # (it is probably usually best to disable wrapping in that case)
-        if all_bullets or text[:2] == u"- " or text[:2] == u"* ":
-            return u"  "
+        if all_bullets or text[:2] == "- " or text[:2] == "* ":
+            return "  "
         elif text[:3] == "#. ":
-            return u"   "
-        return u""
+            return "   "
```
adiroiban, line 281: @hynek it looks like this is a missing code coverage.

I expect that this is very easy to cover, by updating the format used by an existing test.

hynek: Could it be that there was a subtle bug or something working by accident before? 🤔

hynek:
> I know some people would like to see py2.7 and python 3.5 support in Towncrier.

We should keep in mind that:

- towncrier is a tool, so the version it runs on is rather circumstantial. Especially given it’s a tool you nominally only need while releasing software.
- There’s plenty of towncrier releases for those versions. :)

src/towncrier/_builder.py
adiroiban: Nope. Just missing test.

Here is an example of an updated test that will provide full cover for this feature

```diff
diff --git a/src/towncrier/test/test_build.py b/src/towncrier/test/test_build.py
index 66a1123..671f2e2 100644
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -622,6 +622,9 @@ class TestCli(TestCase):
     def test_bullet_points_false(self):
         """
         When all_bullets is false, subsequent lines are not indented.
+
+        The automatic ticket number inserted by towcier will allign with the
+        manual bullet.
         """
         runner = CliRunner()
 
@@ -634,7 +637,24 @@ class TestCli(TestCase):
                 )
             os.mkdir("newsfragments")
             with open("newsfragments/123.feature", "w") as f:
-                f.write("wow!\n~~~~\n\nAdds levitation.")
+                f.write(
+                    "wow!\n"
+                    "~~~~\n"
+                    "\n"
+                    "No indentation at all."
+                    )
+            with open("newsfragments/124.bugfix", "w") as f:
+                f.write(
+                    "#. Numbered bullet list."
+                    )
+            with open("newsfragments/125.removal", "w") as f:
+                f.write(
+                    "- Hyphen based bullet list."
+                    )
+            with open("newsfragments/126.doc", "w") as f:
+                f.write(
+                    "* Asterisk based bullet list."
+                    )
 
             result = runner.invoke(
                 _main,
@@ -666,8 +686,29 @@ class TestCli(TestCase):
             wow!
             ~~~~
 
-            Adds levitation.
+            No indentation at all.
             (#123)
+
+
+            Bugfixes
+            --------
+
+            #. Numbered bullet list.
+               (#124)
+
+
+            Improved Documentation
+            ----------------------
+
+            * Asterisk based bullet list.
+              (#126)
+
+
+            Deprecations and Removals
+            -------------------------
+
+            - Hyphen based bullet list.
+              (#125)
             """
             ).lstrip(),
         )
```

adiroiban:
> The code coverage reports are only "informational" they are not blocking a PR.

I mean the overall "project" report is optional.

I think that the patch coverage should be required 100% as the minimum level to ensure a glorious future.

adiroiban:
@hynek I have enabled auto-merge for this PR.
Can you try to apply the updated  test from the path .
I hope that with that we can get 100% coverage and this will be automatically merged.

src/towncrier/_builder.py
hynek: But how come we seem to have lost it now? I'm just on my phone ATM so I can't check but the old mixture of u and non-u strings looks sketchy? 🤔

hynek:
> @hynek I have enabled auto-merge for this PR.
> 
> Can you try to apply the updated  test from the path .
> 
> I hope that with that we can get 100% coverage and this will be automatically merged.

Sure, but I'm not positive I can get to a computer until tomorrow.

If you want to speed things up, you could try to "suggest" the change in PR so I can just accept it? 🤔 It should be `alt-g` IIRC.

PR #381, opened 2022-04-11
Bootstrap pre-commit.ci

hynek:
In an absolutely perfect demonstration, the applied missing test in #380 broke pyupgrade. This PR will prevent this happening in the future. :)

---

This adds no new code linters or formatters, but ensures that flake8 (and some
minor others) are run using pre-commit.ci on each PR.

That's very fast and fixes problems in-PR, not requiring review turnarounds.

Additionally we get more stable linters via version pinning for free. The
updates are capped to once per month to prevent unnecessary churn.

Since flake8 now runs over the whole code base, I've added a few noqas where it
made sense.


hynek:
Hmmm it looks like I might not be authorized to add pre-commit to this repo. 🤔

hynek:
Turns out I was spamming @glyph, but it's live now!

See https://results.pre-commit.ci/run/github/48647797/1649661064.5Ui8YSZ6SNekzeJLUWgBbA and the wonderful `total time to completion: 6.1s`

adiroiban:
force the merge... codecov.io just being codecov.io  and not reporting a status :)

PR #382, opened 2022-04-11
Add black & isort to pre-commit

hynek:
Use the default 88 columns from Black while allowing for 99 columns for longer (doc)strings.

---

At least I've looked at Twisted and it seems like it doesn't change the default?

Anyhow with this merged, there should be two categories less of bikeshedding and unnecessary review cycles. :)


hynek:
I'd like to point out once again: `total time to completion: 5.5s` ;) it was done before the first GHA build even started.

hynek:
I would also like to suggest to:

- make passing pre-commit mandatory in branch settings
- switch to squash and merge. i'm already somehow the third most prolific contributor just because i like small commits in my own branches 😅

adiroiban:
Thanks for the cleanup.

> make passing pre-commit mandatory in branch settings

Done.

---------

Can we have these commits to the `.git-blame-ignore-revs` file ?

-------------


> switch to squash and merge. i'm already somehow the third most prolific contributor just because i like small commits in my own branches

I am using `squash and merge` for internal projects.

But for other twisted repos the rule is to always merge with a "normal" commit merge.

I think that for towncrier is OK not to squash.

It should make it easier to keep multiple parallel branches in sync with each other and the main branch.

hynek:
> Can we have these commits to the `.git-blame-ignore-revs` file ?

I was planning on that, but wanted to make sure they don’t get rewritten. If it’s all that’s missing, I’d go and do it.

PR #383, opened 2022-04-12
Shrink CI matrix

### LINKED ISSUE #379: Save the planet
I think that is safe to run pypy test only on Linux.

They take ridiculous amount of time on Win ... compared with less than 2 minutes on Linux.

image

And I think that for macOS and Windows, we can live with just running cPython 3.10 tests.

All this to reduce the resource used to run our CI matrix.

hynek:
Everything except Linux is very slow on GHA.

macOS is very similar to Linux, so we can safely drop it and it's fine to test
one CPython and one PyPy on Windows.

Saves the planet...eh... fixes #379

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,9 +121,6 @@ jobs:
           - name: CPython 3.9
             tox: py39
             action: '3.9'
-          - name: PyPy 3.8
-            tox: pypy38
-            action: pypy-3.8
         task:
           - name: Test
             tox: tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,10 @@ jobs:
             tox: build
 
     steps:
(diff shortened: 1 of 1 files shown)


hynek:
I think we've got it!

23 →  11 jobs 
8:08 → 5:48 mins (~30%)

hynek:
Now it took 7+mins, almost 6 mins being from pypy on Windows…maybe we should drop it after all? The potential for bugs over CPython on Windows and the number of expected users of runing towncrier on PyPy3 on Windows don't seem to make it worth it…

adiroiban (approved):
Thanks for the cleanup.

I think is safe to remove pypy from Windows and only keep it for Linux.

I don't expect people using pypy with Towncrier , on Windows :)

If there are people using it, and we found a bug, we can re-add it.

Thanks!

hynek:
3:5! ;)

PR #391, opened 2022-06-24
bugfix in _writer.py

mikesongming:
the version-already-produced error raising cannot be reached in expected single file mode like the test case in :test_write:WritingTests.test_append_at_top_with_hint

### DIFF
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -26,7 +26,7 @@ def append_to_newsfile(
     else:
         existing_content = [""]
 
-    if top_line and top_line in existing_content:
+    if top_line and top_line in existing_content[-1]:
         raise ValueError("It seems you've already produced newsfiles for this version?")
 
     with open(os.path.join(directory, filename), "wb") as f:
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -193,7 +193,7 @@ def __main(
         start_string = config["start_string"]
         news_file = config["filename"]
 
-        if config["single_file"]:
+        if not config["single_file"]:
             # When single_file is enabled, the news file name changes based on the version.
             news_file = news_file.format(
                 name=project_name, version=project_version, project_date=project_date
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -193,8 +193,9 @@ def __main(
         start_string = config["start_string"]
         news_file = config["filename"]
 
-        if not config["single_file"]:
-            # When single_file is enabled, the news file name changes based on the version.
+        if config["single_file"] is False:
+            # The release notes for each version are stored in a separate file.
+            # The name of that file is generated based on the current version and project.
             news_file = news_file.format(
                 name=project_name, version=project_version, project_date=project_date
             )
--- a/src/towncrier/newsfragments/391.bugfix
+++ b/src/towncrier/newsfragments/391.bugfix
@@ -1,0 +1,3 @@
+
+fix `single_file` option related newsfile naming and duplicate version
+writing behavior, to be consistent with documentation.
--- a/src/towncrier/newsfragments/391.bugfix
+++ b/src/towncrier/newsfragments/391.bugfix
@@ -1,3 +1,3 @@
+The detection of duplicate release notes was fixed and recording changes of same version is no longer triggered.
 
-fix `single_file` option related newsfile naming and duplicate version
-writing behavior, to be consistent with documentation.
+Support for having the release notes for each version in a separate file is working again. This is regression introduced in VERSION 19.9.0rc1.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -2,8 +2,8 @@
 # See LICENSE for details.
 
 import os
-from pathlib import Path
 
+from pathlib import Path
 from subprocess import call
 from textwrap import dedent
 
@@ -525,7 +525,7 @@ class TestCli(TestCase):
                     "--date",
                     "01-01-2001",
                     "--yes",
-                ]
+                ],
             )
             # not git repository, manually remove fragment file
             Path(f"newsfragments/{fragment_file}").unlink()
@@ -538,12 +538,18 @@ class TestCli(TestCase):
                     '[tool.towncrier]\n single_file=false\n filename="{version}-notes.rst"'
                 )
             os.mkdir("newsfragments")
-            results.append(do_build_once_with("7.8.9", "123.feature", "Adds levitation"))
+            results.append(
+                do_build_once_with("7.8.9", "123.feature", "Adds levitation")
+            )
             results.append(do_build_once_with("7.9.0", "456.bugfix", "Adds catapult"))
 
             self.assertEqual(0, results[0].exit_code, results[0].output)
             self.assertEqual(0, results[1].exit_code, results[1].output)
-            self.assertEqual(2, len(list(Path.cwd().glob("*-notes.rst"))), "one newfile for each build")
+            self.assertEqual(
+                2,
+                len(list(Path.cwd().glob("*-notes.rst"))),
+                "one newfile for each build",
+            )
             self.assertTrue(os.path.exists("7.8.9-notes.rst"), os.listdir("."))
             self.assertTrue(os.path.exists("7.9.0-notes.rst"), os.listdir("."))
 
@@ -621,7 +627,7 @@ class TestCli(TestCase):
                     "--date",
                     "01-01-2001",
                     "--yes",
-                ]
+                ],
             )
             # not git repository, manually remove fragment file
             Path(f"newsfragments/{fragment_file}").unlink()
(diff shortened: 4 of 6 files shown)


mikesongming:
> Thanks for the update.
> 
> I only see this about single file
> 
> > If a single file is used, the content of that file gets overwritten each time.
> 
> Can you add more info about why someone would want to use this and how to use it?
> 
> Maybe the automated tests should be updated the reflect the real usage.
> 
> Note I am just doing "emergency reviews" for this project.
> 
> If the tests are green we can merge this PR. No problem.
> 
> But if you are using the "single file" feature it would be nice to document it and make sure the automated tests will fail if any future change will introduce regressions for this feature.

After reading docs and source code of `v21.9.0`, I guess that there are two use cases for the `single_file` flag:

1. `single_file=true`   :  the default choice for most projects. People hopes to have only one fixed **news file** given by the confiuration option `filename`, and  in each `towncrier build` to packed newly versioned change log at the top of **news file**. In case of duplicate version submitting, some author added the "version-already-produced error raising" code in _write.py:29-30 as: 
```
    if top_line and top_line in existing_content:
```
However, that branch of code woule never be triggered due to the 'splitting by start_string' and writing logic afterwards.

2. `single_file=false` : in some unknown scenario, people may hope writing to unique news files in each `towncrier build` run, as [depicted in docs](https://towncrier.readthedocs.io/en/latest/index.html?highlight=single_file#further-options): 
> single_file = true  # if false, filename is formatted like `title_format`.
However, build.py:196:200 implemented in the reverse direction:

 ```
       if config["single_file"]:
            # When single_file is enabled, the news file name changes based on the version.
            news_file = news_file.format(
                name=project_name, version=project_version, project_date=project_date
            )
```

Hope my explanation helps and booth discrepancies have been fixed in this PR. 

My another issue #390 has extracted related code snippets for details.

**Comment by mikesongming** [comment]
@adiroiban please hold the PR, additional fixes needed

**Comment by mikesongming** [comment]
@adiroiban After adding a news fragment, ci workflow passed with all success. Please check and approve the PR.

https://github.com/twisted/towncrier/actions/runs/2642268377

File: src/towncrier/build.py [hunk 1]
```diff
@@ -193,7 +193,7 @@ def __main(
         start_string = config["start_string"]
         news_file = config["filename"]
 
-        if config["single_file"]:
+        if not config["single_file"]:
             # When single_file is enabled, the news file name changes based on the version.
```
adiroiban, line 197: Is this change ok ?

I was expecting something like this

```suggestion
        if config["single_file"] is True:
            # When single_file is enabled,
            # the release notes for each version are stored in a separate file.
            # The name of that file is generated based on the current version and project.
```

src/towncrier/newsfragments/391.bugfix
```diff
@@ -0,0 +1,3 @@
+
```
adiroiban, line 1: I think that we have 2 separate issues here.

* The detection of duplicate release notes was fixed and SOME PREVIOUS BEHAVIOUR is no longer triggered.
* Support for having the release notes for each version in a separate file is working again. This is regression introduced in VERSION X.Y.Z.

src/towncrier/test/test_build.py
```diff
@@ -500,54 +501,92 @@ def test_no_package_changelog(self):
             ).lstrip(),
         )
 
-    def test_single_file(self):
+    def test_single_file_false(self):
         """
-        Enabling the single file mode will write the changelog to a filename
-        that is formatted from the filename args.
+        Disabling the single file mode will write the changelog to a filename
```
adiroiban, line 506: I think that the docstring of the tests should be self contained and not link to the documentation.

It should explain the targeted feature, regardless of the documentation.

In some way, the docstring of  this test serves as the "real" documentation :)

src/towncrier/test/test_build.py
```diff
@@ -500,54 +501,92 @@ def test_no_package_changelog(self):
...
+        that is formatted from the filename args, w.r.t. [documentation]
+        (https://towncrier.readthedocs.io/en/latest/index.html?highlight=single_file#further-options)
+        as:
+            single_file = true  # if false, filename is formatted like `title_format`.
```
adiroiban, line 510:
```suggestion
        When `single_file = false` the release notes for each version are stored
        in a separate file.
        The name of the file is defined by the `title_format` configuration value.
```

src/towncrier/test/test_build.py
```diff
@@ -500,54 +501,92 @@ def test_no_package_changelog(self):
...
+            Path(f"newsfragments/{fragment_file}").unlink()
+            return result
 
-            self.assertEqual(0, result.exit_code, result.output)
+        results = []
+        with runner.isolated_filesystem():
+            with open("pyproject.toml", "w") as f:
+                f.write(
+                    '[tool.towncrier]\n single_file=false\n filename="{version}-notes.rst"'
+                )
```
adiroiban, line 539: A bit easier to read.
```suggestion
                f.write(
                    '[tool.towncrier]\n'
                    ' single_file=false\n'
                    ' filename="{version}-notes.rst"\n'
                    ''
                )
```

src/towncrier/test/test_build.py
```diff
@@ -500,54 +501,92 @@ def test_no_package_changelog(self):
             ).lstrip(),
         )
 
-    def test_single_file(self):
+    def test_single_file_false(self):
```
adiroiban, line 504: maybe a better description of the targeted feature

```suggestion
    def test_release_notes_in_separate_files(self):
```

src/towncrier/test/test_build.py
```diff
@@ -568,55 +607,79 @@ def test_singlefile_errors_and_explains_cleanly(self):
             result.output,
         )
 
-    def test_single_file_false(self):
+    def test_single_file(self):
```
adiroiban, line 610: Maybe a better description for the feature

```suggestion
    def test_all_version_notes_in_a_single_file(self):
```

src/towncrier/test/test_build.py
```diff
@@ -568,55 +607,79 @@ def test_singlefile_errors_and_explains_cleanly(self):
...
         """
-        If formatting arguments are given in the filename arg and single_file is
-        false, the filename will not be formatted.
+        "If a single file is used, the content of that file gets overwritten each time."
```
adiroiban, line 612:
```suggestion
        When `single_file = true` the single file is used to store the notes
        for multiple versions.
        
        The name of the file is fixed and is not extrapolated with any variable.
```

src/towncrier/test/test_build.py
```diff
@@ -568,55 +607,79 @@ def test_singlefile_errors_and_explains_cleanly(self):
...
+            Path(f"newsfragments/{fragment_file}").unlink()
+            return result
 
-            self.assertEqual(0, result.exit_code, result.output)
+        results = []
+        with runner.isolated_filesystem():
+            with open("pyproject.toml", "w") as f:
+                f.write(
+                    '[tool.towncrier]\n single_file=true\n filename="{version}-notes.rst"'
+                )
```
adiroiban, line 641: I think that this is the important part.
```suggestion
                f.write(
                    '[tool.towncrier]\n'
                    ' single_file=true\n'
                    ' # The `filename` variable is fixed and not formated\n`
                    ' # in any way.\n'
                    ' filename="{version}-notes.rst"\n
                    ''
                )
```

adiroiban (changes requested):
Changes for the single_file looks good.

The variable name is confusing :( ... so dedicated documentation should help a lot.

------------

This needs-changes and I am not sure that the `if top_line and top_line in existing_content[-1]:` change is ok.

Do we have tests for this?
Was any of the existing test failing?
 
Maybe have this in a separate PR.

Thanks again

src/towncrier/_writer.py
```diff
@@ -26,7 +26,7 @@ def append_to_newsfile(
     else:
         existing_content = [""]
 
-    if top_line and top_line in existing_content:
+    if top_line and top_line in existing_content[-1]:
```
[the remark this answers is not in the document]
mikesongming: It should be ok. 

I guess you mean "start marker" as the head line "Release Notes" in front of the `start_string` in my test case? 
Similar changelog setting also appears in the test case `test_write.WritingTests.test_append_at_top_with_hint`:

```
        with open(os.path.join(tempdir, "NEWS.rst"), "w") as f:
            f.write(
                "Hello there! Here is some info.\n\n"
                ".. towncrier release notes start\nOld text.\n"
            )
```

Whether have a "start marker" or not, the struct of `single_file` newsfile would be:
```
[start marker]

[start string]

top_line     # for newer version 
content     # for newer version

top_line     # for older version
content     # for older version

...
```
After the clause `existing_content = existing_content.split(start_string, 1)`, `existing_content` becomes a list:

```
[ start_marker, start_string, previous_changes ]
```
or just `[previous_changes ]`.

In either case, the original predicate `top_line in existing_content` always fails and the supposed raise never happens. That's why I use `top_line in existing_content[-1]` to judge if `top_line` appears in previous content.

src/towncrier/test/test_build.py
mikesongming: It is really tedious 😂

mikesongming:
@adiroiban All suggested changes in upon comments have been updated, and a detailed explanation about `single_file` is also updated in README.rst. 

Question about `if top_line and top_line in existing_content[-1]:` part is replied in the comments.

All existing tests passed in the workflow.

adiroiban (approved):
many thanks for the wonderful work.

Hope all is green and I will merge.

mikesongming:
glad to make a contribution to towncrier :)

PR #392, opened 2022-06-27
Enforce PR failure based on a single job and fail the job if a dependent job was skipped.

adiroiban:
This tries to work around an issue with GitHub.

We have a PR configured to allowing merging only if Job A is green.

If you have a job A depending on another job B, if job B fails, job A is skipped.

But with job A skipped, GitHub is happy and it allows the PR to be merged.

Implement the nice move from @altendky from https://github.com/Chia-Network/chia-blockchain/pull/11880/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,7 +236,7 @@ jobs:
         verbose: true
 
   all:
-    name: All
+    name: All success
     runs-on: ubuntu-latest
     needs:
       - build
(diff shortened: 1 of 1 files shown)


adiroiban:
This PR is started without a news fragment to test that the `catch-all` job fails.

adiroiban:
@altendky @graingert  do you have time to take a look and see if we can merge this ?

Thanks :)

altendky (approved):
It reads as I would expect.  I look forward to hearing nothing more about this because it works.  `:]`  `*fingers crossed*`  Thanks.

webknjaz:
Hey, I noticed you're reinventing the wheel here :)
https://github.com/re-actors/alls-green

PR #396, opened 2022-07-16
Run check-manifest and other linters via GitHub Actions.

adiroiban:
# Description

This is take2 for #292 

Enabled check manifest and pre-comit as part of GitHub Actions CI

We have pre-commit hook so I am not sure if this is needed.

I  tried to simplify the job and removed the job matrix.

Fixes #292

# Checklist

* [NA] Make sure changes are covered by existing or new tests.
* [NA] For at least one Python version, make sure local test run is green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,39 +151,21 @@ jobs:
       run: |
         codecov -n "GitHub Actions - ${{ matrix.task.name}} - ${{ matrix.os.name }} ${{ matrix.python.name }}"
 
-  check:
-    name: ${{ matrix.task.name}} - ${{ matrix.python.name }}
+  lint:
+    name: Linters
     runs-on: ubuntu-latest
-    needs:
-      - build
-    strategy:
-      fail-fast: false
-      matrix:
-        python:
-          # Using second most recent minor release for whatever little
-          # increase in stability over using the latest minor.
-          - name: CPython 3.9
-            tox: py39
-            action: 3.9
-        task:
-          - name: Check Newsfragment
-            tox: check-newsfragment
+    env:
+      TOX_PARALLEL_NO_SPINNER: 1
 
     steps:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
-    - name: Download package files
-      uses: actions/download-artifact@v3
-      with:
-        name: dist
-        path: dist/
-
-    - name: Set up ${{ matrix.python.name }}
+    - name: Set up python
       uses: actions/setup-python@v3
       with:
-        python-version: ${{ matrix.python.action }}
+        python-version: "3.10"
 
     - name: Install dependencies
       run: python -m pip install --upgrade pip tox
@@ -191,7 +173,14 @@ jobs:
     - uses: twisted/python-info-action@v1
 
     - name: Tox
-      run: tox -c tox.ini -e ${{ matrix.task.tox }}
+      run: tox -c tox.ini --parallel -e pre-commit
+
+    - name: Tox
+      run: tox -c tox.ini --parallel -e check-newsfragment
+
+    - name: Tox
+      run: tox -c tox.ini --parallel -e check-manifest
+
 
   pypi-publish:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.rst
```diff
@@ -68,12 +69,14 @@ We recommend the following workflow:
       *Running the test suite* for details.
 
 
-   c. Document any user facing changes in one of the `/docs/` files.
+   c. Document any user-facing changes in one of the `/docs/` files.
```
altendky: I think?

```suggestion
   c. Document any user-facing changes in one of the ``/docs/`` files.
```

tox.ini
```diff
@@ -33,37 +33,4 @@ commands =
    # specifying the entry point script in `{envbindir}`.
    coverage run -p --module twisted.trial {posargs:towncrier}
    coverage combine -a
-
-[testenv:build]
```
altendky, line 37: Why is this going away?

setup.cfg
```diff
@@ -0,0 +1,3 @@
+[flake8]
+# Allow for longer test strings. Code is formatted to 88 columns by Black.
+max-line-length = 99
```
altendky, line 3: I'm still in the boat of using the program's default config file.  Even when using the 'single config file for everything' option you still end up with multiple single config files.

.github/workflows/ci.yml
```diff
@@ -151,47 +151,36 @@ jobs:
...
 
     - name: Tox
-      run: tox -c tox.ini -e ${{ matrix.task.tox }}
+      run: tox -c tox.ini --parallel -e pre-commit
+
+    - name: Tox
+      run: tox -c tox.ini --parallel -e check-newsfragment
+
+    - name: Tox
+      run: tox -c tox.ini --parallel -e check-manifest
```
altendky: Use the relevant names for the step name.  Why parallel for single environments?  How about `if: always()` so you get all results in a single run rather than having to iterate through multiple failures via multiple commits and ci runs.

.github/workflows/ci.yml
```diff
@@ -151,47 +151,36 @@ jobs:
...
-      with:
-        name: dist
-        path: dist/
-
-    - name: Set up ${{ matrix.python.name }}
+    - name: Set up python
       uses: actions/setup-python@v3
       with:
-        python-version: ${{ matrix.python.action }}
+        python-version: "3.10"
```
altendky: These days I tend to keep my checks one version behind (3.9 not 3.10).  They are there to check the code, not see if the tools work with the latest Python.  But definitely not a big deal, just sharing a thought.

altendky (commented):
Not my personal CI style (https://github.com/twisted/towncrier/pull/292#issuecomment-739961496) but it'll work.  Here's a few thoughts.

tox.ini
adiroiban: my bad. I thought that we don't need it and that it was already moved to GitHub CI only.

I will revert... 

My hope was that we can use standard tools without the need of extra tools.

.github/workflows/ci.yml
adiroiban: I will revert. thanks. no big preference.
Just trying to reduce the number of redirections :)

setup.cfg
adiroiban: The plan it to move more things from setup.py into setup.cfg ... I am using it in #398

.github/workflows/ci.yml
```diff
@@ -163,10 +163,14 @@ jobs:
...
             tox: py39
-            action: 3.9
+            python-version: '3.9'
         task:
           - name: Check Newsfragment
             tox: check-newsfragment
+          - name: Check package manifest
+            tox: check-manifest
+          - name: Check pre-commit
+            tox: pre-commir
```
adiroiban:
```suggestion
            tox: pre-commit
```

adiroiban:
Thanks for your help.

In the end, this is not much better.

it just helped me understand the (what I consider convoluted) way in which towncrier CI works... with all the build phase and tox_build.sh external script. 

I would still like to merge it as I hope that in the future we can update the CI and remove the build and tox_build.sh script.

setup.cfg
altendky: Any project using black will still have to have a `pyproject.toml` for any config there, afaik.  Anyways, not a bug, just a passing opinion.

tox.ini
altendky: I wasn't sure where you were going so maybe it isn't needed now?  It was an honest question.

adiroiban: This is still need.
The towncrier is built once.
Then the wheel is uploaded as an artifact.
Then the tests will download the artifact and run the tests agains it.

------------

I am not convinced that we need all this dance... but I think is better to refactor this in another PR.

adiroiban:
I have enbabled auto-merge for this.

There are some README updates...and no other big changes.

PR #398, opened 2022-07-16
Remove own usage of incremental.

### LINKED ISSUE #308: Remove dependency on incremental in favor of general compatible solution
> 
> 
> 
> 
> 
> 

adiroiban:
# Description
Fixes #308

To simplify the number of dependencies towncrier can use a string as its version.

Also, when finding the version of a project, just use `str()` on the project version instancce. 

project using `incremental` are still supported, as str(incremental.Version) does not return what we want... and is ok to keep incrementla for backwards compatibility.

# Todo

> **For anything else, just call str(version_object) and strip it.**

# Checklist


hynek:
Do you still want to do this Adi? Or wanna start fresh? (I support the change, but this branch is _very_ stale)

adiroiban:
I still would like to have a simple string as the version for towncrier.
Right now, I don't have the motivation and time to continue the work.

adiroiban:
This can be closed as there is https://github.com/twisted/towncrier/pull/491

PR #399, opened 2022-07-16
Update workflow trigger to new trunk branch name

altendky:
# Description
The main branch was renamed from `master` to `trunk`.  This is just catching the workflow triggers up with that.

# Checklist

* [ ] ~Make sure changes are covered by existing or new tests.~
* [ ] ~For at least one Python version, make sure local test run is green.~


.github/workflows/ci.yml
```diff
@@ -2,10 +2,10 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ trunk ]
     tags: [ "**" ]
   pull_request:
-    branches: [ master ]
+    branches: [ trunk ]
```
adiroiban: Maybe allow PR  to any branch... to simplify the CI config.
```suggestion
```

adiroiban (approved):
Thanks. Nicely spotted :)

.github/workflows/ci.yml
altendky: Yes, I do actually quite prefer that behavior, even aside from the simplification.

PR #402, opened 2022-08-23
Uninvent the wheel w/ `re-actors/alls-green` @ GHA

webknjaz:
# Description
$sbj.

Ref: https://github.com/twisted/towncrier/pull/392#issuecomment-1224220279

# Checklist


adiroiban (commented):
Thanks for the PR.

-----

i am not sure if this is needed.

I am -1 for this.

I think that the current code just works, and is easy to audit.

What is wrong with the current code?

Why we need to use an external code?

What we gain  by using the external code?

Having one more extra dependency for which a security audit is required is just one extra liability.

--------

In light of all the NPM security issues and all supply chain attacks,
 when possible, I prefer a bit of duplication and reduce the risk of depending on external code that might introduce a vulnerability.

---------

But if others think that this is a good idea, I am not going to block the merge.

webknjaz:
> Having one more extra dependency for which a security audit is required is just one extra liability.

FWIW, if you care about security, all the external actions should be pinned using SHA-1s.

adiroiban:
Good point. Thanks. I think that we can use the all-green shared code with a SHA pin.

And also pin the pypi upload code.

For the oficial GitHub actions and our twisted action I think that we should be ok.

Thanks again!

webknjaz:
> And also pin the pypi upload code.

I authored that one too! Consider me your supply chain :)

webknjaz:
> For the oficial GitHub actions and our twisted action I think that we should be ok.

By the way, it is possible to set up an allow-list of actions now. See https://github.com/twisted/towncrier/settings/actions.

adiroiban (approved):
Thanks for the update.

I hope all tests will be green and we can merge this.

Issue #408, opened 2022-08-30
Fix the build of docs on RTD


adiroiban:
It looks like the version is required for Sphinx

adiroiban:
If was fixed in #409

PR #409, opened 2022-08-30
#408 Fix Read the docs build.

adiroiban:
# Description

The build on RTD for trunk fails due to missing version in docs/conf.py 

# Changes

The version was previously removed as it was a static version.
It was not replaced with the real version.

The copyright was also updated to include the version and build time.

The RTD was updated not to generate PDF and EPUB.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,6 +29,11 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
+from datetime import date
+
+from towncrier import __version__ as towncrier_version
+
+
 extensions = []
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -44,10 +49,25 @@
 master_doc = "index"
 
 # General information about the project.
+_today = date.today()
 project = "Towncrier"
-copyright = "2017, Amber Brown"
+copyright = "{}, Towncrier contibutors. Ver {}. Built on {}".format(
+    _today.year,
+    towncrier_version.public(),
+    _today.isoformat(),
```
hynek: I believe this is problematic, because it doesn't allow for reproducible builds. Maybe git ref instead?

adiroiban: Thanks for the review.

Is this a big problem for the docs?
In practice, I expect that we generate the docs in Read the Docs once, and forget about it.

This was copied from twisted/twisted.

If this is good enough, I prefer to merge it as it is.

exarkun: If the build is reproducible then it won't matter when it was built because whenever it was built, you'll get the same output. :)  This is a pretty nice thing.

If the version of the docs being built is available then I don't think you should need to know the date.

I don't know if the doc build is reproducible even _without_ this change so I don't know how much it matters now - but I would expect reproducibility to be a goal that is eventually worked towards and achieving it will require undoing this change, so perhaps it's better to just remove the date now.

docs/conf.py
```diff
@@ -44,10 +49,25 @@
 master_doc = "index"
 
 # General information about the project.
+_today = date.today()
 project = "Towncrier"
-copyright = "2017, Amber Brown"
+copyright = "{}, Towncrier contibutors. Ver {}. Built on {}".format(
```
exarkun: Typo in "contributors" here.

docs/conf.py
adiroiban: I have removed the build date.

We still have the copyright year.

I have also fixed the typo.

I hope this is ready to merge.

hynek: The year isn't [really necessary either](https://hynek.me/til/copyright-years/), but that takes things too far for this PR

hynek (approved):

adiroiban:
Thanks for your reviews.

I think we are in good shape.

Happy to see the dev docs available as part of the Read The Docs build.
it should make it easier to read and review the content in the future.

hynek:
This is a good moment to switch the default to "stable" instead of "latest"?

adiroiban:
Done. Thanks for the headup.

PR #413, opened 2022-09-01
Remove non-existant tox env

hynek:
So I grew a bit irritated by the contributor experience while fixing up 400 so this is the beginning of _Hynek's Fall of Developer Experience_ – sorry Adi. ;)

My goal is to give contributors a much faster feedback cycle and to simplify some things along the way.


adiroiban (approved):
Thanks!

adiroiban:
I don't really use tox for local dev... so I don't have much love for it.
Thanks for the cleanup.

hynek:
Yeah but that's my main beef: per default you can't run the tests without it. Gonna try to fix it.

Issue #410, opened 2022-08-31
Code coverage via GitHub action


adiroiban:
This is an experiment.

I don't have time now to work on it, but I left a few comments/links

https://github.com/ewjoachim/python-coverage-comment-action

----------

The idea is to generate the coverage report only using GitHub Actions

This includes the commit status and comment with the report.

Basically, reinvent codecov.io

Some ideas here 

https://hynek.me/articles/ditch-codecov-python/

https://itnext.io/github-actions-code-coverage-without-third-parties-f1299747064d

https://josh-ops.com/posts/github-code-coverage/

https://dev.to/thejaredwilcurt/coverage-badge-with-github-actions-finally-59fa

PR #414, opened 2022-09-01
Refactor _get_default_compare_branch

hynek:
I couldn't live with myself contributing monkeypatching a mock, so here's a clean fix:

Make it a computation instead of an action and drop the patch/mock.

### DIFF
--- a/src/towncrier/check.py
+++ b/src/towncrier/check.py
@@ -19,11 +19,13 @@ def _run(args, **kwargs):
     return check_output(args, **kwargs)
 
 
-def get_default_compare_branch(base_directory, encoding):
-    branches = (
-        _run(["git", "branch", "-r"], cwd=base_directory).decode(encoding).splitlines()
-    )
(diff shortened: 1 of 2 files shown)


adiroiban (approved):
awesome.
Happy to see unittest.mock removed.

hynek:
Is coverage failing because there's fewer lines? Can't figure it out. 😳

hynek:
I see the coverage apparently went down in test_check, so merging.

adiroiban:
Coverage was reported twice.

The overall project coverage went down.
This is expected when lines that previously had coverage are removed.
And all the not-covered lines are kept.
The percentage of covered lines is now less.

I never block a merge on project coverage.
I think it has no value in helping with branch protection.

On twisted/twisted we allow for some variation ...so even if it drops 0.02% we still show it as green.

--------

We will have this problem, even if we use our own coverage #410

PR #415, opened 2022-09-01
Add missing local test dep and document it

hynek:
Now it's possible to just run `trial towncrier`.

I've also removed some blatant lies from the contributing guide :D


adiroiban (approved):
Thanks

PR #418, opened 2022-09-02
Bring coverage of _git.py and create.py to 100%

hynek:
I've also taken the liberty to refactor git handling in check a little bit, such that _all_ non-test git calls live in `_git.py` together.

**edit**: since the change to create is so trivial, I've folded it in here.


src/towncrier/check.py
```diff
@@ -81,20 +71,16 @@ def __main(comparewith, directory, config):
     encoding = getattr(sys.stdout, "encoding", "utf8")
```
twm: This doesn’t do what the comment above says it does, right? If `sys.stdout.encoding is None` then `encoding` will be `None`.

twm (approved):
Changes here look great! I had one unrelated comment.

src/towncrier/check.py
hynek: Yeah it absolutely doesn't. I'll remove the construct, because we don't do Python 2 anymore.

hynek:
Here's the output of `coverage run -m twisted.trial towncrier; and coverage combine; coverage report`

```
Name                              Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------
src/towncrier/_project.py            38      6     16      5    76%   45, 51, 56, 71, 76-78
src/towncrier/_settings/load.py      71      1     34      2    97%   28->31, 36
src/towncrier/_writer.py             23      0     18      1    98%   36->39
src/towncrier/check.py               61      2     24      1    96%   108-109
-----------------------------------------------------------------------------
TOTAL                               552      9    228      9    97%
```

Can we relax the merge rules while codecov is on strike?

hynek:
Screenshot 2022-09-02 at 09 52 30@2x

liarsssssss

adiroiban:
> liarss

I think the issue is on our side.

I guess that codecov.io was so unstable in the past, that someone decided to ingore codecov.io errors.

In the logs I see a failure on codecov.io reporting, but the CI step is green.

I was expecting the CI step to fail.

https://github.com/twisted/towncrier/runs/8152641264?check_suite_focus=true#step:8:25

PR #419, opened 2022-09-02
Refactor _writer & bring to 100% coverage

hynek:
I've been trying to understand _writer.py and while doing so, I've refactored it into something that fits my brain better.

Turned out, the branch that was missing from coverage can't happen with the current implementation (see comment for explanation).

### DIFF
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -24,7 +24,9 @@ def append_to_newsfile(
     """
     news_file = os.path.join(directory, filename)
 
-    header, prev_body = _load_existing_content(news_file, start_string, single_file)
+    header, prev_body = _figure_out_existing_content(
+        news_file, start_string, single_file
+    )
 
     if top_line and top_line in prev_body:
         raise ValueError("It seems you've already produced newsfiles for this version?")
@@ -49,7 +51,7 @@ def _write_news(f, header, start_string, new_content, old_body):
         f.write(f"\n\n{old_body}")
 
 
-def _load_existing_content(news_file, start_string, single_file):
+def _figure_out_existing_content(news_file, start_string, single_file):
     """
     Try to read *news_file* and split it into header (everything before
     *start_string*) and the old body (everything after *start_string*).
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -34,13 +34,12 @@ def append_to_newsfile(
             f.write(header)
             if start_string:
                 f.write("\n\n" + start_string + "\n")
-            else:
-                xxx
 
         f.write(content)
+
         if old_body:
             f.write("\n\n")
-        f.write(old_body.lstrip())
+        f.write(old_body)
 
 
 def _load_existing_content(news_file, start_string, single_file):
@@ -62,6 +61,6 @@ def _load_existing_content(news_file, start_string, single_file):
 
     t = content.split(start_string, 1)
     if len(t) == 2:
-        return t[0].rstrip(), t[1]
+        return t[0].rstrip(), t[1].lstrip()
 
-    return "", content
+    return "", content.lstrip()
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -11,7 +11,7 @@ import os
 
 
 def append_to_newsfile(
-    directory, filename, start_string, top_line, content, single_file=True
+    directory, filename, start_string, top_line, content, single_file
 ):
 
     news_file = os.path.join(directory, filename)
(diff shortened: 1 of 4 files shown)


hynek:
Please don't merge yet I've had an epiphany at the gym how simplify further :)

hynek:
ok, it's ⚡️perfect⚡️ now – pls review 😇

src/towncrier/test/test_write.py
```diff
@@ -109,6 +109,7 @@ def test_append_at_top(self):
                 wrap=True,
                 versiondata={"name": "MyProject", "version": "1.0", "date": "never"},
             ),
+            True,
```
adiroiban: I think that using keyword arguments makes the test a bit easier to read.

```suggestion
            single_file=True,
```

src/towncrier/test/test_write.py
```diff
@@ -221,6 +222,7 @@ def test_append_at_top_with_hint(self):
...
```
adiroiban:
```suggestion
            single_file=True,
```

src/towncrier/_writer.py
```diff
@@ -11,33 +11,55 @@
...
-    if top_line and top_line in existing_content[-1]:
+    header, prev_body = _figure_out_existing_content(
+        news_file, start_string, single_file
+    )
+
+    if top_line and top_line in prev_body:
         raise ValueError("It seems you've already produced newsfiles for this version?")
 
-    with open(os.path.join(directory, filename), "wb") as f:
+    with open(news_file, "w", encoding="utf8") as f:
```
adiroiban: I guess that this might cause the failure on windows.
I prefer to always keep the file in binary mode so that Python will not mess with newlines.

adiroiban (approved):
Thanks for the update.

Once the windows test are green, this should be merged.

src/towncrier/_writer.py
hynek: doesn't seem to help 🤔 https://github.com/twisted/towncrier/pull/419/commits/ee54060b04fc99f525cbc10b9b8ebcfa1d4a54c3

I'll to keep digging, this makes no sense.

hynek: ok this is funny, because now a different test is failing – one that _relied_ on newline magic: https://github.com/twisted/towncrier/runs/8167043027?check_suite_focus=true

this is gonna be a long morning

hynek:
OK so here's the deal:

_Writing_ it out without touching newlines seems to be the correct way of handling things. Otherwise some `\r` are converted. The open keyword for that is `newlines=""`.

Since `_builder.find_fragments` doesn't convert newlines, because it loads using bytes, we get `\r\n` in the fragments coming into `append_to_newsfile`

However.

_Reading_ an existing news file while leaving newlines in place breaks the detection of `start_string` whose default includes a `\n`: `".. towncrier release notes start\n"`

If we read a file with `\r\n` it doesn't match, because the line becomes `.. towncrier release notes start\r\n`.

---

~It's a total mystery to me how this worked until now. But I'll try to figure it out.~

OK I get it. `append_to_newsfile` actually used unicode/universal newlines for reading, but not for writing. So fixing this is easy, I'm just not sure this whole thing makes a lot of sense.

hynek:
OK, this is done as far as I'm concerned.

I'm deeply unhappy with what this refactoring has unearthed, but that's a topic for another day.

adiroiban (approved):
Thanks.

I think that is good that you found this bug.
And now we have better documentation/comments for towncrier's behaviour.

A few weeks ago I had no idea that `single_file` is a thing and that people might want to use it.

But, with GitHub Release automate release notes generation providing a simple alternative to towncrier, I guess that for towncrier to continue to stay relevant it shouod support all kind of different usage scenarios.

src/towncrier/test/test_write.py
```diff
@@ -18,6 +18,8 @@
 
 
 class WritingTests(TestCase):
+    maxDiff = None
```
adiroiban, line 21: I guess we can leave this as it is. No preference.

adiroiban:
Thanks, @hynek for the help with the coverage.

I am using something like this for codecov.yaml

```
coverage:
  precision: 2
  round: down
  range: 95..100
  status:
    patch:
      default:
        target: 100%
    project:
      default:
        target: auto
        # Temporary allow for a bit of slack in overall code coverage due to
        # swinging coverage that is not triggered by changes in a PR.
        threshold: 0.02%
      tests:
        target: 100%
        paths:
          - "src/towncrier/test/*"

```

It does 3 things:

* Make sure new code / modified code has at least 100% coverage :)
* Allows some slack to general project reporting, If one env (like macOS or Win) goes down we will see a bigger drop so this will get red
* Makes sure all the testing code will always have 100% - no point in keeping testing code that is not executed by the CI. This helps with the test code cleanup.

PR #425, opened 2022-09-06
Fix documentation of fragment configuration in README

### LINKED ISSUE #424: Configuration of fragments in readme is incorrect
The readme suggest that fragments should be configured in the form

```
[tool.towncrier.feat]
[tool.towncrier.fix]
```
However both the code and https://towncrier.readthedocs.io/en/stable/configuration.html shows that the correct format is

```
[tool.towncrier.fragment.feat]
[tool.towncrier.fragment.fix]
```

jenshnielsen:
# Description

Update to match what the code expects and how this is documented
in https://towncrier.readthedocs.io/en/stable/configuration.html

Fixes #424 

# Checklist

* [ ] Make sure changes are covered by existing or new tests. N/A Readme only change 
* [ ] For at least one Python version, make sure local test run is green. N/A readme only change
* [x] Make sure all GitHub Actions checks are green (they are automatically checking all of the above). N/A Readme only change

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -173,12 +173,12 @@ Furthermore, you can customize each of your own fragment types using:
 
     [tool.towncrier]
 	# To add custom fragment types, with default setting, just add an empty section.
-    [tool.towncrier.feat]
-    [tool.towncrier.fix]
+    [tool.towncrier.fragment.feat]
+    [tool.towncrier.fragment.fix]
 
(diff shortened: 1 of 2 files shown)


adiroiban (approved):
Thanks. Much appreciated.

I will merge this as this is better than what we have in trunk.

Another option is not to have any example config inside README and just link to the docs.

But can be done in another PR.

Issue #365, opened 2022-01-22
[docs] Please document how to generate new entries that are not tied to an existing issue/PR


offbyone:
I don't want to have the bookkeeping requirement of creating a github issue for everything I mention in the changelog... but I'm not sure what those files need to look like.

adiroiban:
Hi. I don't understand the issue.

The towncrier release notes "fragments" are not tied to GitHub issues/PR

You can create fragments using Jira/Trac or any system.

-------

A release announcement has the release notes groped into various categories.

To help with support and make it easy to find more detatils about a certain released featured or bug, there is a issue/ticket ID associated.

The category and ticket ID meta data are encoded in the release fragment file name.

I guess that you want to create "fragments" with just the category marker and without the issue marker.

I don' think that this is a common usecase and is not supported for now.
Feel free to send a PR.

One workaround is to use '0' as the ticket id.
Then you can have a custom template that will skip the link for ticket id 0, but include the release fragment content.

Hope it helps

offbyone:
My question is, how to create fragments that are not tied to _any_ issue tracker. Suppose I'm just working on code for myself, where I want the changelog, but I don't use an issue tracker. How do I generate fragments for that?

adiroiban:
maybe something like this?

You will still need a "code" for each release note item.


you can configure the output not to include a link 

```diff
diff --git a/pyproject.toml b/pyproject.toml
index df86fa4..df7096b 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
     package = "towncrier"
     package_dir = "src"
     filename = "NEWS.rst"
-    issue_format = "`#{issue} <https://github.com/hawkowl/towncrier/issues/{issue}>`_"
+    issue_format = "{issue}"
```

have a file like

```
$ cat src/towncrier/newsfragments/SOME-FEATURE.removal.rst
Some text for this new feature to be released
```

the result is  like

```
$ towncrier build --draft
Loading template...
Finding news fragments...
Rendering news fragments...
Draft only -- nothing has been written.
What is seen below is what would be written.

towncrier 21.9.0 (2022-06-10)
=============================

Deprecations and Removals
-------------------------

- Some test (SOME-FEATURE)
```

SmileyChris:
[...]

PR #428, opened 2022-09-15
Orphan news fragments

SmileyChris:
# Description

In addition to having fragments that are linked to issues/unique identifiers, it would be nice to allow "unlinked" fragments.

For example, a changes directory with the following files:

```
101.feature -> "Great feature"
102.feature -> "Added this too"
+5fe6f5c5.feature -> "Some small feature that doesn't have a related issue"
```

...will output

```
...
Features
--------

- Great Feature (#101)
- Added this too (#102)
- Some small feature that doesn't have a related issue
```

Anything after a `+` means that it's an unlinked fragment, so the rest of the base filename is ignored. `towncrier create +.feature` adds a random hex after the base name for you.

It's been on my mind for a while, and I started discussing it in #365. Code speaks louder than words though :)

# Checklist


exarkun:
How does this interact with `check`?  Can a project opt out of allowing this fragment type?

SmileyChris:
I don't see how it would change the functionality of `check` in any way. It's just a special case for the base name of the fragment - it still follows the standard crier rules the extension being a valid category.

I didn't make any configuration option for this currently. Could be added if that was a requirement (maybe defining the "magic character", and if falsey, then it's a disabled option?)

src/towncrier/test/test_builder.py
```diff
@@ -81,3 +81,9 @@ def test_strip_with_counter(self):
             parse_newfragment_basename("  007.feature.3", ["feature"]),
             ("7", "feature", 3),
         )
+
+    def test_unlinked_feature(self):
```
adiroiban: a docstring describing why this tests exist , and what is the input and expect result would be great.

The docstring is the only way we can test that the test is ok :)

Without a docstring, a test can do anything and still be a "good" test :)

src/towncrier/test/test_create.py
```diff
@@ -182,3 +182,22 @@ def test_file_exists(self):
 
         self.assertEqual(type(result.exception), SystemExit)
         self.assertIn("123.feature.rst already exists", result.output)
+
+    def test_create_unlinked_fragment(self):
+        """Ensure a random hash is added to unlinked fragment files."""
+        runner = CliRunner()
```
adiroiban: I prefer to write the test docstrings as test scenarios.

"ensure" is kind of redendand in the context of a test.
We always write tests to ensure something.

No need to be too specific in the docstring.
A random is good enough... it can be hex, but maybe later we want to have some random names (like docker does :)

```suggestion
    def test_create_unlinked_fragment(self):
        """
        When called for a ticket ID with value C{+}, the create CLI
        automatically expands the new file to a random value.
        """
        runner = CliRunner()
```

docs/quickstart.rst
```diff
@@ -76,6 +76,10 @@ We can create some example news files to demonstrate::
     $ echo 'The final part is ignored, so set it to whatever you want.' > src/myproject/newsfragments/8765.removal.txt
     $ echo 'misc is special, and does not put the contents of the file in the newsfile.' > src/myproject/newsfragments/1.misc
 
+If the news fragment doesn't need to be linked to any ticket ID or other identifier, start the file name with ``+``::
```
adiroiban: Maybe 

```suggestion
If the news fragment doesn't need to be linked to any ticket ID or other identifier, start the file name with ``+``.
The content will still be included in the final release notes,
in the category corresponding to the file extension::
```

README.rst
```diff
@@ -93,6 +93,10 @@ For example::
 
     towncrier create 123.feature
 
+If a news fragment is not tied to an issue, use `+` as the basename (a random hash will be added to the filename to keep it unique):
```
adiroiban, line 96: Do we need to also talk about this in the README?

The risk is updating the implementation in the future, and forgeting to update the docs in all places.

I think that keeping README short is a good goal.

We have the docs for a full explanation of what towncrier can do.

src/towncrier/newsfragments/428.feature
```diff
@@ -0,0 +1 @@
+Ability to create fragments not tied to issues via a leading `+` in the filename (e.g. `+anything.feature`).
```
adiroiban: Maybe a bit more explicit, and in the "tone" of a release note.

```suggestion
You can now create fragments that are not associated with issues. Start the name of the fragment with `+` (e.g. `+anything.feature`). The content will be included in the final release notes.
```

src/towncrier/_builder.py
```diff
@@ -38,6 +38,8 @@ def parse_newfragment_basename(
     if len(parts) == 2:
         ticket, category = parts
         ticket = strip_if_integer_string(ticket)
+        if ticket.startswith("+"):
+            ticket = ""
```
```suggestion
        if ticket.startswith("+"):
            # Tickets with names/IDs starting with a + are considered
            # orphan fragments.
            # Just use an empty ticket ID so that we can continue
            # procesing it.
            ticket = ""
```

docs/quickstart.rst
adiroiban: Maybe called them "orphan" news fragments in the docs?

Not bug deal, but might help in the future with bug reporting and help us have simple word to describe this feature.

src/towncrier/_builder.py
```diff
@@ -90,6 +92,7 @@ def find_fragments(
     """
     content = OrderedDict()
     fragment_filenames = []
+    unlinked_fragment_counter: DefaultDict[str | None, int] = defaultdict(int)
```
adiroiban: I think that this needs a docstring/comment.

```suggestion
    # ADD A COMMENT WHY WE NEED THIS :)
    unlinked_fragment_counter: DefaultDict[str | None, int] = defaultdict(int)
```

src/towncrier/_builder.py
```diff
@@ -176,14 +184,13 @@ def split_fragments(
...
 
-            if texts.get(content):
-                texts[content] = sorted(texts[content] + [ticket])
-            else:
-                texts[content] = [ticket]
-
-            section[category] = texts
+            tickets = texts.setdefault(content, [])
+            if ticket:
+                # Don't add the ticket if it's blank (meaning it's an unlinked fragment).
```
adiroiban: mabye something like this

the `if ticket` and then a comment about `if not ticket` is a bit confusing.
 
```suggestion
            if not ticket:
                """
                Don't add the ticket if we don't have one.
                This can happen for unlinked / orphan fragments.
                """
                continue
```

src/towncrier/_builder.py
```diff
@@ -201,9 +208,10 @@ def issue_key(issue: str) -> tuple[int, str]:
         return (-1, issue)
 
 
-def entry_key(entry: tuple[str, Sequence[str]]) -> list[tuple[int, str]]:
-    _, issues = entry
-    return [issue_key(issue) for issue in issues]
+def entry_key(entry: tuple[str, Sequence[str]]) -> tuple[str, list[tuple[int, str]]]:
+    content, issues = entry
+    # Empty issues (unlinked fragments) should sort last by content.
```
adiroiban: Not sure why... but I am not -1 for this :)

adiroiban:
@exarkun  towncrier check should be fine. It only checks for file type / file category and not for file ticket ID.

This is already supported

```
$ towncrier check --compare-with trunk
Looking at these files:
----
1. /home/adi/chevah/towncrier/src/towncrier/newsfragments/+bla.feature
2. /home/adi/chevah/towncrier/file.txt
----
Found:
1. /home/adi/chevah/towncrier/src/towncrier/newsfragments/+bla.feature
```

adiroiban (changes requested):
Thanks for the PR.

Great work.

If you care that much about this feature, I think we can have it in the "official" template.

----------

I think I have already implemented something like this for my project... but never ended up using it...as everything should have a ticket :)

My implementation was using a ticket id starting with `0`.

So you can have `01.feature` and `02.feature`  .... and the custom template will just skip rendering the tickets ID if they start with `0`.

But I am fine "standardizing" this functionalty... I think is a sign of bad practice... but who am I to judge what is good and what is wrong :)

-------

The only big worry is about the "+" character.

Maybe there are trackers in the wild for which the ticket ID start with a `+`.
I don't know much about bug trackers. Have only used Trac and GitHub... but I know that Jira allows you to start a ticket ID with some characters.


But I think we can go with "+" and if there is an issue, add a configuration option to disable this feature, or use a different marker for orphan fragments.

Maybe instead of `+` we can use `orphan-RANDOM-VALUE.type` as the file name on disk.

The `create` CLI can still allow `+.type` but when writing the actual file, the file is wrote a `orphan-RANDOM.type` 

Just an idea.

I think the PR looks very good.

Only a few minor comments blocking the final merge.

--------

I have used "orphan" but feel free to continue to use "unlinked" if you prefer.

SmileyChris:
I'm totally fine to have the name bikeshedded. Orphan might be better :)

I added in a configuration option so if you used some weird tracker that had a format starting with `+` then you can configure a different prefix (e.g. `"orphan-"`), or set it to `""` to disable this logic altogether.

README.rst
SmileyChris: I agree that it's probably not the best place, but I think the reason I put it here is there *isn't* any other documentation section on `create`.

src/towncrier/_builder.py
SmileyChris: It's exlpained when we use it, but I'll add something...

src/towncrier/test/test_builder.py
SmileyChris: parse_newfragment_basename doesn't do the orphan magic now, so this test was removed.

src/towncrier/test/test_create.py
SmileyChris: Done

docs/quickstart.rst
SmileyChris: I've consistently implemented the wording of "orphan news fragments" across the changes now.

src/towncrier/_builder.py
SmileyChris: Code not done here any more, and where it was moved to is more explicit.

src/towncrier/_builder.py
SmileyChris: Clarified.

src/towncrier/_builder.py
SmileyChris: 👍🏻 It means we're consistent, and I updated the docs clarify that this is the behaviour.

src/towncrier/newsfragments/428.feature
SmileyChris: Improved this.

adiroiban (approved):
Looks good to merge.
Awesome work. Thanks!

exarkun:
> I don't see how it would change the functionality of `check` in any way. It's just a special case for the base name of the fragment - it still follows the standard crier rules the extension being a valid category.
> 
> I didn't make any configuration option for this currently. Could be added if that was a requirement (maybe defining the "magic character", and if falsey, then it's a disabled option?)

It seems like it wasn't a requirement :)

To elaborate on my concern, currently a project may be using `towncrier check` to ensure there is a newsfragment _and_ a ticket.  Once news fragments don't require a t icket, `towncrier check` only ensures there is a newsfragment so the "and a ticket" part of the check has been lost.

On further reflection, I'm not sure `towncrier check` ever _actually_ ensured "and a ticket".  And now, looking at the implementation, I see that it was (and is) satisfied by _any_ news fragment being added.  So, indeed, this PR doesn't make any difference to the feature I was concerned with.

Thanks for your contribution!

adiroiban:
I think the current state of `towncrier check` is something like this.

It checks to see that you have not forgot to add a newsfragment.
It checks that the fragment is for a known category / type.

It doesn't check the validity of the content, or the ticket ID

---------

I guess that we can have a future PR to add support for checking the ticket ID format,
and as part of that, add a rule whether only orphan fragments are fine, or a branch MUST have at least one ticket ID fragment.

PR #429, opened 2022-09-16
Make configuration a proper class, push settings & check coverage to 100%

hynek:
# Description

It's time to say goodbye to the ungodly dictionary everywhere.

A few notes:

- I'm hereby proposing the `with_isolated_runner` decorator because the majority of our tests have the repetition of creation a `click.Runner` and then adding an unnecessary indentation level. I haven't applied it to other tests, but I think this would improve our test suit immensely.
- A few functions too a `definitions: list[str]` and used it as such, but definitions aren't a list, they're a dict and it worked by accident. I've renamed the argument to reflect what it actually is and just pass `definitions.keys()` into it.
- this PR is coming a day late, thanks for the merge conflicts #428 🤪

# Checklist

### DIFF
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -7,17 +7,47 @@ import os
 import sys
 
 from collections import OrderedDict
-from typing import Any, Mapping
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Mapping
 
 import pkg_resources
 
 from .._settings import fragment_types as ft
 
 
+if TYPE_CHECKING:
+    # We only use Literal for type-checking and Mypy always brings its own
+    # typing_extensions so this is safe without further dependencies.
+    if sys.version_info < (3, 8):
+        from typing_extensions import Literal
+    else:
+        from typing import Literal
+
 if sys.version_info < (3, 11):
     import tomli as tomllib
 else:
     import tomllib
+
+
+@dataclass
+class Config:
+    package: str
+    package_dir: str
+    single_file: bool
+    filename: str
+    directory: str | None
+    version: str | None
+    name: str | None
+    sections: Mapping[str, str]
+    types: Mapping[str, Mapping[str, Any]]
+    template: str
+    start_string: str
+    title_format: str | Literal[False]
+    issue_format: str | None
+    underlines: list[str]
+    wrap: bool
+    all_bullets: bool
+    orphan_prefix: str
 
 
 class ConfigError(Exception):
@@ -34,7 +64,7 @@ _underlines = ["=", "-", "~"]
 
 def load_config_from_options(
     directory: str | None, config_path: str | None
-) -> tuple[str, Mapping[str, Any]]:
+) -> tuple[str, Config]:
     if config_path is None:
         if directory is None:
             directory = os.getcwd()
@@ -43,10 +73,10 @@ def load_config_from_options(
         config = load_config(base_directory)
     else:
         config_path = os.path.abspath(config_path)
-        if directory:
-            base_directory = os.path.abspath(directory)
-        else:
+        if directory is None:
             base_directory = os.path.dirname(config_path)
+        else:
+            base_directory = os.path.abspath(directory)
         config = load_config_from_file(os.path.dirname(config_path), config_path)
 
     if config is None:
@@ -55,7 +85,7 @@ def load_config_from_options(
     return base_directory, config
 
 
-def load_config(directory: str) -> Mapping[str, Any] | None:
+def load_config(directory: str) -> Config | None:
 
     towncrier_toml = os.path.join(directory, "towncrier.toml")
     pyproject_toml = os.path.join(directory, "pyproject.toml")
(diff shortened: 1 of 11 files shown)


src/towncrier/_settings/load.py
```diff
@@ -43,10 +73,10 @@ def load_config_from_options(
         config = load_config(base_directory)
     else:
         config_path = os.path.abspath(config_path)
-        if directory:
-            base_directory = os.path.abspath(directory)
-        else:
+        if directory is None:
             base_directory = os.path.dirname(config_path)
+        else:
+            base_directory = os.path.abspath(directory)
```
hynek, line 79: I've flipped this logic to mirror the logic from above, because it was confusing to have the same check inverted.

hynek:
I had to run to the gym but I’ve got a half-finished test for the missing check branch ready. I’ll finish it when I’m back.

hynek:
OK I'm declaring victory here, please review away!

src/towncrier/_settings/load.py
```diff
@@ -7,19 +7,49 @@
...
+from typing import TYPE_CHECKING, Any, Mapping
 
 import pkg_resources
 
 from .._settings import fragment_types as ft
 
 
+if TYPE_CHECKING:
+    # We only use Literal for type-checking and Mypy always brings its own
+    # typing_extensions so this is safe without further dependencies.
```
exarkun, line 20: I don't understand this comment.  What is always safe?  And what does it mean to put a runtime check like `sys.version_info < (3, 8)` inside an `if TYPE_CHECKING:` guard (which guarantees it will never run).  Does mypy know what `sys.version_info < (3, 8)` means?

src/towncrier/_settings/load.py
```diff
@@ -134,22 +164,22 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Mapping[str, Any]:
...
+        types=types,
+        template=template,
+        start_string=config.get("start_string", _start_string),
+        title_format=config.get("title_format", _title_format),
+        issue_format=config.get("issue_format"),
+        underlines=config.get("underlines", _underlines),
+        wrap=wrap,
+        all_bullets=all_bullets,
+        orphan_prefix=config.get("orphan_prefix", "+"),
+    )
```
exarkun, line 185: This is excellent.

src/towncrier/build.py
```diff
@@ -114,51 +114,47 @@ def __main(
...
+    with open(config.template, "rb") as tmpl:
         template = tmpl.read().decode("utf8")
 
     click.echo("Finding news fragments...", err=to_err)
 
-    definitions = config["types"]
-
-    if config.get("directory"):
-        fragment_base_directory = os.path.abspath(config["directory"])
+    if config.directory:
```
exarkun: I wonder if the behavior of this conditional is intentional or if the following would be better:

```suggestion
    if config.directory is not None:
```

src/towncrier/test/helpers.py
```diff
@@ -0,0 +1,50 @@
+from functools import wraps
+from pathlib import Path
+
+from click.testing import CliRunner
+
+
+def read(filename):
```
exarkun: with type annotations this probably doesn't even need a docstring

src/towncrier/test/test_build.py
```diff
@@ -98,17 +86,72 @@ def test_command(self):
...
-            result = runner.invoke(_main, ["--draft", "--date", "01-01-2001"])
+        # Create a temporary directory, run Towncrier from there and assert
+        # it didn't litter into it.
+        with tempfile.TemporaryDirectory() as td:
+            os.chdir(td)
+            result = runner.invoke(cli, ("--yes", "--dir", str(project_dir)))
+
+            self.assertEqual([], list(Path(td).glob("*")))
+            # Staying in the temporary directory break cleanup on Windows.
+            os.chdir(project_dir)
```
exarkun: This probably needs to be in a finally or done with `addCleanup` or a failure from that `assertEqual` just above will cause problems (otoh why doesn't trial do this for you)

src/towncrier/_settings/load.py
hynek: It's safe to import it from typing_extensions, because when there's a type-checker, there should be typing_extensions.

Yes it does, unlike ImportErrors: https://hynek.me/til/typing-protocol/

I could probably always import from typing_extensions, but I like to have it explicit like this, so we can kill those branches wholesale once we drop 3.7.

src/towncrier/build.py
hynek: I wondered the same thing, but I tried to stay away from functional changes as well as I could. If our types are correct, it should indeed work your way!

src/towncrier/test/helpers.py
hynek: Yeah I was torn. Adding a docstring would make it more difficult to understand so I've decided against the hobgoblin. :)

src/towncrier/test/test_build.py
hynek: hmmm what you're saying is to use the non-contextmanager version…didn't think of that, I'll change it. Saves an indentation level too!

hynek: ah much better – thanks!

src/towncrier/test/helpers.py
hynek: I have added type hints to helpers, to "document" them. I've YOLO'ed the decorator typing to avoid depending on typing_extentions for now.

hynek:
I believe I've addressed everything?

SmileyChris (approved):
Looks great to me! Sorry for the merge hassle because of my orphan news fragment work 😅

hynek:
JFTR, despite the confusing message, I believe this needs an approving review from @twisted/twisted-contributors to merge.

adiroiban (approved):
Sorry. I will not have time to look into this this week.

I think that this can be merged.

In the future, I think is best to explicitly re-request a review from @twisted/twisted-contributors  so that a notification is raise.

JP did an initial review, but anyone else can pick up the PR and continue to review.

Many thanks.

For towncrier, I think we can be more flexible and merge with ease :)

adiroiban:
ok... I wasn't aware of the auto-merge. but I think it's ok :)

Thanks again, Happy to see a big coverage improvement

hynek:
Thanks Adi! I won't rest until it's 100%. :)

PR #430, opened 2022-09-23
Switch from tox to nox

hynek:
I know this is gonna be controversial, but I want to offer it to you anyhow. If this gets declined, I'll just keep using it locally.

# What

[nox](https://nox.thea.codes/en/stable/) is similar to tox, but instead of a rather unidiomatic substitution syntax within an ini file it uses plain Python. I hope you'll agree that the noxfile.py is a lot more readable than the tox.ini. It is my impression that a noxfile.py can be fully understood by anyone who has never used nox.

This is not a push the latest cool bleeding edge; it's used among _many_ others by pip.

# Why

The current tox file is pretty bad; I don't use it myself. The handling of coverage etc works for CI, but is confusing, useless (coverage combine after each step), and slow locally (**switching to nox gave me a ~15% speedup**). In other words, I'm proposing this out of own frustration, so I don't have to work around it. The current tox file is a sore thumb and unless someone else fixes it, it will remain so. That's a bad first impression for new contributors.

I'm under the impression that others aren't super happy with the current state either, since we also do this idiosyncratic tox_build.sh thing, I think it's better to use a richer tool.

# Additional Notes

I have renamed setup.cfg into .flake8 because that's what it is, and I don't expect anything ever getting added to setup.cfg, since we live in the golden days of pyproject.toml.

ci.yml can probably be simplified some further, but I tried to preserve the character and the diff small.

I have removed the ref install of incremental because I admittedly couldn't make work (even with explicit pip), but also because we want to get rid of it and it's just not worth those two casts.

`options={"bdist_wheel": {"universal": "1"}}` only makes sense if you support Python 2 which we don't.


adiroiban:
I also don't use tox... so I am ok with nox.

Happy to experiment with somenthing new.

For my projects I am using [paver](https://pythonhosted.org/Paver/) ... selected 10 years ago.

Looks like `nox` is focused on testing.

I am using `paver` and `tox` for general standard project scripting, not only running tests.

But I guess that for small projects, standardizing the test execution is all that it's needed.

adiroiban:
Thanks for the PR and showing a lot of love for this project.

As long as the tests are green and the docs are updated, I think we can go with it.

Locally, I don't use tox since it's very slow...so I just create separate virtualenvs and manually run the tests via trial... but is not ideal.

I am happy to experiment with nox.

hynek:
> Thanks for the PR and showing a lot of love for this project.

Thank you right back Adi! Towncrier would be far from what it is without your continued efforts.


> Looks like `nox` is focused on testing.

I guess that's the official focus, but being able to use regular Python and run commands, it's a super-powered-yet-portable scripting for me.

Check out the stuff [I do in doc2dash](https://github.com/hynek/doc2dash/blob/41bc93fe5d72f221f5eb5536fd569a422179ebca/noxfile.py#L164-L208) where I started using it.

To stress once more: I'm no graciously moving all my projects from tox to nox – I just feel like in our use cases it's cleaner and better solution.

altendky:
I haven't worked with nox, so no real opinion there.  Since you mentioned https://github.com/twisted/towncrier/blob/e301d673ddfa73046a111bca3b3800ba30b9829e/tox_build.sh though, I'll point out that at least the extract/build/cd/...  whatever mess is no longer needed since this was accepted as default behavior by `build` when you do not specify either `--wheel` nor `--sdist`.  It's an odd UX in my mind, but none the less lets you do the job.  So calls to `build` could be simplified and still retain the effective testing of 'can i build the wheel from the sdist'.

hynek:
> I haven't worked with nox, so no real opinion there. Since you mentioned [`e301d67`/tox_build.sh](https://github.com/twisted/towncrier/blob/e301d673ddfa73046a111bca3b3800ba30b9829e/tox_build.sh) though, I'll point out that at least the extract/build/cd/... whatever mess is no longer needed since this was accepted as default behavior by `build` when you do not specify either `--wheel` nor `--sdist`. It's an odd UX in my mind, but none the less lets you do the job. So calls to `build` could be simplified and still retain the effective testing of 'can i build the wheel from the sdist'.

Yeah this is true. If that were the only thing, I would've probably just replaced it.

I mean I leave it to y'all to look at the changes and decide whether it's an improvement or not.

src/towncrier/_project.py
```diff
@@ -51,7 +52,7 @@ def get_version(package_dir: str, package: str) -> str:
         return version.strip()
 
     if isinstance(version, Version):
-        return version.base().strip()
+        return cast(str, version.base().strip())
```
adiroiban, line 55: I guess that this is more related to https://github.com/twisted/towncrier/pull/431 

but we can have it here, no problem.

src/towncrier/newsfragments/430.misc
```diff
@@ -0,0 +1,2 @@
+We now use `Nox <https://nox.thea.codes/>`_ instead of Tox.
+You may have to adjust a few command lines, but overall contributing should now be more comfortable.
```
adiroiban, line 2: .. more comfortable.

arguably :) 

I guess that for people who like tox and already have tox, this is not more comfortable.

Unless @altendky  is not -1 for this, I think this can be merged.

Since there is a report of speed improvements, I am +1 for this.

nox, tox, potayto, potahto ... still not paver to me :)

adiroiban (approved):
I don't see anything wrong with this PR... but for a long time, I in the team opposing tox  ... so not much love for tox from me.

@altendky is anything wrong with this PR that you should block the merge?

Thanks you guys!

adiroiban:
@hynek I think this can be merged.

I am not sure if the casts for project.py are needed...but that is not a blocker.

thanks

hynek:
Thanks Adi! They're needed because I've stopped using a Git version of incremental.

Our mypy config _should_ complain about redundant casts.

PR #434, opened 2022-09-28
Capture & assert over master branch warning

hynek:
This is an addendum to #400 that prevents that the warning is shown while running tests and also makes sure it's actually raised.

### DIFF
--- a/src/towncrier/test/test_check.py
+++ b/src/towncrier/test/test_check.py
@@ -4,6 +4,7 @@
 import os
 import os.path
 import sys
+import warnings
 
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


hynek:
JFTR I know about assertWarns, but it's useless for this case. It should really be expanded to not require all values. The warning in this case is raised in Twisted's _synctest with weird paths.

adiroiban (approved):
Thanks.

For `trial`, I think it would be useful to automatically catch the warnings emitted for each test and have a way to fail on "un-asserted" warnings.

I think pytest has something like this.

Issue #128, opened 2018-09-06
Markdown output support?


bruxisma:
Hello,

I was wondering if a bit more support for Markdown might be possible? I've more or less successfully ~tricked~ *convinced* towncrier to output Markdown for my changelog file, however there's no way to change or enforce the H1 title output in this case. Right now to sort of get around it, I've done the following in my pyproject.toml:

```toml
[tool.towncrier]
package = 'dripstone'
package_dir = 'src'
filename = 'CHANGELOG.md'
directory = ".changelog"
template = ".changelog/template.md"
title_format = "# {name} {version} ({project_date})"
issue_format = "#{issue}"
start_string = "<!-- TOWNCRIER -->\n"
underlines = ["<!--- --->", "##", "###"]
```

This (plus a slight change to the default template to make it more markdown friendly) then gets generated into

```markdown
# Dripstone 0.1.0 (2018-09-06)
<!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- ---><!--- --->

<!-- TOWNCRIER -->


## Bugfixes

* Fixed Whatever (#1234)
* fixed some more (#12345)
```

Which is uhhh, *a terrible hack* and makes me feel uncomfortable. I can try to submit a patch if you'd like, but I'm just curious if this is relevant to the project's goals.

slurps-mad-rips:
I suppose! It'd be nice if I didn't have to set all the fields to get it to work with markdown, but it's not a big deal 🤷‍♀️

deeplook:
What's the status of Markdown support now, two years later? ;)

timwsuqld:
@slurps-mad-rips Are you able to share your `template.md` changes as well? Thanks

slurps-mad-rips:
@timwsuqld it's located [here](https://github.com/slurps-mad-rips/dripstone/blob/master/.changelog/template.md) and I haven't touched this project in a long time, and will most likely delete it at some point.

timwsuqld:
@slurps-mad-rips Thanks for that. I eventually hacked together my own, and looking at yours we came to roughly the same place. 

[...]

PR #435, opened 2022-09-28
De-duplicate & expand docs

hynek:
Sorry, this is another big one 😇

I think our docs are our weakest point and the sole reason, why towncrier isn't more widely used.

There's tons of duplication, missing information and flat-out lies.

I would love to write a how-to for #128 but this needs some more big-picture work.

Here's some more notes:

- I have replaced the badges with something that is actually useful to the users. The codecov badge even was lying about 98%. We don't need a CI status badge anymore, there's a green checkmark in the GitHub UI nowadays.
- I'm sorry for the Queen, but I've killed a bunch of British English, because the rest is in American English too.
- I've switched the docs to Furo, because it's not 2010 anymore.
- I've renamed the quickstart to tutorial, to be more in line with common terms and with https://diataxis.fr which is fantastic.
- I've reordered the reference a bit, because configuration comes before customization.
- I've also split it into Narrative, Reference and Development in the hopes that someone (possibly me) fills more of the narrative. :)
- The configuration docs need **a lot more polish**, but I had to draw a line somewhere.
- I've added CLI reference; we should expand it.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,6 @@
 * [ ] Create a file in `src/towncrier/newsfragments/`. Describe your
   change and include important information. Your change will be included in the public release notes.
 * [ ] Make sure all GitHub Actions checks are green (they are automatically checking all of the above).
+* [ ] Ensure `docs/tutorial.rst` is still up-to-date.
+* [ ] If you add new **CLI arguments** (or change the meaning of existing ones), make sure `docs/cli.rst` reflects those changes.
```
adiroiban, line 12: I think this ok for now.

In the future maybe we can use something like https://pypi.org/project/sphinx-click/ to automatically keep the reference up to date.

adiroiban (approved):
Docs at https://towncrier--435.org.readthedocs.build/en/435/ are awesome

I will not have time soon to look into the details of this, but I think is best to have this merged.

If anything needs improvements we can create smaller PRs for that change and discuss it.

----------

I agree that the current docs are a mess, so this looks so much better.

Thanks for https://diataxis.fr/ 
I wasn't aware of it.

Again
Awesome work!

.github/PULL_REQUEST_TEMPLATE.md
hynek: Yes and no…I think the reference docs should do more than parroting `--help` output. Maybe some kind of combination idk.

hynek:
Thanks for the quick review Adi! Hope this lays the ground for better docs in the future!

.github/PULL_REQUEST_TEMPLATE.md
adiroiban: I agree that in the docs we can do more than copy/pasting `--help`

But we should make sure the `--help` output is useful and can serve as documentation.

Having all the `--help` output in the docs can serve as a base-line.

But if there is anyting important, I would prefer to have it inside the `--help` output and not keep it only for the HTML docs :)

I use the `--help` command more often than checking the HTML / narrative docs for a project.

PR #436, opened 2022-09-28
Add a Markdown how-to

### LINKED ISSUE #128: Markdown output support?

hynek:
I believe this fixes #128.

While working on this I've noticed a big problem: the order of the categories is always alphabetical!? This needs to be fixed.

### DIFF
--- a/docs/keep-a-changelog.rst
+++ b/docs/keep-a-changelog.rst
@@ -1,0 +1,162 @@
+How to Keep a Changelog in Markdown
+===================================
+
+`Keep a Changelog <https://keepachangelog.com/>`_ is a standardized way to format a news file in `Markdown <https://en.wikipedia.org/wiki/Markdown>`_.
+
+This guide shows you how to configure ``towncrier`` for keeping a Markdown-based news file of a project without using any Python-specific features.
+Everything used here can be use with any other language or platform.
+
+This guide makes the following assumptions:
+
+- The project lives at https://github.com/twisted/my-project/.
+- The news file name is ``CHANGELOG.md``.
+- You store the news fragments in the ``changelog.d`` directory at the root of the project.
+
+Put the following into your ``pyproject.toml`` or ``towncrier.toml``:
+
+.. code-block:: toml
+
+   [tool.towncrier]
+   directory = "changelog.d"
+   filename = "CHANGELOG.md"
+   start_string = "<!-- towncrier release notes start -->\n"
+   underlines = ["", "", ""]
+   template = "changelog.d/changelog_template.jinja"
+   title_format = "## [{version}](https://github.com/twisted/my-project/tree/{version}) - {project_date}"
+   issue_format = "[#{issue}](https://github.com/twisted/my-project/issues/{issue})"
+
+   [tool.towncrier.fragment.security]
+   name = "Security"
+
+   [tool.towncrier.fragment.removed]
+   name = "Removed"
+
+   [tool.towncrier.fragment.deprecated]
+   name = "Deprecated"
+
+   [tool.towncrier.fragment.added]
+   name = "Added"
+
+   [tool.towncrier.fragment.changed]
+   name = "Changed"
+
+   [tool.towncrier.fragment.fixed]
+   name = "Fixed"
+
+
+Next create the news fragment directory and the news file template:
+
+.. code-block:: console
+
+   $ mkdir changelog.d
+
+And put the following into ``changelog.d/changelog_template.jinja``:
+
+.. code-block:: jinja
+
+   {% if sections[""] %}
+   {% for category, val in definitions.items() if category in sections[""] %}
+
+   ### {{ definitions[category]['name'] }}
+
+   {% for text, values in sections[""][category].items() %}
+   - {{ text }} {{ values|join(', ') }}
+   {% endfor %}
+
+   {% endfor %}
+   {% else %}
+   No significant changes.
+
+
+   {% endif %}
+
+
+Next, create the news file with an explanatory header::
+
+   $ cat >CHANGELOG.md <<EOF
+   # Changelog
+
+   All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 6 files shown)


docs/keep-a-changelog.rst
```diff
@@ -0,0 +1,162 @@
...
+Put the following into your ``pyproject.toml`` or ``towncrier.toml``:
+
+.. code-block:: toml
+
+   [tool.towncrier]
+   directory = "changelog.d"
+   filename = "CHANGELOG.md"
+   start_string = "<!-- towncrier release notes start -->\n"
+   underlines = ["", "", ""]
+   template = "changelog.d/changelog_template.jinja"
```
adiroiban: For now is ok.
I expect that Markdown is a common use case, so we might want to have a predefined markdown template, that is maintained by the main dev team.

docs/keep-a-changelog.rst
```diff
@@ -0,0 +1,162 @@
...
+   title_format = "## [{version}](https://github.com/twisted/my-project/tree/{version}) - {project_date}"
+   issue_format = "[#{issue}](https://github.com/twisted/my-project/issues/{issue})"
+
+   [tool.towncrier.fragment.security]
+   name = "Security"
```
adiroiban: Previously the config was something like this

```
    [[tool.towncrier.type]]
        directory = "security"
        name = "Security Fixes"
        showcontent = true

    [[tool.towncrier.type]]
        directory = "feature"
        name = "New Features"
        showcontent = true
```

I guess the main issue is due to pytoml not using an ordered dict.

adiroiban (approved):
Thanks. Looks good.

I think we should merge this.

I am not using markdown, but I expect that this is a very common use case.
This is why, in the future we might want to incorporate the template in towcrier.

Maybe have some "predefined" templates, that when connfigured will automatically also set/overwrite the `underlines`, `title_format` and `issue_format`

So that you can have just

```
   [tool.towncrier]
   directory = "changelog.d"
   filename = "CHANGELOG.md"
   start_string = "<!-- towncrier release notes start -->\n"
   template = "markdown"
```

When towncrier will see a config with `markdown` as the value, will not search for that file, but instead use some predefined values, with a template distributed with towncrier.

docs/keep-a-changelog.rst
hynek: Ooof yeah you're right. It used to be an array but is a dict now and tomli gives it alphabetically:

```python
{'tool': {'towncrier': {'directory': 'changelog.d',
                        'filename': 'CHANGELOG.md',
                        'fragment': {'added': {'name': 'Added'},
                                     'changed': {'name': 'Changed'},
                                     'deprecated': {'name': 'Deprecated'},
                                     'fixed': {'name': 'Fixed'},
                                     'removed': {'name': 'Removed'},
                                     'security': {'name': 'Security'}},
                        'issue_format': '[#{issue}](https://github.com/twisted/my-project/issues/{issue})',
                        'start_string': '<!-- towncrier release notes start '
                                        '-->\n',
                        'template': 'changelog.d/changelog_template.jinja',
                        'title_format': '## '
                                        '[{version}](https://github.com/twisted/my-project/tree/{version}) '
                                        '- {project_date}',
                        'underlines': ['', '', '']}}}
```

adiroiban: For reference.
This was implemented here https://github.com/twisted/towncrier/pull/370/files

I think the old format is still supported.

----------

This looks like a bug for the new format.

adiroiban: I am not sure what toml lib we a re using, but it looks like ordered dict are not a thing :)

https://github.com/uiri/toml/pull/2

and also not wanted in the general TOML spec

https://github.com/toml-lang/toml/issues/162

hynek: In modern Pythons, dicts are _always_ ordered. This is a deliberate action by tomli based on the spec that you linked. I've opened #437 to track this.

hynek:
Please don't merge until #438 is merged and I have adapted this to the array style – thanks!

hynek:
I just realized this one is not dependent on docs being updated so here it is. I've re-done it using the array-style definition.

adiroiban (approved):
Thanks. All good.

`directory` instead of `file_extension` is ugly ... but this life :)

SmileyChris:
I was working on better markdown integration a while back, just resurrected that stash and made a PR for it: #439

Might be a much easier way to handle markdown formatting...

PR #438, opened 2022-09-28
Un-deprecate array-style custom fragments

### LINKED ISSUE #437: Need a way to set the order of fragments
The new way of defining custom fragments uses a dictionary that is unfortunately sorted alphabetically by tomli.

For example this is from my Markdown tutorial in #436:

```python
{'tool': {'towncrier': {'directory': 'changelog.d',
                        'filename': 'CHANGELOG.md',
                        'fragment': {'added': {'name': 'Added'},
                                     'changed': {'name': 'Changed'},
                                     'deprecated': {'name': 'Deprecated'},
                                     'fixed': {'name': 'Fixed'},
                                     'removed': {'name': 'Removed'},
                                     'security': {'name': 'Security'}},
                        'issue_format': '[#{issue}](https://github.com/twisted/my-project/issues/{issue})',
                        'start_string': '<!-- towncrier release notes start '
                                        '-->\n',
                        'template': 'changelog.d/changelog_template.jinja',
                        'title_format': '## '
                                        '[{version}](https://github.com/twisted/my-project/tree/{version}) '
                                        '- {project_date}',
                        'underlines': ['', '', '']}}}
```

The fragment types are defined as:

```
[tool.towncrier.fragment.security]
name = "Security"

[tool.towncrier.fragment.removed]
name = "Removed"

[tool.towncrier.fragment.deprecated]
name = "Deprecated"

[tool.towncrier.fragment.added]
name = "Added"

[tool.towncrier.fragment.changed]
name = "Changed"

[tool.towncrier.fragment.fixed]
name = "Fixed"
```

In hindsight, it's unfortunate we've switched away from using a list for that but it is what it is.

I can think of two way to solve this:

1. add an `order` key to the fragments:

```toml
[tool.towncrier.fragment.security]
name = "Security"
order = 1

[tool.towncrier.fragment.removed]
name = "Removed"
order = 2
```

2. add an option:

```toml
[tool.towncrier]
category_order = ["security", "removed", "..."]
```

I think I prefer option 2, because you don't have to re-number all entries if you change the order.

The question is what would happen if a catefory is missing – I would suggest an error.

hynek:
Adding order to the dicts is just extra work – let's use an ordered structure if order matters.

I would like to change `directory`, but that's a different PR.

Fixes #437

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -36,6 +36,12 @@ Custom fragment types
 ``towncrier`` allows defining custom fragment types.
 Custom fragment types will be used instead ``towncrier`` default ones, they are not combined.
 
+There are two ways to add custom fragment types.
+
+
+Defining Custom Fragment Types With a TOML Mapping
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 Users can configure each of their own custom fragment types by adding tables to
 the pyproject.toml named ``[tool.towncrier.fragment.<a custom fragment type>]``.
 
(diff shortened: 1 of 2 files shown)


docs/configuration.rst
```diff
@@ -59,6 +65,44 @@ For example, if you want your custom fragment types to be ``["feat", "fix", "cho
    showcontent = false
 
 
+.. warning::
+
+   Since TOML mappings aren't ordered, the sections are always rendered alphabetically.
```
adiroiban, line 70: I guess you can have an "order_index" and sort the dict in jinja or python.

But this requires extra configuration work and extra template or python work.

adiroiban (approved):
Looks good.

Thanks.

--------

I am using the same towcnrier from 2017 that I have patched to meet my requirements.

I think that by now latest towcrier can meet my requirement ...but I will not touch my towncrier system.
As long as it works, I will not fix it.

---------

I am not sure how this was not detected during the automated tests.
Or when the new format was implement maybe we didn't care about the order.

docs/configuration.rst
hynek: Yeah it just doesn't make sense to force an ordering on an unordered data structure.

We should focus on making the ordered version nicer (better key names, make `showcontent` optional).

PR #446, opened 2022-11-01
Remove unnecessary casts

hynek:
Since incremental has been released, we have type hints for it now.


adiroiban (approved):
Thanks for the cleanup!

PR #447, opened 2022-11-04
Generate coverage reports using only GitHub Actions, without codecov.io

### LINKED ISSUE #410: Code coverage via GitHub action

adiroiban:
# Description

Fixes #410

This tries to generate coverage reports only via GitHub Actions...so that we don't have to depend on codecov.io

# Drive-by

I have removed the codeowner as I don't think it helps.

### DIFF
--- a/.github/scripts/pr_comment.js
+++ b/.github/scripts/pr_comment.js
@@ -33,7 +33,7 @@ module.exports = async ({github, context, process}) => {
             process.env.GITHUB_WORKSPACE + "/" + process.env.COMMENT_BODY, 'utf8');
         var comment_id = null
         var comment_marker = '\n' + process.env.COMMENT_MARKER
-        var comment_body =  body + comment_marker
+        var comment_body = body + comment_marker
 
         var comments = await octokit_rest.issues.listComments({
             owner: context.repo.owner,
--- a/.github/scripts/pr_comment.js
+++ b/.github/scripts/pr_comment.js
@@ -24,12 +24,23 @@ module.exports = async ({octokit_rest, context, process}) => {
     */
     var doAction = async function() {
 
+        fs = require('fs');
+        var body = fs.readFile(
+            process.env.GITHUB_WORKSPACE + "/" + process.env.COMMENT_BODY)
+
+        var comments = await octokit.rest.issues.listComments({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            issue_number: context.event.number,
+          })
+        console.log(comments)
+
         await octokit_rest.issues.createComment({
             owner: context.repo.owner,
             repo: context.repo.repo,
             issue_number: context.event.number,
-            body: process.env.COMMENT_BODY,
-          });
+            body: body,
+          })
 
     }
 
--- a/.github/scripts/pr_comment.js
+++ b/.github/scripts/pr_comment.js
@@ -7,7 +7,8 @@ Update the content of the existing comment.
 
 https://octokit.github.io/rest.js/v19
 */
-module.exports = async ({octokit_rest, context, process}) => {
+module.exports = async ({github, context, process}) => {
+    var octokit_rest = github
     if (context.eventName != "pull_request") {
         // Only PR are supported.
         return
@@ -24,23 +25,49 @@ module.exports = async ({octokit_rest, context, process}) => {
     */
     var doAction = async function() {
 
-        fs = require('fs');
-        var body = fs.readFile(
-            process.env.GITHUB_WORKSPACE + "/" + process.env.COMMENT_BODY)
+        console.log(context)
 
-        var comments = await octokit.rest.issues.listComments({
+        fs = require('fs');
+
+        const body = fs.readFileSync(
+            process.env.GITHUB_WORKSPACE + "/" + process.env.COMMENT_BODY, 'utf8');
+        var comment_id = null
+        var comment_marker = '\n' + process.env.COMMENT_MARKER
+        var comment_body = body + comment_marker
+
+        var comments = await octokit_rest.issues.listComments({
             owner: context.repo.owner,
             repo: context.repo.repo,
-            issue_number: context.event.number,
+            issue_number: context.payload.number,
           })
+
         console.log(comments)
 
+        comments.data.forEach(comment => {
+            if (comment.body.endsWith(comment_marker)) {
+                comment_id = comment.id
+            }
+        })
+
+        if (comment_id) {
+            // We have an existing comment.
+            // update the content.
+            await octokit_rest.issues.updateComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                comment_id: comment_id,
+                body: comment_body,
+            })
+            return
+        }
+
+        // Create a new comment.
         await octokit_rest.issues.createComment({
             owner: context.repo.owner,
             repo: context.repo.repo,
-            issue_number: context.event.number,
-            body: body,
-          })
+            issue_number: context.payload.number,
+            body: comment_body,
+        })
 
     }
 
--- a/.github/scripts/pr_comment.js
+++ b/.github/scripts/pr_comment.js
@@ -1,0 +1,42 @@
+/*
+Have a single comment on a PR, identified by a comment marker.
+
+Create a new comment if no comment already exists.
+Update the content of the existing comment.
+
+
+https://octokit.github.io/rest.js/v19
+*/
+module.exports = async ({octokit_rest, context, process}) => {
+    if (context.eventName != "pull_request") {
+        // Only PR are supported.
+        return
+    }
+
+    var sleep = function(second) {
+        return new Promise(resolve => setTimeout(resolve, second * 1000))
+    }
+
+    /*
+    Perform the actual logic.
+
+    This is wrapped so that we can retry on errors.
+    */
+    var doAction = async function() {
+
+        await octokit_rest.issues.createComment({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            issue_number: context.event.number,
+            body: process.env.COMMENT_BODY,
+          });
+
+    }
+
+    try {
+        await doAction()
+    } catch (e) {
+        await sleep(5)
+        await doAction()
+    }
+}
--- a/.github/scripts/pr_comment.js
+++ b/.github/scripts/pr_comment.js
@@ -33,7 +33,7 @@ module.exports = async ({github, context, process}) => {
             process.env.GITHUB_WORKSPACE + "/" + process.env.COMMENT_BODY, 'utf8');
         var comment_id = null
         var comment_marker = '\n' + process.env.COMMENT_MARKER
-        var comment_body = body + comment_marker
+        var comment_body =  body + comment_marker
 
         var comments = await octokit_rest.issues.listComments({
             owner: context.repo.owner,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,8 +96,20 @@ jobs:
 
     - run: nox --python ${{ matrix.python.action }} -e ${{ matrix.task.nox }} -- --use-wheel dist/*.whl
 
+    - name: Store coverage file
+      uses: actions/upload-artifact@v3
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -232,6 +244,37 @@ jobs:
...
+      uses: actions/download-artifact@v3
+      with:
+        name: coverage-${{ github.run_id }}
+        path: .
+
+    - name: Combine coverage
+      run: coverage combine
+
+    - name: Report coverage
+      run: coverage report
```
altendky: Maybe also create an html report and artifact it for download?

.github/workflows/ci.yml
```diff
@@ -232,6 +244,37 @@ jobs:
...
 
+  coverage-report:
+    name: Coverage report
+    runs-on: ubuntu-latest
+    needs:
+      # We are waiting only for test jobs.
+      - test-linux
+      - test-windows
+    steps:
+    - uses: actions/checkout@v3
```
altendky, line 270:
```suggestion
    - uses: actions/checkout@v3
      with:
        fetch-depth: 0
        
```

.github/workflows/ci.yml
adiroiban: My plan is to send the diff coverage report as a comment... downloading the artifact looks complicated.

adiroiban:
I think this is ready for an initial feedback

needs-review

.github/workflows/ci.yml
altendky: Yeah, it's wouldn't be a primary use.  But, it is how you can you get a nice view of the code and coverage without having somewhere to upload the html.  Anyways, not critical, just a thought that can also be easily added later.

.github/workflows/ci.yml
```diff
@@ -232,6 +258,54 @@ jobs:
...
+        COMMENT_BODY: coverage-report.txt
+      with:
+        script: |
+          const script = require(`${process.env.GITHUB_WORKSPACE}/.github/scripts/pr_comment.js`)
+          // Only pass top level objects as GHA does dependecy injection.
+          await script({github, context, process})
+
+    - name: Diff coverage
+      run: |
+        git fetch
```
altendky: This is still needed after `fetch-depth: 0` above?

.github/workflows/ci.yml
```diff
@@ -232,6 +258,54 @@ jobs:
...
+      # We are waiting only for test jobs.
+      - test-linux
+      - test-windows
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Install dependencies
+      run: python -m pip install --upgrade pip coverage[toml] diff_cover
```
altendky: Not much point in updating pip if you don't do that first and then make a second pip call to install the other stuff.  Both options are available below as suggestions.

(I just realized I could add multiple comments in the same discussion at the same time so that this can work...)

.github/workflows/ci.yml
```diff
@@ -137,8 +150,21 @@ jobs:
 
     - run: nox --python ${{ matrix.python.action }} -e ${{ matrix.task.nox }} -- --use-wheel dist/*.whl
 
+    - name: Store coverage file
+      uses: actions/upload-artifact@v3
+      with:
+        # We have a single artifact shared for each workflow run.
+        # in this way we should intefere with coverage reports from other PRs
+        # or previous runs for the same PR.
```
altendky: Slight wording tweak, though I'm not sure there's really a need to have the run ID in the name.  Yes, the files need to be unique but AFAIK uploading like this uses names within a namespace for this run.  And actually, if you get a new run ID for reruns then I think this setup wouldn't work in case a job flakes and you rerun it as you would then get two separate artifacts.

```suggestion
        # We have a single artifact shared for each workflow run.
        # in this way we should not interfere with coverage reports from other PRs
        # or previous runs for the same PR.
```

.github/workflows/ci.yml
```diff
@@ -137,8 +150,21 @@ jobs:
...
+        name: coverage-${{ github.run_id }}
+        # Is important to keep the unique names for the coverage files
+        # so that when uploaded to the same artifact, they don't overlap.
```
altendky:
```suggestion
        # It is important to keep the unique names for the coverage files
        # so that when uploaded to the same artifact, they don't overlap.
```

.github/workflows/ci.yml
```diff
@@ -96,8 +96,21 @@ jobs:
...
+        path: .coverage*
```
altendky: See comments in duplicated section below.

Side note...  I really like cross-platform matrices to avoid this.  `:]`

noxfile.py
```diff
@@ -36,9 +36,8 @@ def tests(session: nox.Session) -> None:
     session.run("coverage", "run", "--module", "twisted.trial", *posargs)
 
     if os.environ.get("CI") != "true":
+        # When running the test localy, show the coverage report.
```
altendky:
```suggestion
        # When running the test locally, show the coverage report.
```

pyproject.toml
```diff
@@ -91,7 +91,11 @@ branch = true
 source = ["towncrier"]
 
 [tool.coverage.paths]
-source = ["src", ".nox/*/site-packages"]
+source = [
+    "src",
+    "*.nox/*/site-packages",
+    "*.nox\\*\\*\\site-packages"
+]
```
altendky: I guess this config with slashes each direction is compensating for now for the Windows issues discussed in https://github.com/nedbat/coveragepy/issues/991.  I'd suggest just using the 6.6.0b1 that has a fix for that but...  https://github.com/pypa/pip/issues/11565  I'm not sure, towncrier may or may not trigger that bug.  So maybe best to just wait and leave a comment here as reference.

```suggestion
source = [
    "src",
    "*.nox/*/site-packages",
    # required until coverage 6.6.0 is used: https://github.com/nedbat/coveragepy/issues/991
    "*.nox\\*\\*\\site-packages"
]
```

.github/workflows/ci.yml
altendky:
```suggestion
      run: python -m pip install --upgrade coverage[toml] diff_cover
```

altendky:
```suggestion
      run: |
        python -m pip install --upgrade pip
        python -m pip install --upgrade coverage[toml] diff_cover
```

altendky (commented):
I know this isn't part of the PR, but it's on topic so I'll at least comment.

https://github.com/twisted/towncrier/blob/a86b23b1e206cdf5897d59f6c72a1df8d227ce8c/pyproject.toml#L107-L110

Why would we omit tests?  Not-covered test code is also bad.

Anyways, I ended up with a lot of comments.  `:|`  I hope they are helpful.

.github/workflows/ci.yml
adiroiban: I don't know. I think so. Let's see.

```suggestion
```

adiroiban:
thanks @altendky  for the review. 

> Why would we omit tests? Not-covered test code is also bad.

I don't know. I have enabled the test coverage. Looks 100%

I also think that test code coverage is super important... you might think that you have tests covering the prod code but in fact the tests are always skipped.

adiroiban:
> Maybe also create an html report and artifact it for download?

Searching for artifacts, downloading the artifacts, extracting the archive, and loading in a local browser... seems complicated.

For me is easier to just run the tests and check the local report.

---------

The idea with CI is just to raise a warning if coverage is missing.

.github/workflows/ci.yml
adiroiban: I am missing the comment.

I am happy to see less duplication, but I think is best to have it in separate PR.

Thanks

.github/scripts/pr_comment.js
```diff
@@ -0,0 +1,80 @@
...
+
+        console.log(context)
+
+        fs = require('fs');
+
+        const body = fs.readFileSync(
+            process.env.GITHUB_WORKSPACE + "/" + process.env.COMMENT_BODY, 'utf8');
+        var comment_id = null
+        var comment_marker = '\n' + process.env.COMMENT_MARKER
+        var comment_body =  body + comment_marker
```
altendky: Looks like a stray space got introduced in https://github.com/twisted/towncrier/pull/447/commits/8208990ecbdbd4454d39a3f461215c7aba748e62#diff-e24afedc0691665628968c647e1b47778e5678bcc154c1c7e83ab9b1d63680b2.

```suggestion
        var comment_body = body + comment_marker
```

adiroiban:
Thanks Kyle for the review.

I think that this is ready for another review.

For now, the output/feedback is done via a single comment that is updated and contains the output from `coverage report` and `diff-cover`

.github/workflows/ci.yml
altendky: This is copy pasted from the other similar section and contains the same typos that were present there and have been fixed.  They should be fixed here as well.

But, I'm going to go ahead and trigger a run here, cancel it partially completed, rerun the cancelled jobs and see if my concern about using the `github.run_id` seems accurate.

altendky (commented):
Thanks for the cleanup, a couple things left to check on.

adiroiban:
Re-run will keep the run-id. Is not ideal ... but I think it should work.

altendky:
Well, I was trying over at https://github.com/twisted/towncrier/actions/runs/3401839419 and yes, the re-runs/attempts do retain the same run ID as you say.  So, what is the need for the run ID here?  Just so that if you download the artifact locally you know what run it corresponds to?

altendky (approved):
Thanks!

adiroiban:
I added the run-id to make sure we don't mix coverages between commits or PRs

Based on the docs from here 

https://github.com/actions/upload-artifact#uploading-to-the-same-artifact

> Each artifact behaves as a file share. Uploading to the same artifact multiple times in the same workflow can overwrite and append already uploaded files:

But re-reading it I see "in the same workflow"

So I think that the run ID is not needed.

at first I thought that the artifacts are shared between multiple workflows and runs.

I will remove the run-id as it is not needed

adiroiban:
Many thanks, Kyle for the review and useful feedback.

I will not merge it just now as I would like to apply the same process to twisted/twisted and apply any changes made there to keep the code in sync.

here is how it looks for twisted/twisted https://github.com/twisted/twisted/pull/11745

We still need a real PR with a real diff coverage...but we can see if after this is merged and update the format once we have a real example

adiroiban:
I now see that diff-covers supports markdown output. I have enabled it.

diff-covers also supports json report so in theory we can parse that output and generate inline PR comments, similar to codecov.io

I will give it a try here.

----------

For me, the HTML artifact report doesn't look helpful.
I would like to have everything reported inside the PR via comments or inline diff comments/annotations.

So future work, if someting is missing I think is best to contribute to diff-cover so that the markdown or json reports will do what we want.


------------

For example, I see that `diff-cover` range detection can't merge ranges splited by a comment line.

```
coverage report
src/towncrier/build.py                  92      7     30      2    93%   123-126, 138-142
```

```
diff-cover
src/towncrier/build.py (22.2%): Missing lines 123,125-126,138,140-142
```

adiroiban:
I will merge this an open a new PR in which I will try to get inline comments working.

adiroiban:
A better example for the report is here https://github.com/twisted/towncrier/pull/451

hynek:
should we maybe roll this back for now, since it's breaking on all PRs that aren't by us?

adiroiban:
Yes. I think this should be reverted. @hynek  Will you do that?
Thanks!

PR #451, opened 2022-11-06
Update coverage report to add inline PR comments.

### LINKED ISSUE #450: Have inline PR comments for coverage
We should try to replace the codecov.io inline comments for missing coverage.

adiroiban:
# Description

Fixes #450

Try to add inline comments for PR coverage 

To make it easier to develop. I added some testing code for JS.
I guess that if all is ok, we can move it into a separate dedicated shared action.

In theory, we can convert this to Python code.


adiroiban:
I see that codecov.io is not using PR comments, but commit check warnings

image

I will continue working on this and try to use the `checks` API

adiroiban:
I tried to see how to use the GitHub Check Annotations, but it is not clear how to add annotations

https://docs.github.com/en/rest/checks/runs#list-check-run-annotations

At the same time, I think that we should not rely too much on GitHub Api.

Instead, we should dedicated our time to improve the `diff-cover` project so that the markdown report is helpful.

I am going to close this PR.

PR #453, opened 2022-11-24
Add --keep option to allow to generate newsfile, but keep newsfragmen…

### LINKED ISSUE #129: Add option `--no`
For batch-processing (e.g. when testing), an option `--no` would be helpful:
```
--no           Do not ask for confirmation to remove news fragments, assume "no".
```

fizyk:
…ts - closes #129

# Description
Resolves issue #129 It's not a --no option, but given the generality of the --no option, --keep is a bit more descriptive than proposed --no.

# Checklist


adiroiban:
Hi Grzegorz. Many thanks for the nice PR.

I don't really understand why this is needed.

There is the "draft" command which is expected to generate the release notes, without removing the fragments.

It is true that right now, the draft command generates the result in the standard output.

But maybe a better way to fix this is to add a file option to the existing "draft" command.

This is just a suggestion.

fizyk:
@adiroiban I do not have strong opinion on where. It seemed natural, although the --draft could also be the solution.

The underlying issue suggests it would be useful to generate a `current` changelog for development documentation, my issue is that having numerous fragment types I want to have a general changelog with all the changes and additionally specific categories being generated into its own, separate changelog file as well.

adiroiban:
Thanks for the explanation. I think '--keep' makes sense.

> The underlying issue suggests it would be useful to generate a current changelog for development documentation, 

I think that this is "draft" use case.
But "draft" has clear documentation that it sends the result to stdout... so... it's complicated.

I will check the PR now and I hope we can merge it.

Thanks again

src/towncrier/_git.py
```diff
@@ -10,18 +10,25 @@
...
+    elif answer_keep:
+        click.echo("Keeping the following files:")
     else:
         click.echo("I want to remove the following files:")
 
     for filename in fragment_filenames:
         click.echo(filename)
 
+    if answer_keep and not answer_yes:
+        return
```
adiroiban: Can we can "answer_yes == True" here ?

I was expecting that this is validated in the caller code.
By handling this branch, somebody might think that this is a valid use-case.

```suggestion
    if answer_keep:
        # Not proceeding with the removal of the files.
        return
```

src/towncrier/build.py
```diff
@@ -26,6 +28,18 @@ def _get_date() -> str:
     return date.today().isoformat()
 
 
+def validate_answer(ctx: Context, param: Option, value: bool) -> bool:
```
adiroiban: Maybe just make it a private method, not to encourage its usage outside of towncrier.

I thing that the missing docs is ok.
I can underestand the code even without docs.

```suggestion
def _validate_answer(ctx: Context, param: Option, value: bool) -> bool:
```

src/towncrier/build.py
```diff
@@ -67,9 +81,18 @@ def _get_date() -> str:
...
     flag_value=True,
     help="Do not ask for confirmation to remove news fragments.",
+    callback=validate_answer,
+)
+@click.option(
+    "--keep",
+    "answer_keep",
+    default=None,
+    flag_value=True,
+    help="Keep the newsfragments.",
```
adiroiban: Can we have more help text here?

In an ideal case, the whole info from the "cli.rst" file would be placed here.

In this way, we can use cli-man to generate a man version of the docs...and we can also use `towncrier build --help` to get that info.

these days, when I use a CLI, in order to get some documentation, I first try my luck with "--help" and only if it fails I use "man" or the project website.

src/towncrier/test/test_build.py
```diff
@@ -407,6 +407,56 @@ def test_no_confirmation(self):
             self.assertFalse(os.path.isfile(fragment_path1))
             self.assertFalse(os.path.isfile(fragment_path2))
 
+    def test_keep_fragments(self):
+        runner = CliRunner()
```
adiroiban: I think that is very important to have a docstring for the test to describe its scope and purpose

Later we might have many overlaping tests and it will be hard to know which tests is for which purpose.

```suggestion
    def test_keep_fragments(self):
        """
        The `--keep` option will build the full final news file
        without deleting the fragment files and without
        any extra CLI interaction or confirmation.
        """
        runner = CliRunner()
```

src/towncrier/test/test_build.py
```diff
@@ -407,6 +407,56 @@ def test_no_confirmation(self):
...
+            call(["git", "config", "user.name", "user"])
+            call(["git", "config", "user.email", "user@example.com"])
+            call(["git", "add", "."])
+            call(["git", "commit", "-m", "Initial Commit"])
+
+            result = runner.invoke(_main, ["--date", "01-01-2001", "--keep"])
+
+            self.assertEqual(0, result.exit_code)
+            path = "NEWS.rst"
+            self.assertTrue(os.path.isfile(path))
```
adiroiban: Maybe also have an assertion before calling the CLI that the NEWS file doesn't exists.

```suggestion
            # The NEWS file is created.
            # So this is not just `--draft`.
            self.assertTrue(os.path.isfile("NEWS.rst"))
```

src/towncrier/test/test_build.py
```diff
@@ -407,6 +407,56 @@ def test_no_confirmation(self):
...
+            self.assertTrue(os.path.isfile(fragment_path1))
+            self.assertTrue(os.path.isfile(fragment_path2))
+
+    def test_yes_keep_error(self):
+        runner = CliRunner()
```
adiroiban: Can you please add a description to this test.

I am not sure why you are calling `build` twice

We don't have tests for the test code, so a docstring for the tests function is the only think that we have to assert that the test is valid.

maybe also assert that the news file is not created and the fragment files are not removed.

```suggestion
    def test_yes_keep_error(self):
        """
        It will fail to perform any action when the
        conflicting --keep and --yes options are provided.
        """
        runner = CliRunner()
```

adiroiban (changes requested):
This looks very good. Many thanks for writing the docs and tests.

It helps a lot with the review.

I left a few minor comments.

The biggest comment is the lack of docstring for the new tests.
Please use it to describe the targeted end-user functionality.

The towncrier project don't really has a good knnowledge of what it should do and how it should do.
Towncrier is pretty flexible and various people are using it for their use cases... and without a docstring for a tests, it's hard to know what is a side effect and what is an intended feature.

------------

I understand the use case, and I think that this is specific enough not to be included in the tutorial.

-----

> If you add new configuration options (or change the meaning of existing ones), make sure docs/configuration.rst reflects those changes.

@hynek  I think that here we should say

```diff
-If you add new configuration options
+If you add new template options
```
This PR has only CLI options so I think that we are fine.

src/towncrier/_git.py
fizyk: Hmm.... you're right

src/towncrier/test/test_build.py
fizyk: I call it twice because the click triggers a callback in the order it parses the command line, not in the order in which they are defined in the command. That's why I also added the validator to both `--yes` and `--keep` 

I'll add the comment to test as well.

docs/cli.rst
```diff
@@ -50,6 +50,11 @@ Build the combined news file from news fragments.
    Do not ask for confirmations.
    Useful for automated tasks.
 
+.. option:: --keep
+
+   Do not ask for confirmations. But keep news fragments.
+   Useful for automated tasks. Where you have to generate several different news fragments for different parties.
+
```
adiroiban: I think the important part here is that fragments are not removed.

I don't know if "Useful for automated tasks" adds any new info on top of "Don't ask for confirmation".
If it doesn't ask, it should be good for automation.

I don't know what to day about the specific use case.
From this one sentence, I don't fully undestand what "Where you have to generate several different news fragments for different parties." should do.

If we want to have this information in the docs,
I think we should create a separate doc page with use case scenarios.

Something similar to "How to keep changelong in markdown"

In this way we can add more info.

I guess that you are using towncrier to generate a "Release-Notes-Everyting" and a separate "Release-Notes-Bugs" and "Release-Notes-Security" and "Release-Notes-Features"... but it is not clear.

```suggestion
   Don't delete news fragments after the build and don't ask for confirmation whether to delete or keep the fragments.

```

src/towncrier/_git.py
```diff
@@ -10,18 +10,26 @@
...
-def remove_files(fragment_filenames: list[str], answer_yes: bool) -> None:
+def remove_files(
+    fragment_filenames: list[str], answer_yes: bool, answer_keep: bool
+) -> None:
     if not fragment_filenames:
         return
 
     if answer_yes:
         click.echo("Removing the following files:")
+    elif answer_keep:
```
adiroiban: Just a very minor comment. We can merge with this.
Do we need to be verbose here ?

I guess that it can help with troubleshooting to detect which files are used for the build process... so we can keep it.

We can just return ASAP and remove_files is a NOOP for `answer_keep = Yes`

adiroiban (approved):
Thanks. Changes looks good.
Only very minor comments that should not be a blocker.

Somehow, I have messed up the coverage reporting back to GitHub PR.
The coverage is ok, but posting back the comment fails.

I will try to fix `trunk` and then we can merge this.

Many thanks again for your help!

src/towncrier/_git.py
fizyk: Hmm....  This could kill two birds with one stone, not sure how would you feel about it though

```python
def remove_files(
    fragment_filenames: list[str], answer_yes: bool, answer_keep: bool
) -> None:
    if not fragment_filenames:
        return

    try:
        if answer_yes:
            click.echo("Removing the following files:")
        elif answer_keep:
            click.echo("Keeping the following files:")
            # Not proceeding with the removal of the files.
            return
        else:
            click.echo("I want to remove the following files:")
    finally:
        # Will always be printed, even for answer_keep to help with possible troubleshooting
        for filename in fragment_filenames:
            click.echo(filename)

    if answer_yes or click.confirm("Is it okay if I remove those files?", default=True):
        call(["git", "rm", "--quiet"] + fragment_filenames)
```

adiroiban: Thanks for the updates.

Usually, I don't like clever code :)

But if you like it, I am ok to merge something like this... maybe with an extra comment.
But I think the code is not that hard to read.
I don't see any good reason why not to have this code approved.

Maybe remove "elif". I as general rule, I try to code with as little else/elif as possible.

```python
def remove_files(
    fragment_filenames: list[str], answer_yes: bool, answer_keep: bool
) -> None:
    if not fragment_filenames:
        return

    try:
        if answer_keep:
            # Not proceeding with the removal of the files.
            click.echo("Keeping the following files:")
            return  # See the logic from finally.

        if answer_yes:
            click.echo("Removing the following files:")
        else:
            click.echo("I want to remove the following files:")
    finally:
        # Always show the source files,
        # even for answer_keep to help with possible troubleshooting
        for filename in fragment_filenames:
            click.echo(filename)

    if answer_yes or click.confirm("Is it okay if I remove those files?", default=True):
        call(["git", "rm", "--quiet"] + fragment_filenames)
```

adiroiban:
I/we still need to fix the coverage reporting issue.

I did an initial investigation, and it looks like forks only get read permission

I can see that for the main repo, the permissions are `write` ... for example for the latest trunk merge

https://github.com/twisted/towncrier/actions/runs/3404794371/jobs/5662304592#step:1:17

But for this PR the permissions are only `read`

https://github.com/twisted/towncrier/actions/runs/3563302652/jobs/5985987960#step:1:17


@hynek @altendky Do you any a quick fix for standalone coverage reporting on forks ?

Thanks!

hynek:
Maybe https://github.com/py-cov-action/python-coverage-comment-action? I think actions have different token handling

src/towncrier/test/test_build.py
```diff
@@ -407,6 +407,70 @@ def test_no_confirmation(self):
...
 
+    def test_keep_fragments(self):
+        """
+        The `--keep` option will build the full final news file
+        without deleting the fragment files and without
+        any extra CLI interaction or confirmation.
+        """
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
```
hynek: Could you use the with_isolated_runner decorator in new code please?

https://github.com/twisted/towncrier/blob/trunk/src/towncrier/test/helpers.py#L23

src/towncrier/test/test_build.py
```diff
@@ -407,6 +407,70 @@ def test_no_confirmation(self):
...
+        It will fail to perform any action when the
+        conflicting --keep and --yes options are provided.
+
+        Called twice with the different order of --keep and --yes options
+        to make sure both orders are validated since click triggers the validator
+        in the order it parses the command line.
+        """
+        runner = CliRunner()
+
+        with runner.isolated_filesystem():
```
hynek: Same as above

hynek (changes requested):

adiroiban:
>  I think actions have different token handling

As far as I know, the actions are just a "step" , so they run in the context of the job.

We are using the default token...and the default token only has read permissions.
[As documented here](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)

There is this setting [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions)  
but I don't know what are the side effects

src/towncrier/test/test_build.py
fizyk: done :)

hynek: Thanks! :)

hynek:
@adiroiban sadly I'm rather clueless at all these things. If you want to just quickly unblock contributions I can offer [my method](https://hynek.me/articles/ditch-codecov-python/) that doesn't comment, but at least works for now. People just have to look into the CI output for missing lines.

adiroiban:
Thanks @hynek . Yes, the commit status might be the only "safe" option.

In theory, we could give write access to forks, since our workflow has explicit permissions... but if in the future a new workflow is created without explicit permissions, this can lead to a security void, as we might accidentally give full access to the repo to forks.

At the same time, I think that new contributors need to have their first PR manually approved, so the workflow code is not executed right away, unattended.

I will look to create a new PR that will always send the commit status and will post a comment only for non-forks.

src/towncrier/_git.py
```diff
@@ -10,17 +10,26 @@
 import click
 
 
-def remove_files(fragment_filenames: list[str], answer_yes: bool) -> None:
+def remove_files(
+    fragment_filenames: list[str], answer_yes: bool, answer_keep: bool
+) -> None:
```
hynek: so, generally speaking I'm not a fan of this shell-style Python (that towncrier is full of and not specific to this PR – I'm just trying to have less of it :))

So while we're at refactoring: could we split this up into a function that does _not_ live within `_git.py` & returns a list of files that should be removed and minimizing `remove_files` to something resembling:

```python
def remove_files(files: list[str]) -> None
    if files:
        call(["git", "rm", "--quiet"] + files)
```

fizyk: sure :)

fizyk: done :)

src/towncrier/_remover.py
```diff
@@ -0,0 +1,30 @@
...
+from __future__ import annotations
+
+import click
+
+from towncrier._git import remove_files
+
+
+def remove_news_fragment_files(
+    fragment_filenames: list[str], answer_yes: bool, answer_keep: bool
+) -> None:
```
hynek: Close, but not quite what I had in mind.

We haven't really achieved anything by this refactoring, because this function still deletes the files which makes it an action with pretty heavy side-effects.

My suggestion was that this function returns the `fragment_filenames` as a list and the caller calls `_git.remove_files()` on it unconditionally.

---

Secondarily, you've moved this function into a separate module, so you can easily monkeypatch it without affecting the CLI entry point – correct?

That's pretty heavy "test damage".

How about going a bit functional here:

```python
_git.remove_files(
    find_news_fragment_files_for_removal(  # <-- better name for remove_news_fragment_files after changes
        fragment_filenames, answer_yes, answer_keep,
        confirm_fn=partial(click.confirm, "Is it okay if I remove those files?", default=True)
    )
)
```

In the test you would simply pass `confirm_fn=lambda: True` or `confirm_fn=lambda: False` – look ma no patching!

I would usually not suggest something like that on a public repo, but if the alternative is adding a new module for just one function that is called in one place, it seems worth it – would you agree?

src/towncrier/newsfragments/129.feature
```diff
@@ -0,0 +1 @@
+Added `--keep` option to build command that allows to generate newsfile but keep newsfragments in place. This option can not be used together with `--yes`.
```
hynek:
```suggestion
Added ``--keep`` option to the ``build`` command that allows to generate a newsfile, but keeps the newsfragments in place.
This option can not be used together with ``--yes``.
```

src/towncrier/_remover.py
fizyk: Hmmm.... you're right. I got lost a bit between reading your suggestion and sitting down to implement it.

---

When I look at it, the intermediate function only makes a decision on whether to remove the files or not. And would return the fragments_list or not. 
I wonder if it wouldn't be even better if we had something like this:

```python
if should_remove_news_fragment(fragment_filenames, answer_yes, answer_keep):
    _git.remove_files(fragment_files)
```

Then both functions here could be tested separately. And... still, the should_remove_news_fragments could receive confirm_fn partial...

---

Sorry, I tend to split functions sooner, especially if the original module is already several hundred lines big. I'll move it back since... you're right, it's not called anywhere else, and that's a perfect rule to keep 🤔 .

fizyk: @hynek I don't think the functional approach will help in regards to the changed test.

The test that got the mock updated is an integration one (tests from all the way up by calling click command) and... I moved the click calls from `_git.remove` so unless I'll keep the intermediate within `_git` module, you'd have to change the mock anyway 🤔 

And to actually make use of the functional approach the test itself would have to be a unit one.
I'm not sure how to approach it.
And the last question, shall I move the intermediate function into the build module?

hynek: > Sorry, I tend to split functions sooner, especially if the original module is already several hundred lines big.

yeah the CLI entry points need to be refactored HARD, but let's not blow this up even further 😇

---

since the list is never changed, agreed on `should_remove_news_fragments`

---

ok this click.confirm business is annoying. I was gonna suggest an abstraction, but I've noticed that it's literally only used in that one place so I'm gonna say to KISS and patch away on build.py without the confirm callable, unless you'd like to add more *unit*tests.

fizyk: > but let's not blow this up even further

whew 😌 

---
> unless you'd like to add more unittests

I can't think of any that the integration tests wouldn't already cover... so not at this time. 
I'll try to remember next time I make a contribution though :)

hynek (approved):
sorry for the delay, I've been sick all week.

i think given the circumstances this is good to go.

adiroiban:
Many thanks @fizyk for the changes and @hynek for the review.
I was busy at work pushing a review before the holiday break.

I think we can also look at pushing a new release for towncrier... to get this feature and also get latest documentation

hynek:
Agreed – having the news docs would be great!

fizyk:
@hynek @adiroiban thank you both for the review, discussion and satisfactory work :)
And I'm glad I could also close one issue from someone's issue board :)

PR #455, opened 2022-12-19
Revert Generate coverage reports using only GitHub Actions

hynek:
As discussed in https://github.com/twisted/towncrier/pull/447#issuecomment-1357529374


adiroiban (approved):
Thanks!

hynek:
the irony of codecov hanging is not lost on me

Issue #353, opened 2021-06-29
Github releases integration


elgalu:
Hi, thanks for this nice tool, question,

Does it support [GitHub releases](https://github.com/github/hub) ? i.e. to automatically `hub release create -m ${MESSAGE} $(date +%Y%m%d%H%M%S)` with the same rst/markdown content that get's added to the Changelog? e.g.

```sh
# Create a GitHub Release from master using release notes from a file
# https://hub.github.com/hub-release.1.html
hub release create --copy -F release-notes.rst v2.3.0
```

elgalu:
Actually it would probably be better to use the official <https://github.com/cli/cli>

```sh
gh release create 0.0.2 --title "# test gh release title" --notes "# test gh release notes" -F CHANGELOG.md dist/slurmops-0.0.1*
#=> https://github.com/elgalu/some-project/releases/tag/0.0.2
```

webknjaz:
I suppose you could use `--draft` to get the contents of the unreleased notes in your release pipeline and then send that to GitHub releases.

adiroiban:
I don't understand what is the requested functionality.

AFAIK, GitHub releases is just GitHub proprietary take on git annotated tags.

You can run towncrier to create the `release-notes.rst` file and then just call GitHub proprietary tools or API to create the release.

webknjaz:
> AFAIK, GitHub releases is just GitHub proprietary take on git annotated tags.

It's actually a bit of a wrapper that is bound to a Git tag, may have an extra piece of markdown attached, a pre-release label and a linked gh discussion.

> You can run towncrier to create the `release-notes.rst` file and then just call GitHub proprietary tools or API to create the release.

I'd like to add that many people using RST would want to post-process it by converting Sphinx-specific roles into something simpler with `sed` and running `pandoc` on that to convert it into markdown since GH won't render the note as RST.

Check out https://github.com/ansible/pylibssh/blob/3b0beb1/.github/workflows/build-test-n-publish.yml#L469-L516 for an example of conversion and, as a bonus, also outputting it to the GH job summary like this: https://github.com/ansible/pylibssh/actions/runs/3705179529.

PR #456, opened 2022-12-20
Release 22.12.0

adiroiban:
# Description

This is the release branch for 22.12.0

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -4,7 +4,7 @@
 
 
 towncrier 22.12.0 (2022-12-21)
-=============================
+==============================
 
 No changes since the previous release candidate.
 
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -2,6 +2,13 @@
 
 .. towncrier release notes start
 
+
+towncrier 22.22.0 (2022-12-21)
+=============================
+
+No changes since the previous release candidate.
+
+
 towncrier 22.12.0rc1 (2022-12-20)
 ==================================
 
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,7 +3,7 @@
 .. towncrier release notes start
 
 
-towncrier 22.22.0 (2022-12-21)
+towncrier 22.12.0 (2022-12-21)
 =============================
 
 No changes since the previous release candidate.
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -10,7 +10,7 @@ No changes since the previous release candidate.
 
 
 towncrier 22.12.0rc1 (2022-12-20)
-==================================
+=================================
 
 Features
 --------
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -2,6 +2,44 @@
 
 .. towncrier release notes start
 
+towncrier 22.12.0rc1 (2022-12-20)
+==================================
+
+Features
+--------
+
+- Added ``--keep`` option to the ``build`` command that allows generating a newsfile, but keeps the newsfragments in place.
+  This option can not be used together with ``--yes``. (`#129 <https://github.com/hawkowl/towncrier/issues/129>`_)
+- Python 3.11 is now officially supported. (`#427 <https://github.com/hawkowl/towncrier/issues/427>`_)
+- You can now create fragments that are not associated with issues. Start the name of the fragment with ``+`` (e.g. ``+anything.feature``).
+  The content of these orphan news fragments will be included in the release notes, at the end of the category corresponding to the file extension.
+
+  To help quickly create a unique orphan news fragment, ``towncrier create +.feature`` will append a random string to the base name of the file, to avoid name collisions. (`#428 <https://github.com/hawkowl/towncrier/issues/428>`_)
+
+
+Improved Documentation
+----------------------
+
+- Improved contribution documentation. (`#415 <https://github.com/hawkowl/towncrier/issues/415>`_)
+- Correct a typo in the readme that incorrectly documented custom fragments in a format that does not work. (`#424 <https://github.com/hawkowl/towncrier/issues/424>`_)
+- The documentation has been restructured and (hopefully) improved. (`#435 <https://github.com/hawkowl/towncrier/issues/435>`_)
+- Added a Markdown-based how-to guide. (`#436 <https://github.com/hawkowl/towncrier/issues/436>`_)
+- Defining custom fragments using a TOML array is not deprecated anymore. (`#438 <https://github.com/hawkowl/towncrier/issues/438>`_)
+
+
+Deprecations and Removals
+-------------------------
+
+- Default branch for `towncrier check` is now "origin/main" instead of "origin/master".
+  If "origin/main" does not exist, fallback to "origin/master" with a deprecation warning. (`#400 <https://github.com/hawkowl/towncrier/issues/400>`_)
+
+
+Misc
+----
+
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -81,7 +81,7 @@ ignore_missing_imports = true
 requires = [
     "setuptools ~= 44.1.1",
     "wheel ~= 0.36.2",
-    "incremental == 17.5.0",
+    "incremental == 22.10.0",
```
adiroiban, line 84: The incremental related tests was failing for the release candidate version format...so at first I try updating incremental with the hope that normalization is fixed.

It is not... but maybe is ok to ask for a newer incremental.

Best, would be to not depend on incremental :)

Let me know if this should be reverted or moved to a separate PR.

src/towncrier/_project.py
```diff
@@ -13,7 +13,7 @@
 from importlib import import_module
 from types import ModuleType
 
-from incremental import Version
+from incremental import Version as IncrementalVersion
```
adiroiban, line 16: I renamed this as the generic `Version` was a bit confusing

src/towncrier/test/test_project.py
```diff
@@ -48,14 +48,14 @@ def test_tuple(self):
...
-    @skipIf(version is None, "Needs importlib.metadata.")
+    @skipIf(metadata_version is None, "Needs importlib.metadata.")
     def test_incremental(self):
         """
         An incremental Version __version__  is picked up.
         """
         pkg = "../src"
 
-        self.assertEqual(version("towncrier"), get_version(pkg, "towncrier"))
+        self.assertEqual(metadata_version("towncrier"), get_version(pkg, "towncrier"))
```
adiroiban, line 58: This is the failing test for RC.

I renamed to `metadata_version` as the generic `version` was a bit confusing.

NEWS.rst
```diff
@@ -2,6 +2,44 @@
...
 .. towncrier release notes start
 
+towncrier 22.12.0rc1 (2022-12-20)
+==================================
+
+Features
+--------
+
+- Added ``--keep`` option to the ``build`` command that allows generating a newsfile, but keeps the newsfragments in place.
+  This option can not be used together with ``--yes``. (`#129 <https://github.com/hawkowl/towncrier/issues/129>`_)
```
hynek, line 19: we should update the template – all PR links still point to Hawkie

src/towncrier/_project.py
```diff
@@ -50,8 +50,12 @@ def get_version(package_dir: str, package: str) -> str:
     if isinstance(version, str):
         return version.strip()
 
-    if isinstance(version, Version):
-        return version.base().strip()
+    if isinstance(version, IncrementalVersion):
+        # FIXME:https://github.com/twisted/incremental/issues/81
```
adiroiban, line 54: Not sure if this hack is ok... I am not happy with it.. but at the same time I don't know what to do.

I would rather have a green test, than remove the test.

adiroiban (commented):
This should be ready for review

adiroiban:
The release candidate and tag have not yet been created in PyPI and I am waiting for an initial review of this PR.
I am not sure if the changes are ok... they result in green tests, but might not be the best way to fix the tests.

Thanks!

hynek (approved):
yolo

adiroiban:
Thanks for the review. I will try to follow the process from [here](https://towncrier.readthedocs.io/en/latest/release.html) and do a public pre-release on PyPI

adiroiban:
release candidate triggered

* here https://github.com/twisted/towncrier/releases/tag/22.12.0rc1
* and here https://github.com/twisted/towncrier/actions/runs/3741354542

If all is OK I will update the version and do the final release.

webknjaz:
@adiroiban hey, I noticed https://github.com/twisted/towncrier/releases/tag/22.12.0rc1.

May I suggest you to use the `☑️ Create a discussion for this release` checkbox when making the final release? This allows having discussions linked to GH Releases, creating a dedicated space for people to share a release feedback in a consolidated place. Here's an example of what it looks like: https://github.com/ansible-community/ansible-test-gh-action/releases/tag/v1.12.0 — there's a `💬 Join discussion` button in the bottom right corner that people may follow.

P.S. This probably requires enabling Discussions in the repo settings before making the release. In older repos, it is often needed to also create a separate category, typically called `📢 Announcements` and using the `🔒 Announcement` discussion format. I also typically type `Updates from maintainers` into the description field of the category.

---

Another observation is that you pasted RST there, but it only supports Markdown. In https://github.com/twisted/towncrier/issues/353#issuecomment-1359432962, I've linked to a GHA-based example, but a short TL;DR version for local invocations would be:
```shell
sudo apt install -y pandoc && pandoc --from=rst --to=gfm --output=changelog.md changelog.rst
```

N.B. `gfm` stands for GitHub-Flavored Markdown.

webknjaz:
> and here [twisted/towncrier/actions/runs/3741354542](https://github.com/twisted/towncrier/actions/runs/3741354542)

If you look at https://github.com/twisted/towncrier/actions/runs/3741354542/jobs/6350891579#step:9:19 closer, you'll see warnings being emitted:
```console
Checking dist/towncrier-22.12.0rc1-py3-none-any.whl: PASSED with warnings
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.
Checking dist/towncrier-22.12.0rc1.tar.gz: PASSED with warnings
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.
```

P.S. In general, it's a good idea to run `twine check --strict` at early stages. My pypi-publish action also has a strict mode, but it's not pleasant to see hard failures at the very last moment before the release.

adiroiban:
PyPi release candidate is live https://pypi.org/project/towncrier/22.12.0rc1/

----------

Thanks @webknjaz  for the input

I was expecting that we already run something like `twine check --strict` :(
We should do it... in a separate PR :)

--------

This is what I have when creating a GitHub release

I don't see the create discussion button... I think it needs the "discussion" feature.

webknjaz:
> I don't see the create discussion button... I think it needs the "discussion" feature.

It'll be a drop-down when discussions are enabled. Note that you can only do this on GH Release creation, updating an existing one won't work.

webknjaz:
Happy to report that the pre-release doesn't break the smoke tests of my Sphinx extension: https://github.com/sphinx-contrib/sphinxcontrib-towncrier/actions/runs/3742185592.

adiroiban:
We don't close the comments on a PR. I think we can use the release PR for discussion.
There is not much activity over town crier. I think the discussions for the PRs and issues are more that we need.

I don't have much time for towncrier.
I think the current release notes from GitHub releases are good enough. Yes...the links are not nice...but we should be fine.  

In theory with the new --keep feature we should be able to have 2 templates...one for RST and one for Markdown

When possible I try  not to sell my soul to platform specific features 😀

NEWS.rst
```diff
@@ -2,6 +2,13 @@
 
 .. towncrier release notes start
 
+
+towncrier 22.22.0 (2022-12-21)
```
pawelad: I believe there's a typo here - `22.22.0` instead of `22.12.0`.

adiroiban: Thanks. Much appreciated

adiroiban:
22.10.0 GHA release triggered [here](https://github.com/twisted/towncrier/actions/runs/3750164167)

[PyPI is ready](https://pypi.org/project/towncrier/)

----------

Many thanks for your help with this release.

---------

Just a note. It's hard for me to find free time for doing/driving the towncrier release.
If anyone want to help with driving / pushing new towncrier releases, the release process is documented here https://towncrier.readthedocs.io/en/stable/release.html

I would be happy to guide and assist

webknjaz:
> We don't close the comments on a PR. I think we can use the release PR for discussion. There is not much activity over town crier. I think the discussions for the PRs and issues are more that we need.

Many folks don't like leaving comments on closed things. Besides, issues/PRs are more about bugs then feedbacks. This PR diff does not include all the changes that got into the release anyway. Discussions have a different structure and semantic purpose. Many maintainers experimented sending folks to Discussions for community support to keep the bug tracker filled with things they plan to work on. Here's one example of how it's done: https://github.com/pradyunsg/furo/issues/new/choose.

> I don't have much time for towncrier. I think the current release notes from GitHub releases are good enough. Yes...the links are not nice...but we should be fine.
> 
> In theory with the new --keep feature we should be able to have 2 templates...one for RST and one for Markdown

That wouldn't work well because the fragments may use bits of specific syntax. So if people are enclosing RST fragments, and they are injected into the changelog via Jinja without any sort of conversion, you'll get a “Markdown” file filled with RST syntax all over the place and that's exactly what we see now.

> When possible I try not to sell my soul to platform specific features :grinning:

True, but PRs are also platform-specific to platforms. They aren't part of Git :-P
I like how discussions provide a place for people to interact with each other without disturbing the maintainers.

webknjaz:
> If anyone want to help with driving / pushing new towncrier releases, the release process is documented here [towncrier.readthedocs.io/en/stable/release.html](https://towncrier.readthedocs.io/en/stable/release.html)

I can look at automating the process more sometime, maybe, if that'll help.

adiroiban:
**Discussions** are now back online.

I love automation :) so, bring it in :)

PR #460, opened 2022-12-21
Use a GitHub Action to post the pull request checklist as a comment

namurphy:
# Description

This PR moves the checklist from the pull request template into a separate file that gets added as a comment to pull requests.  This makes the pull request template a lot less cluttered.  I also added a short greeting, along with hyperlinks to the files mentioned in the checklist.  

To see this GitHub Action in action, check out [this comment](https://github.com/PlasmaPy/PlasmaPy/pull/1808#issuecomment-1329924433).

The maintainers should feel free to edit this PR too.

# Checklist


altendky:
I haven't reviewed https://github.com/harupy/comment-on-pr, but I feel like this will run into the same permissions issues with external PRs that was causing trouble around the coverage reports @adiroiban?

adiroiban:
Thanks for the PR.

Nick, I don't think that the action will work for fork PRs.

I tried to check the source project at https://github.com/harupy/comment-on-pr but I can't see any active PR from a non-team member.

I also tried to look for an example on PlasmaPy, but I couldn't find a PR from a non contributors.

Also, for first time contributors, the GitHub Actions workflow is not trigerred by default.

------------

We tried to implement something similar but the automatic GitHub API token generated by GitHub Action has read-only permissions for forks.

You can check a read-only token here https://github.com/twisted/towncrier/pull/453/checks#step:1:17

And the exact same workflow, triggered by a non-fork team member here

https://github.com/twisted/towncrier/actions/runs/3756273445/jobs/6382171392#step:1:17

----------

I am not against this change, but I am not sure it will work.

namurphy:
> I also tried to look for an example on PlasmaPy, but I couldn't find a PR from a non contributors.

Here's an example from a first-time contributor where the checklist comment was posted: https://github.com/PlasmaPy/PlasmaPy/pull/1830

> Also, for first time contributors, the GitHub Actions workflow is not trigerred by default.

It appears that the checklist comment on that PR was posted right after the first commit.  (By clicking on the ❌ next to the initial commit, it looks like the workflow wasn't triggered for that commit.)

> I am not against this change, but I am not sure it will work.

What would you think of merging this to test it?  I could submit a quick follow-up PR to perhaps add a pre-commit hook, and we could see if the comment gets posted.  If it doesn't, then I would submit a PR to revert the change.

adiroiban:
It looks like the action for that PR is executed with [full write permissions](https://github.com/PlasmaPy/PlasmaPy/pull/1830/checks#step:1:17).

```
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

--------

And I can see that the change from this PR creates a new workflow that uses default API token permissions.

But for towncrier, we want restricted write permission.

Now, if we just enable `Issues: write` permission, (without full write) I think that should be ok, and this can be merged.

I don't think that it is worth compromising the security, just to have a checklist as a comment.

------------

At the same time, I think that you might want to know the list, before creating a PR.
Otherwise, each time you submit a PR, you will not know what checklist to follow.

-------------

A note for the team, while reviewing the default security  for GitHub API token for GitHub action I observed that workflows were setup with read and write access by default.

I reconfigured to default to read-only.

image

.github/workflows/checklist.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+- pull_request_target
+
+jobs:
+  comment:
+    name: Checklist
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: harupy/comment-on-pr@master
```
adiroiban: and here, if we decided to use this action, I think that we should not use `master`

We should pin a certain commit.

This will make sure that things will not break.

At the same time, if somehow `harupy/comment-on-pr` repo is accidentally compromised, we will not automatically run the latest code.

At least we run a known commit ID.

----------

There is the still possibility for the whole repo to be compromises so that a trusted commit `abc123` from 3 years ago is replaced with a new commit with the same `abc123` hash , but with malicious code.

adiroiban:
I think that we need to be a bit extra careful with `towncrier` in terms of security.

For a release, I am running towncrier manually and I always review what it does.

But there might be people out there doing full automation and they might have towncrier automatically called as part of their automated release/publish pipeline.

In this way, if there is a security issue in towncrier, it can be propagated to the final release.

exarkun:
This seems like needless complexity and needless additional coupling to GitHub.  "This makes the pull request template a lot less cluttered" doesn't seem like much of a motivation.  "Cluttered" is a vague criticism.  One could just as easily argue that the new thing is "cluttered".

I am not an active towncrier maintainer or contributor but this isn't the kind of change I would welcome on any projects where I am active.

namurphy:
We've found it to be quite helpful in the project I work most with, though we started with a more cluttered PR template than the one that towncrier currently has and don't have as pressing security needs.  I'm not an expert on security, so I'll leave it up to the maintainers. If it would take a while to figure out if there are any security issues, it's probably not worth investing too much time on given other pressing needs.

If we do continue the discussion here, I would gently remind us about the [code of conduct](https://github.com/twisted/towncrier/blob/trunk/CODE_OF_CONDUCT.md).

.github/workflows/checklist.yml
namurphy: Thank you — that is a really good point!  This is advice I should probably apply in some other repositories also.

adiroiban:
Considering the security implications and the extra added complexity I am -1 on this.

I continued the conversation since I was curious how this can be done in a secure way.

In GitHub Actions, we store a PyPI publish token/secret and we want to make sure this stays safe. Anyone that get holds of that token can publish malicious towncrier packages to pypi.

And if we are not careful, someone can find that token by just having GitHub Action execute something like `echo ${{ secrets.PYPI_TOKEN }}`

adiroiban:
In terms of security, we can run this action in the context of `pull_request_target` instead of `pull_request` but we need to be extra careful to make sure we don't leak the secrets.

I think that for the PR checklist comment, it's not worth the risk and complexity.

.github/workflows/checklist.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+- pull_request_target
+
+jobs:
+  comment:
+    name: Checklist
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v0.1.0
+    - uses: harupy/comment-on-pr@v0.1.0
```
adiroiban, line 12: We should be even more strict, and pin a commit not the tag.

A tag can be updated with a simple write, so you end up with different souce code for the same tag.

The source code for the same commit sha can only be changed if you rewrite the whole history, and find a way to generate the exact same  sha.

```suggestion
    - uses: harupy/comment-on-pr@d83982f39ec865c951b0d1ad0e9e0d79d801e790
```

PR #462, opened 2022-12-22
Add ongoing package linting & inspection

### LINKED ISSUE #457: Add strict packaging validation as part of CI checks
> and here [twisted/towncrier/actions/runs/3741354542](https://github.com/twisted/towncrier/actions/runs/3741354542)

If you look at https://github.com/twisted/towncrier/actions/runs/3741354542/jobs/6350891579#step:9:19 closer, you'll see warnings being emitted:
```console
Checking dist/towncrier-22.12.0rc1-py3-none-any.whl: PASSED with warnings
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.
Checking dist/towncrier-22.12.0rc1.tar.gz: PASSED with warnings
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.
```

P.S. In general, it's a good idea to run `twine check --strict` at early stages. My pypi-publish action also has a strict mode, but it's not pleasant to see hard failures at the very last moment before the release.

_Originally posted by @webknjaz in https://github.com/twisted/towncrier/issues/456#issuecomment-1359524318_

hynek:
Fixes #457

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - uses: hynek/build-and-inspect-python-package@v1
+    - uses: hynek/build-and-inspect-python-package@f336040a31ea03a2551adcc994d3ddb2f3ea3f7f
 
     - name: Set up ${{ matrix.python.name }}
       uses: actions/setup-python@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
+    - uses: hynek/build-and-inspect-python-package@v1
+
     - name: Set up ${{ matrix.python.name }}
       uses: actions/setup-python@v4
       with:
@@ -189,14 +191,6 @@ jobs:
     - run: nox -e ${{ matrix.task.nox }}
       if: ${{ matrix.task.run-if }}
 
-  package:
-    name: Build & verify package
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v3
-      - uses: hynek/build-and-inspect-python-package@v1
-
   pypi-publish:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
     name: Check tag and publish
@@ -256,7 +250,6 @@ jobs:
       - test-linux
       - test-windows
       - check
-      - package
       - pypi-publish
     steps:
       - name: Require all successes
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,6 +189,13 @@ jobs:
     - run: nox -e ${{ matrix.task.nox }}
       if: ${{ matrix.task.run-if }}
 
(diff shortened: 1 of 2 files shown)


hynek:
So what this revealed is that we've been packaging `src/towncrier.egg-info` in our SDist all along and I don't know why and I don't know how to stop it. 😳

.github/workflows/ci.yml
```diff
@@ -189,6 +189,13 @@ jobs:
     - run: nox -e ${{ matrix.task.nox }}
       if: ${{ matrix.task.run-if }}
 
+  package:
+    name: Build & verify package
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: hynek/build-and-inspect-python-package@v1
```
adiroiban: Should we pin this to a specific commit?

Similar to that we did for

```
uses: pypa/gh-action-pypi-publish@37f50c210e3d2f9450da2cd423303d6a14a6e29f
```

This is mainly for security reasons.

.github/workflows/ci.yml
```diff
@@ -249,6 +256,7 @@ jobs:
       - test-linux
       - test-windows
       - check
+      - package
```
adiroiban: This is ok... but maybe `pypi-publish` should depende on `package` .

Maybe we want to block a release if the package is not valid.

.github/workflows/ci.yml
```diff
@@ -189,6 +189,13 @@ jobs:
     - run: nox -e ${{ matrix.task.nox }}
       if: ${{ matrix.task.run-if }}
 
+  package:
+    name: Build & verify package
```
adiroiban:
```suggestion
  package-validation:
    name: Verify package
```

I think that at this point, the build part is secondary.

We will not publish the package that is built in this job.
So having the "build" name, might suggest that this is the part where we build the package that will be published.

I know that most of the time, the package built here is the same as the one that is built in the "build" job (line 14)

```
jobs:
  build:
```

And maybe, we should not have a separate job.
Instead, just add this step to the existing `build` job.

adiroiban (approved):
Many thanks for the PR.

I have a few medium important comments.

Nothing that should block this PR.
We can update the changes later in a separate PR.

The important part is to make sure we get a "red-alert" when the package is not ok.

----------

I think that pinning the `hynek/build-and-inspect-python-package@v1` to a specific commit instead of a tag is important.

I will let you decide if this is a security issue or not.

.github/workflows/ci.yml
hynek: is someone keeping those pins up to date?

adiroiban: I think that GHA dependencies should be handled as any other testing / build / CI dependency.

And as a general rule, we should pin every package that we use during the build system

Now, another question.

Do "hynek/build-and-inspect-python-package" need to be a GHA action ?
Can it be distributed installed and executed just like any other python package?

Just asking.

adiroiban (approved):
thanks for the updates. I think this can be merged.

.github/workflows/ci.yml
hynek: That would be pointless since it doesn't do much novel things except calling established tools. This way you add only one like of YAML and it will just work, including handling package dependencies.

PR #468, opened 2023-01-17
Fix `create` for a bare orphan in a section not adding random hex

SmileyChris:
# Description

When running `towncrier create section/+.feature`, the random hex isn't appended.

# Checklist


src/towncrier/test/test_create.py
```diff
@@ -172,17 +172,39 @@ def test_create_orphan_fragment(self):
         """
         runner = CliRunner()
 
+        fragments = []
         with runner.isolated_filesystem():
```
hynek: Could this be `@with_isolated_runner` instead?

src/towncrier/test/test_create.py
```diff
@@ -172,17 +172,39 @@ def test_create_orphan_fragment(self):
...
+            self.assertEqual(0, result.exit_code)
+            result = runner.invoke(
+                _main, ["subsection/+.feature"], catch_exceptions=False
+            )
+            self.assertEqual(0, result.exit_code, result.output)
+
+            for _, dirs, files in os.walk("foo/newsfragments", topdown=False):
+                for file in files:
+                    fragments.append("/".join(dirs + [file]))
+                    print(dirs)
```
hynek: oops

src/towncrier/test/test_create.py
```diff
@@ -172,17 +172,39 @@ def test_create_orphan_fragment(self):
...
+            result = runner.invoke(_main, ["+.feature"])
+            self.assertEqual(0, result.exit_code)
+            result = runner.invoke(
+                _main, ["subsection/+.feature"], catch_exceptions=False
+            )
+            self.assertEqual(0, result.exit_code, result.output)
+
+            for _, dirs, files in os.walk("foo/newsfragments", topdown=False):
+                for file in files:
+                    fragments.append("/".join(dirs + [file]))
```
hynek: Given you're using os.path everywhere, this should probably be something like `os.path.join(*dirs, file)`?

hynek (changes requested):
Since you're now handling prefixes of different lengths, do we have a test for that case? It's easy to catch an off-by-1 here.

src/towncrier/test/test_create.py
SmileyChris: Sure, I was just copying the style of the tests surrounding when I wrote this originally, but I'll switch to that decorator.

src/towncrier/test/test_create.py
SmileyChris: Too used to pytest that just hides print statements unless the test fails :P

src/towncrier/test/test_create.py
SmileyChris: Again, the style of the other surrounding tests hard code '/', and I was trying to minimize changes - but sure!

SmileyChris:
Test updated, and new test added to check for custom orphan prefix (of different length)

hynek (approved):
thanks!

PR #395, opened 2022-07-16
Use text from debian man pages for CLI help output.

adiroiban:
This tries to fixes #384

The idea is that instead of having a separate `man` file that duplicated the CLI info from the Python script, we should only have the help content in the python script.

From there, we can export it to `man` or `sphinx`.

For this PR I am using the help content from #326

-------

Debian and other distrubtion can run `click-man towncrier` to generate the man content.

click-man has more info for Debian at https://github.com/click-contrib/click-man#debian-packages

---------

The result of this PR

main commnad

```

$ towncrier --help
Usage: towncrier [OPTIONS] COMMAND [ARGS]...

  Towncrier is a utility to produce useful, summarised news files for your
  project. Rather than reading the Git history as some newer tools to produce
  it, or having one single file which developers all write to, towncrier reads
  "news fragments" which contain information useful to end users.

  Towncrier delivers the news which is convenient to those that hear it, not
  those that write it.

  That is, a “news fragment” (a small file containing just enough information
  to be useful to end users) can be written that summarises what has changed
  from the “developer log” (which may contain complex information about the
  original issue, how it was fixed, who authored the fix, and who reviewed the
  fix). By compiling a collection of these fragments, towncrier can produce a
  digest of the changes which is valuable to those who may wish to use the
  software.

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  build*  Build a combined news file from news fragment.
  check   Check for new fragments on a branch.
  create  Create a new news fragment.
```

build command

```
$ towncrier build --help
Usage: towncrier build [OPTIONS]

  Build a combined news file from news fragment.

Options:
  --draft             Render the news fragments to standard output. Don't
                      write to files, don't check versions.
  --config FILE_PATH  Pass a custom config file at FILE_PATH. Default:
                      towncrier.toml or pyproject.toml file, if both files
                      exist, the first will take precedence.
  --dir PATH          Build fragment in directory. Default to current
                      directory.
  --name TEXT         Pass a custom project name.
  --version TEXT      Render the news fragments using given version.
  --date TEXT
  --yes               Do not ask for confirmation to remove news fragments.
  --help              Show this message and exit.
```

check command

```
$ towncrier check --help
Usage: towncrier check [OPTIONS]

  Check for new fragments on a branch.

Options:
  --compare-with BRANCH  Checks files changed running git diff --name-ony
                         BRANCH... BRANCH is the branch to be compared with.
                         Default to origin/master
[...]

PR #470, opened 2023-01-25
Move man page to section 7

### LINKED ISSUE #469: Move master-document manual page to "overview" manual section
The Sphinx configuration currently builds a single manual page, from the entire master document. In `conf.py`:

```python
man_pages = [(master_doc, "towncrier", "Towncrier Documentation", [author], 1)]
```

The resulting manual page does not meet the conventions for section 1 of the manual ("User Commands"). (See the `man-pages(7)` manual page.)

The document should instead be in the manual section 7 ("Overview"), to describe the Towncrier system as a whole.

hynek:
fixes #469

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -133,7 +133,7 @@ latex_documents = [
 
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
-man_pages = [(master_doc, "towncrier", "Towncrier Documentation", [author], 1)]
+man_pages = [(master_doc, "towncrier", "Towncrier Documentation", [author], 7)]
 
 
 # -- Options for Texinfo output -------------------------------------------
--- a/src/towncrier/newsfragments/470.doc
+++ b/src/towncrier/newsfragments/470.doc
@@ -1,0 +1,1 @@
+Moved man page to correct section


exarkun (approved):

adiroiban:
I don't know if this helps.

From what I can see, we don't build the man pages using Sphinx.

And in this PR, this looks like a Sphinx only change.

--------

I don't think that we have any official `man` page version of the doc.

The plan is to use the `click` man page generator helper to generate a man page version of the towncrier CLI tool ... so `click-man towncrier`

Some work here https://github.com/twisted/towncrier/pull/395

--------

So maybe is best to remove the man page info from Sphinx, as that configuration is not used.

hynek:
My presumption here was that distros use it. If we ever have a replacement, sure let's kill it then.

bignose-debian:
> I don't think that we have any official `man` page version of the doc.

The existing documentation is suitable, now that it is moved to manual section 7 to describe the Towncrier system as a whole.

There still is not an official manual page for the 'towncrier' command, suitable for manual section 1. You're correct on that.

> The plan is to use the `click` man page generator helper to generate a man page version of the towncrier CLI tool ... so `click-man towncrier`

I look forward to that being part of the build system.

> So maybe is best to remove the man page info from Sphinx, as that configuration is not used.

I don't see a need to remove this? The two manual pages (one in section 7, one in section 1) serve different purposes, both are useful. See the 'man-pages(7)' manual page for more.

adiroiban:
OK. Makes some sense.

But we don't put any effort into making sure that the Sphinx docs pages are readable as man pages.

I think that if distro want/need a man page, is best to either use click-man to generate the man pages or use a stub page that direct to the webpage.

We now have versioned documentation - https://towncrier.readthedocs.io/en/22.12.0/

So distros can link to older/pin towncrier versions.

bignose-debian:
> But we don't put any effort into making sure that the Sphinx docs pages are readable as man pages.

That's understandable, though disappointing. The Unix manual system is an important part of keeping documentation available in the existing standard system directly on the host, even when offline.

> I think that if distro want/need a man page, is best to either use click-man to generate the man pages

I look forward to this being part of the build system for Towncrier.

> or use a stub page that direct to the webpage.

That would defeat the purpose, since the Unix manual is intended to be documentation directly on the same host. It should be document the topics covered even when offline.

PR #472, opened 2023-01-27
Switch to hatchling

### LINKED ISSUE #467: Build specification should list ‘towncrier.templates’ as a package
When building the package using a PEP 517 build tool such as ‘build’, Setuptools emits this deprecation warning:

> SetuptoolsDeprecationWarning:     Installing 'towncrier.templates' as data is deprecated, please list it in `packages`.
> ############################
> # Package would be ignored #
> ############################
> Python recognizes 'towncrier.templates' as an importable package,
> but it is not listed in the `packages` configuration of setuptools.
> 
> 'towncrier.templates' has been automatically added to the distribution only
> because it may contain data files, but this behavior is likely to change
> in future versions of setuptools (and therefore is considered deprecated).
> 
> Please make sure that 'towncrier.templates' is included as a package by using
> the `packages` configuration field or the proper discovery methods
> (for example by using `find_namespace_packages(...)`/`find_namespace:`
> instead of `find_packages(...)`/`find:`).
> 
> You can read more about "package discovery" and "data files" on setuptools
> documentation page.

Using Python 3.11, ‘build’ version 0.9.0, ‘setuptools’ version 65.6.3.

### LINKED ISSUE #471: Switch to hatchling?
I'm currently three layers[^lay] deep yak-shaving, but given that:

1. [pip / setuptools interaction is broken on 3.12](https://github.com/pypa/setuptools/issues/3661)
2. [Twisted has switched to hatchling](https://github.com/twisted/twisted/blob/trunk/pyproject.toml)

Would ya'll be interested to move towncrier to a static config too? Given the simplicity of our setup.py, it should be a simple switcheroo.

[^lay]: I wanted to simplify our CI according to https://hynek.me/articles/why-i-like-nox/, but we have no pre-release so I wanted to add a pre-release, but just adding 3.12 to the noxfile explodes as explained in 1.

hynek:
As discussed in #471, here's the switch to Hatchling.

I don't see any warnings when building:

```console
$ pipx run build
* Creating venv isolated environment...
* Installing packages in isolated environment... (hatchling ~= 1.12.2, incremental == 22.10.0, wheel ~= 0.38.4)
* Getting build dependencies for sdist...
* Building sdist...
* Building wheel from sdist
* Creating venv isolated environment...
* Installing packages in isolated environment... (hatchling ~= 1.12.2, incremental == 22.10.0, wheel ~= 0.38.4)
* Getting build dependencies for wheel...
* Building wheel...
Successfully built towncrier-22.12.1.dev0.tar.gz and towncrier-22.12.1.dev0-py3-none-any.whl
```

So I surmise it fixes #471 and fixes #467.


src/towncrier/_version.py
```diff
@@ -9,4 +9,6 @@
 
 
 __version__ = Version("towncrier", 22, 12, 1, dev=0)
-__all__ = ["__version__"]
+_hatchling_version = __version__.short()
```
adiroiban, line 14: Maybe add a comment

```suggestion
# The version is exposed in string format to be
# available for the hatching build tools.
_hatchling_version = __version__.short()
```

I think it would be nice to remove the usage of incremental in the public version, and just use a string.

But  I have no idea how to do it in a backward compatible way :(

noxfile.py
```diff
@@ -87,11 +87,10 @@ def docs(session: nox.Session) -> None:
...
-    session.install("build", "check-manifest>=0.44", "twine")
+    session.install("build", "twine")
 
-    session.run("check-manifest", "--verbose")
     # If no argument is passed, build builds an sdist and then a wheel from
     # that sdist.
     session.run("python", "-m", "build")
 
-    session.run("twine", "check", "dist/*")
+    session.run("twine", "check", "--strict", "dist/*")
```
adiroiban, line 90: is this the replacement for `check-manifest` ?

adiroiban (commented):
I saw that check-manifest was running and failing and the other jobs were green.

`twine check --strict` was a pass.

https://github.com/twisted/towncrier/actions/runs/4026949500/jobs/6922085320#step:7:26

Is check_manifest a false positive?

Thanks for the update.

hynek:
Source for the rename: https://docs.readthedocs.io/en/stable/config-file/index.html

noxfile.py
hynek: No, it's just a stricter check for the package (I think including README). We don't have a manifest anymore.

src/towncrier/_version.py
hynek: We can tackle your favorite task next – should be easier now. ;) But yeah it's gonna be breaking either way.

hynek:
> Is check_manifest a false positive?

No, I just forgot to remove it. I'll go thru ci.yml again once this is merged.

hynek:
@adiroiban I have just pushed a special gift for you!

Since we're switching to modern static packaging, I have deprecated `towncrier.__version__` altogether.

I don't think we have to wait for a year to remove it, but in the worst case we can.

---

The two warnings it now raises are nonsense (looks like unittest is iterating over all globals!?) I'm not sure how to get rid of it in trial?

noxfile.py
adiroiban: thanks for the info.

adiroiban (approved):
Many thanks.

I think we can merge this.

Happy to see the removal of manifest and the  deprecation of the `Incremental` version object.

Happy to see the push for `importlib.metadata`

AFAIK this is the future, and in the future towncrier should make use of the same library to get the version for the targeted project.

Thanks again. Nice work!

hynek:
Thanks for the review! We’ll get those cobwebs off Towncrier. :)

Issue #474, opened 2023-02-07
`towncrier create` command doesn't support the (optional) fragment counter


pawelad:
Hi,

I was looking for a "multiple news fragments of the same type linked to one issue" functionality, and I found it through #355 and #119, but unfortunately it doesn't seem to work with the `towncrier create` command:

```console
$ towncrier create 2585.changed.1.md
Usage: towncrier create [OPTIONS] FILENAME
Try 'towncrier create --help' for help.

Error: Invalid value: Expected filename '2585.changed.1.md' to be of format '{name}.{type}', where '{name}' is an arbitrary slug and '{type}' is one of: feature, bugfix, doc, removal, misc
```

It _does_ seem to work when I manually create the file `2585.changed.1.md` though, so it might just be the problem with the command not supporting it?

Also, if this is a supported feature (which I _think_ it is, based on https://github.com/twisted/towncrier/search?q=counter), it would be nice to mention it somewhere in the docs.

I'm happy to help with a PR if needed.

adiroiban:
Thanks for the report.

I think this is an important functionality that is missing.

I have no idea how to fix it in a clean way.

I am using a fork of towncrier in which I have hacked support for something like `2585-1.changed.md`

`.changes.md` should always stay the same as this is the fragment type.


Maybe this can be implemented as `2585.1.changed.md`

In theory for `towncrier` anything before `.changed.md` should be considered the unique fragment/issue ID.

But then, things get complicated when you try to extract the ticket ID to generate a link in the release notes.

---------

One might argue that if 2 fragments are needed, then the PR should be split into 2 separate tickets/PRs... but I disagree :)

So looking forward to see how this can be solved in a smart way :)

SmileyChris:
@adiroiban -- I'm not sure why you need a fork, this is built in functionality: the incremental counter goes *after* the change type.

This works for both `123.feature.1` and `123.feature.1.md`

I remember thinking it was silly that `towncrier create`  didn't do this for you. I got around to submitting a PR so it will 😎

PR #475, opened 2023-02-09
Make `towncrier create` increment fragments rather than failing

SmileyChris:
# Description
```bash
$ ls src/towncrier/newsfragments
459.misc  462.misc  470.doc  472.misc  
$ towncrier create 459.misc
Created news fragment at /home/chris/dev/towncrier/src/towncrier/newsfragments/459.misc.1
$ ls src/towncrier/newsfragments
459.misc  459.misc.1  462.misc  470.doc  472.misc
```

# Checklist


glyph (changes requested):
I don't quite understand what the use-case is here.  If a bug is fixed in a release, it can only be fixed once.  If a feature is added, it can only be added once.

If there *is* a clear reason to do this, it should be `1234-1.bugfix` rather than `1234.bugfix.1`, since `.bugfix` is the extension here (which indicates the file type) and not a `.1` (which would be a man page).

Thanks for taking the time to contribute to towncrier!

hynek:
I think `1234-1.bugfix` is better too, because part of my secret todo list it `123,456.bugfix` so I don’t have to keep multiple news fragments in sync, just to have an entry point to multiple issues. `123-1,456.bugfix` would work because it’s unequivocal.

That said, I don’t see much sense in that, you can just edit the old news fragment or is this about pacifying bots that _require_ a new news fragment?

pawelad:
Isn't this just adding support for an _already implemented_ counter feature (https://github.com/twisted/towncrier/search?q=counter) to the `towncrier create` command?

And FWIW, for us the use case for is having multiple changelog entries (of the same type) linked to one issue. I understand (and agree) that one PR / issue should be about one thing, but unfortunately it isn't always the case for us.
I was pleased to find that `towncrier` already has support for that through the optional counter in the filename, but was then surprised that `towncrier create` doesn't support it, hence #474

src/towncrier/test/test_create.py
```diff
@@ -151,20 +151,45 @@ def test_invalid_section(self):
...
+                "123.feature",
+                "123.feature.1",
+                "123.feature.2",
+            ],
+        )
 
-            self.assertEqual([], os.listdir("foo/newsfragments"))
+    @with_isolated_runner
+    def test_file_exists_with_ext(self, runner: CliRunner):
+        """Ensure we don't overwrite existing files."""
```
adiroiban: this looks like a copy/paste error.

It would help to have a unique docstring that will explain why we have this test, and what is expected from this test.

docs/tutorial.rst
```diff
@@ -90,6 +90,7 @@ You can create those fragments either by hand, or using the ``towncrier create``
 Let's create some example news fragments to demonstrate::
 
    $ echo 'Fixed a thing!' > src/myproject/newsfragments/1234.bugfix
+   $ towncrier create --content 'Fixed thing again!' 1234.bugfix  # since the file existed, this will create a 1234.bugfix.1
```
adiroiban: I think that is best to describe the "counter" functionality in a separate paragraph/section.

This example is already quite heavy and is hard to focus on the core functionality vs extra features.

Thanks

adiroiban:
Thanks @pawelad for the head up.

I can see that the "counter" functionality was added here https://github.com/twisted/towncrier/pull/119

I think that the current team of maintainers is not aware of that functionality, and we would all like not to have the counter as a "file extension".

It looks like the counter feature was added without documentation...but I think that by then, there was no documentation for towncrier.

So, I think that this should be merged with the counter as an extension.

Later, we can see if we can refactor this not to have the counter as an extension.

I think that towncrier also has support for `459.misc.rst` ... if the `counter` feature works as `459.misc.1.rst` I think that we are fine.

SmileyChris:
Ok, I've fixed those minor suggestions for you @adiroiban.

As you've discovered, this is just functionality that towncrier has already, being surfaced more clearly. Hopefully @glyph agrees, I've re-requested him to review.

glyph (approved):
OK. Sounds good. Discussion of improving this feature will go elsewhere.

adiroiban:
Many thanks, Chris for the PR and Glyph for the review.
I was not aware of this functionality in towncrier... you learn something new every day :)

PR #477, opened 2023-02-10
Update documentation about "create -c"

HealthyPear:
# Description

Specify that the use of single quotes to retain RST or Markdow syntax.

# Checklist


glyph (commented):
Can you update the examples to match? It seems confusing that it shows double quotes, but then says you should use single quotes. Also, why not use double quotes? If someone needs to be told that single quotes are better, they're unlikely to know why.

HealthyPear:
> Can you update the examples to match? It seems confusing that it shows double quotes, but then says you should use single quotes. Also, why not use double quotes? If someone needs to be told that single quotes are better, they're unlikely to know why.

Are you referring to the "Creating News Fragments" section?
The examples there using this feature are already correct, that is the point of my PR: to specify why you need to use single quotes, which I did in the `note` box just after the example box.

SmileyChris:
One problem with using single quotes is that "A single quote may not occur between single quotes, even when preceded by a backslash.". So...
```
towncrier create -c 'You can\'t do this!'
```
So technically, double quotes and appropriate escaping is a more complete option.

As an aside, I don't know if this needs a note on the tutorial. Personally I think it'd be fine to mention this only in the cli docs.

exarkun:
> One problem with using single quotes is that "A single quote may not occur between single quotes, even when preceded by a backslash.". So...
> 
> ```
> towncrier create -c 'You can\'t do this!'
> ```
> 
> So technically, double quotes and appropriate escaping is a more complete option.

In POSIX sh this works:

```
towncrier create -c 'You can'\''t not do this!'
```

> 
> As an aside, I don't know if this needs a note on the tutorial. Personally I think it'd be fine to mention this only in the cli docs.

I agree.  Actually, I don't even know if it belongs in the cli docs.  It's entirely up to the shell.  Who is to say the user is even using a POSIX sh at all?  They might be using PowerShell or fish or something.

adiroiban:
Thanks for the feedback. I am closing this as won't fix.

I don't think that this is an towncrier issue.

It's just how various shell works... you can use towncrier on Linux, Unix, Window, with various shells ...and each might handle the arguments in different way

PR #482, opened 2023-02-21
Interactive create

SmileyChris:
# Description
If no filename is given when doing `towncrier create`, interactively ask for the issue number and fragment type (and then launch an interactive editor for the newsfragment content).

```console
$ towncrier create
Issue number (`+` for none): 482
Fragment type (feature, bugfix, doc, removal, misc): feature
($EDITOR opened here)
Created news fragment at /home/chris/dev/towncrier/src/towncrier/newsfragments/482.feature.rst
```

## Reworded interactive edit comment
The comment given to the editor has been reworded to use the similar format as a git commit interactive edit, with the empty space above rather than below:
```
█
# Please write your news content. Lines starting with '#' will be ignored, and
# an empty message aborts.
```

## Default news fragment filename extension
If the file extension of the `filename` configuration setting ends in `.rst` or `.md`, `towncrier create` will append this extension to newsfragments it creates (if the filename didn't provide an explicit extension).
```console
$ towncrier create 1.feature
Created news fragment at /home/chris/dev/towncrier/src/towncrier/newsfragments/1.feature.rst
$ towncrier create 2.feature.txt
Created news fragment at /home/chris/dev/towncrier/src/towncrier/newsfragments/2.feature.txt
```
This is configurable with a new `create_add_extension` setting (`true` by default).

## Add an empty line to the end of news fragments
An empty line is now added to the end of news fragment content, whether entered by the command line option or interactively.
This is configurable with a new `create_eof_newline` setting (`true` by default).

# Checklist

### DIFF
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -48,6 +48,7 @@ class Config:
     wrap: bool
     all_bullets: bool
     orphan_prefix: str
+    create_add_extension: bool = True
 
 
 class ConfigError(Exception):
@@ -181,4 +182,5 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
         wrap=wrap,
         all_bullets=all_bullets,
         orphan_prefix=config.get("orphan_prefix", "+"),
+        create_add_extension=config.get("create_add_extension", True),
     )
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -58,6 +58,7 @@ class Config:
     wrap: bool = False
     all_bullets: bool = True
     orphan_prefix: str = "+"
+    create_add_extension: bool = True
 
 
 class ConfigError(Exception):
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import os
 import sys
-
 from collections import OrderedDict
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Mapping
@@ -13,7 +12,6 @@ from typing import TYPE_CHECKING, Any, Mapping
 import pkg_resources
 
 from .._settings import fragment_types as ft
-
 
 if TYPE_CHECKING:
     # We only use Literal for type-checking and Mypy always brings its own
@@ -48,6 +46,7 @@ class Config:
     wrap: bool
     all_bullets: bool
     orphan_prefix: str
+    create_eof_newline: bool = True
     create_add_extension: bool = True
 
 
@@ -182,5 +181,6 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
         wrap=wrap,
         all_bullets=all_bullets,
         orphan_prefix=config.get("orphan_prefix", "+"),
+        create_eof_newline=config.get("create_eof_newline", True),
         create_add_extension=config.get("create_add_extension", True),
     )
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import os
 import sys
+
 from collections import OrderedDict
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Mapping
@@ -12,6 +13,7 @@ from typing import TYPE_CHECKING, Any, Mapping
 import pkg_resources
 
 from .._settings import fragment_types as ft
+
 
 if TYPE_CHECKING:
     # We only use Literal for type-checking and Mypy always brings its own
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -58,6 +58,7 @@ class Config:
     wrap: bool = False
     all_bullets: bool = True
     orphan_prefix: str = "+"
+    create_eof_newline: bool = True
     create_add_extension: bool = True
 
 
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -45,6 +45,11 @@ DEFAULT_CONTENT = "Add your info here"
     default=DEFAULT_CONTENT,
     help="Sets the content of the new fragment.",
 )
+@click.option(
+    "--eof-newline/--no-eof-newline",
+    default=None,
+    help="Ensure the content ends with an empty line.",
+)
 @click.argument("filename", default="")
 def _main(
     ctx: click.Context,
@@ -53,6 +58,7 @@ def _main(
     filename: str,
     edit: bool | None,
     content: str,
+    eof_newline: bool | None,
 ) -> None:
     """
     Create a new news fragment.
@@ -70,7 +76,7 @@ def _main(
     * .removal - a deprecation or removal of public API,
     * .misc - a ticket has been closed, but it is not of interest to users.
     """
-    __main(ctx, directory, config, filename, edit, content)
+    __main(ctx, directory, config, filename, edit, content, eof_newline)
 
 
 def __main(
@@ -80,11 +86,14 @@ def __main(
     filename: str,
     edit: bool | None,
     content: str,
+    eof_newline: bool | None,
 ) -> None:
     """
     The main entry point.
     """
     base_directory, config = load_config_from_options(directory, config_path)
+    if eof_newline is None:
+        eof_newline = config.create_eof_newline
 
     filename_ext = ""
     if config.create_add_extension:
@@ -165,6 +174,8 @@ def __main(
 
     with open(segment_file, "w") as f:
         f.write(content)
+        if eof_newline and content and not content.endswith("\n"):
+            f.write("\n")
 
     click.echo(f"Created news fragment at {segment_file}")
 
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -14,6 +14,9 @@ import click
 from ._settings import config_option_help, load_config_from_options
 
 
+DEFAULT_CONTENT = "Add your info here"
+
+
 @click.command(name="create")
 @click.pass_context
 @click.option(
@@ -32,30 +35,32 @@ from ._settings import config_option_help, load_config_from_options
 )
 @click.option(
     "--edit/--no-edit",
-    default=False,
+    default=None,
     help="Open an editor for writing the newsfragment content.",
-)  # TODO: default should be true
+)
 @click.option(
     "-c",
     "--content",
     type=str,
-    default="Add your info here",
+    default=DEFAULT_CONTENT,
     help="Sets the content of the new fragment.",
 )
-@click.argument("filename")
+@click.argument("filename", default="")
 def _main(
     ctx: click.Context,
     directory: str | None,
     config: str | None,
     filename: str,
-    edit: bool,
+    edit: bool | None,
     content: str,
 ) -> None:
     """
     Create a new news fragment.
 
-    Create a new news fragment called FILENAME or pass the full path for a file.
-    Towncrier has a few standard types of news fragments, signified by the file extension.
+    If FILENAME is not provided, you'll be prompted to create it.
+
+    Towncrier has a few standard types of news fragments, signified by the file
+    extension.
 
     \b
     These are:
@@ -73,13 +78,29 @@ def __main(
     directory: str | None,
     config_path: str | None,
     filename: str,
-    edit: bool,
+    edit: bool | None,
     content: str,
 ) -> None:
     """
     The main entry point.
     """
     base_directory, config = load_config_from_options(directory, config_path)
+
+    filename_ext = ""
+    if config.create_add_extension:
+        ext = os.path.splitext(config.filename)[1]
+        if ext.lower() in (".rst", ".md"):
+            filename_ext = ext
+
+    if not filename:
+        issue = click.prompt("Issue number")
+        fragment_type = click.prompt(
+            "Fragment type",
+            type=click.Choice(config.types),
+        )
+        filename = f"{issue}.{fragment_type}"
+        if edit is None and content == DEFAULT_CONTENT:
+            edit = True
 
     file_dir, file_basename = os.path.split(filename)
     if config.orphan_prefix and file_basename.startswith(f"{config.orphan_prefix}."):
@@ -91,15 +112,18 @@ def __main(
                 f"{file_basename[len(config.orphan_prefix):]}"
             ),
         )
-    if len(filename.split(".")) < 2 or (
-        filename.split(".")[-1] not in config.types
-        and filename.split(".")[-2] not in config.types
+    filename_parts = filename.split(".")
+    if len(filename_parts) < 2 or (
+        filename_parts[-1] not in config.types
+        and filename_parts[-2] not in config.types
     ):
         raise click.BadParameter(
             "Expected filename '{}' to be of format '{{name}}.{{type}}', "
             "where '{{name}}' is an arbitrary slug and '{{type}}' is "
             "one of: {}".format(filename, ", ".join(config.types))
         )
+    if filename_parts[-1] in config.types and filename_ext:
+        filename += filename_ext
 
     if config.directory:
         fragments_directory = os.path.abspath(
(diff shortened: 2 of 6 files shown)


adiroiban:
Thanks Chris for this nice PR.
I will try to find some time to review this...but I am very busy this need and next week.
I hope someone else will have time to review it.
If someone left unreviewed don't hesitate to ping me and I will to review.

At a first review, my initial comment was to move the configuration changes into a separate PR and focus this PR only on the interactive part.
In this way, it might be easier to merge one part of this PR, without waiting for other part that might need more work/feedback.

But I am also ok to review this together. It's just that it will take more time to review.

Thanks again for your help.

SmileyChris:
> But I am also ok to review this together. It's just that it will take more time to review.

Yes, I threw in a bit more than I should have... but it was all intertwined and it was too messy to provide them as individual PRs. If any area becomes contentious, I'm happy to remove it from this PR.

Preemptively defending the secondary changes:
* **Interactive edit rewording**: this should be pretty straight forward, it just brings things into line with a git interactive commit, and having the empty line first is a nice advantage to allow typing without navigation in most editors.
* **Default news fragment**: this is a completely optional addition, but one that fit nicely with recent discussions on file extensions we were having. Having a `.rst` (or `.md`) extension makes the experience nicer for users who have an editor that styles content based on file type.
* **Adding empty lines to fragments**: functionally it doesn't change anything with generated output, but it is good practice (and without, fragments technically not even technically text files according to IEEE Std 1003.1-2001: A [text file](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_205) consists of a series of [lines](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_205), each of which requires a terminating newline character). A common pre-commit hook adds a newline the end of files.

adiroiban:
Thanks. I will try to take a look at this next week.

adiroiban:
Sorry Chris for this late review.
It was still marked as unread... but went at the bottom of my github notifications list.

If you are still interested in this. I think that as long as the tests are green, this can be merged.

I don't know whey I will have time to test this.

src/towncrier/create.py
```diff
@@ -32,30 +35,38 @@
 )
 @click.option(
     "--edit/--no-edit",
-    default=False,
+    default=None,
```
adiroiban, line 38: Why use the `None` placeholder here ?

src/towncrier/create.py
```diff
@@ -32,30 +35,38 @@
...
+    help="Ensure the content ends with an empty line.",
+)
+@click.argument("filename", default="")
 def _main(
     ctx: click.Context,
     directory: str | None,
     config: str | None,
     filename: str,
-    edit: bool,
+    edit: bool | None,
```
adiroiban, line 54: Why not leave `edit` as always being boolean ?

It looks like we are using `None` here as a placeholder for the default value.

But maybe it's better to have the explicit default value defined in this way.

```suggestion
    edit: bool = True,
```

src/towncrier/create.py
```diff
@@ -32,30 +35,38 @@
...
     "--content",
     type=str,
-    default="Add your info here",
+    default=DEFAULT_CONTENT,
     help="Sets the content of the new fragment.",
 )
-@click.argument("filename")
+@click.option(
+    "--eof-newline/--no-eof-newline",
+    default=None,
```
adiroiban: Can we have the default value defined here?

```suggestion
    default=False,
```

src/towncrier/newsfragments/482.feature.rst
```diff
@@ -0,0 +1 @@
+If no filename is given when doing ``towncrier`` create, interactively ask for the issue number and fragment type (and then launch an interactive editor for the fragment content).
```
adiroiban, line 1: I think that the `--add-extension` is also a new feaure introduced by this PR.

Also, `--eof-neline` is a new feature of this PR.

Should't we have release notes for these features ?

src/towncrier/test/test_create.py
```diff
@@ -92,25 +96,56 @@ def test_content(self):
...
+            content=[content_line],
+            additional_args=["-c", content_line, "--no-eof-newline"],
+            eof_newline=False,
+        )
+
     def test_message_and_edit(self):
         """
         When creating a new message, a initial content can be passed via
         the command line and continue modifying the content by invoking the
         text editor.
```
adiroiban, line 124: Maybe add a comment that the conten is pass `verbatim` and no extra newline or processed is done.

src/towncrier/test/test_create.py
```diff
@@ -151,12 +186,72 @@ def test_invalid_section(self):
...
+        result = runner.invoke(_main, ["123.feature"])
+        self.assertEqual(result.exit_code, 0, result.output)
+
+        fragments = [f.name for f in frag_path.iterdir()]
+        # No '.rst' extension added.
+        self.assertEqual(fragments, ["123.feature.md"])
+
+    @with_isolated_runner
+    def test_odd_filename_extension(self, runner: CliRunner):
+        """Ensure changelog filename extension is not used if not .rst or .md"""
```
adiroiban: `Ensure` is kind of redundand here.
We write tests to make sure all works as expected :)

```suggestion
    def test_no_filename_extension(self, runner: CliRunner):
        """
        When the NEWS filename has no extensions,
        new fragments are created by default without any
        extension.
        """
```

src/towncrier/test/test_create.py
```diff
@@ -151,12 +186,72 @@ def test_invalid_section(self):
...
+        setup_simple_project(extra_config='filename = "the.changes"')
```
adiroiban:
```suggestion
        # The new of the file where towncrier will generate
        # the final release notes is names `RELEASE_NOTES`
        # and for this test, has no file extension.
        setup_simple_project(extra_config='filename = "RELEASE_NOTES"')
```

src/towncrier/test/test_create.py
```diff
@@ -151,12 +186,72 @@ def test_invalid_section(self):
...
+            [
+                "123.feature.1.rst",
+                "123.feature.2.rst",
+                "123.feature.rst",
+            ],
+        )
+
+    @with_isolated_runner
+    def test_file_exists_no_ext(self, runner: CliRunner):
+        """Ensure we don't overwrite existing files."""
```
adiroiban: This looks like the same docstring from `        """Ensure we don't overwrite existing files."""`

I think that this needs a better docstring to explain this test scenarion.

What we are setting up here, what we are doing and what we are expecting.

src/towncrier/test/test_create.py
```diff
@@ -218,16 +414,18 @@ def test_create_orphan_fragment(self, runner: CliRunner):
         self.assertEqual(2, len(fragments))
         change1, change2 = fragments
 
-        self.assertEqual(change1.suffix, ".feature")
+        self.assertEqual(change1.suffix, ".rst")
```
adiroiban, line 413: It looks like the expectation was changed here, without any changes in the test setup.

Is this a change that modified the default behaviour ?

I think that is ok to change the default behaviour, but this needs to be mentioned in the release notes.

This project is all about release notes, so I think it's important for `towncrier` to be a good example of how the release notes should be managed :)

src/towncrier/test/test_create.py
```diff
@@ -245,10 +443,12 @@ def test_create_orphan_fragment_custom_prefix(self, runner: CliRunner):
         self.assertEqual(len(fragments), 1)
         change = fragments[0]
         self.assertTrue(change.stem.startswith("$$$"))
-        # Length should be '$$$' characters and 8 random hex characters.
-        self.assertEqual(len(change.stem), 11)
+        # Length should be '$$$' characters, 8 random hex characters, and ".feature".
+        self.assertEqual(len(change.stem), 3 + 8 + len(".feature"))
```
adiroiban, line 443: Again a change that looks like it's modifying the default behaviour.

adiroiban (changes requested):
Sorry for the late review. and many thanks for the PR. it was easy to follow.

I did a quick code review.

I didn't had time to get the code and run manual tests.

Please see my inline comments and let me know if they make sense.

I think the changes are ok, but it's important to have better release notes for the changes here.

it looks like besides interactive create, this PR also adds 2 new features and breaks at least 1 backward compatibility :)

We can merge this as it is, as long as we document the changes as part of the release notes.

src/towncrier/create.py
SmileyChris: No, since the default of `None` allows falling back to using the configuration value. Setting `True` or `False` allows a user to always override that configuration default.

src/towncrier/create.py
SmileyChris: This is only set to `True` only if the content is still the default content. If a user specifies their own content then, we don't want to default to triggering editing mode. But by leaving it `None`, it makes it possible to provide custom content **and** still explicitly choose to enable editing mode with this flag.

src/towncrier/create.py
SmileyChris: Discussed below

src/towncrier/test/test_create.py
SmileyChris: Now mentioned in release notes

src/towncrier/test/test_create.py
SmileyChris: Now mentioned in release notes

src/towncrier/test/test_create.py
SmileyChris: Because that comment would be incorrect. The newline option is still processed after it's received back from the editor.

SmileyChris: This is specifically testing the case mentioned elsewhere when a user provides custom content but still chooses to edit their content first.

src/towncrier/create.py
adiroiban: We can have it like this... it looks like we are hijacking this variable :)

We should have a separate flag to signal the other use case.

src/towncrier/_settings/load.py
```diff
@@ -52,6 +52,8 @@ class Config:
     wrap: bool = False
     all_bullets: bool = True
     orphan_prefix: str = "+"
+    create_eof_newline: bool = True
+    create_add_extension: bool = True
```
adiroiban, line 56: Should these 2 new config options have the default set to False, for backward compatibility ?

src/towncrier/create.py
```diff
@@ -65,21 +76,44 @@ def _main(
...
+    edit: bool | None,
     content: str,
+    eof_newline: bool | None,
 ) -> None:
     """
     The main entry point.
     """
     base_directory, config = load_config_from_options(directory, config_path)
+    if eof_newline is None:
+        eof_newline = config.create_eof_newline
```
adiroiban: Just trying to understand what is going on here 

```suggestion
    if eof_newline is None:
        # If no explicit flag was provided via CLI
        # default to the value from the configuration file.
        eof_newline = config.create_eof_newline
```

src/towncrier/create.py
adiroiban: Do we need this as a command line option?

Can we just go with the config file ?

src/towncrier/test/test_create.py
adiroiban: OMG :) Are people using this feature ?

I am ok with that, we can have it.

It looks so complicated :) ... but I have never felt the need for `towncrier create`

For me it's always easier to just create the file directly in my text editor, without calling towncrier.

src/towncrier/newsfragments/482.feature.rst
```diff
@@ -0,0 +1,5 @@
...
+
+Now by default, when creating a fragment it will be appended with the ``filename`` option's extension (unless an extension is explicitly provided). For example, ``towncrier create 123.feature`` will create ``news/123.feature.rst``. This can be changed in configuration file by setting `add_extension = false`.
+
+A new line is now added by default to the end of the fragment contents. This can be reverted in the configuration file by setting `add_newline = false` or explicitly set with the ``create`` command line flags ``--eof-newline/--no-eof-newline``.
```
adiroiban: I am ok with this new default value... but maybe it should be placed into a separate section ... as it looks like a backward incompatible change.

src/towncrier/newsfragments/482.feature.rst
```diff
@@ -0,0 +1,5 @@
+If no filename is given when doing ``towncrier`` create, interactively ask for the issue number and fragment type (and then launch an interactive editor for the fragment content).
+
+Now by default, when creating a fragment it will be appended with the ``filename`` option's extension (unless an extension is explicitly provided). For example, ``towncrier create 123.feature`` will create ``news/123.feature.rst``. This can be changed in configuration file by setting `add_extension = false`.
```
adiroiban, line 3: Not sure if this is feature, or a backward compatiblity break :)

adiroiban (commented):
Thanks for the update.

I don't have much time for this reaview...and I don't really understand the need for `towncrier create` ... in general :)

So, I am not the best person to review this.

But we should try go get this merged.

As long as it has docs + tests, it should be ok.

My comments are about:

* backward compatibilty , related to the default values of the new config options
* adding yet another CLI argument to `towncrier create` ... maybe having the value in config file is enough


I hope @hynek or @altendky can take a look and see if we need the extra CLI argument and it it's ok to introduce this backward compatiblity change.

Thanks again

--------

I am using a towncrier version from 2017 ...and I am happy with that. So I don't really care.

I am just trying to maintain this, as part of my release management work for twisted/twisted

src/towncrier/_settings/load.py
SmileyChris: It depends if backwards compatibility is more important than (imo) more sensible defaults. I'll leave that decision to the maintainers to make a call on.

src/towncrier/newsfragments/482.feature.rst
SmileyChris: It's a change in the filename but since the file is parsed by towncrier in exactly the same way, I don't think it's really backwards incompatible

src/towncrier/create.py
SmileyChris: Yeah, it's a pretty obscure setting to ever need to change.

src/towncrier/create.py
SmileyChris: It's the same behaviour as any of the other cli settings that also have a configuration setting. If None, fall back to the configuration, otherwise use the value as provided.

src/towncrier/create.py
SmileyChris: Flag gone now

src/towncrier/newsfragments/482.feature.rst
SmileyChris: As in a separate fragment? I'm not sure where else you mean this should go.

src/towncrier/_settings/load.py
adiroiban: I am not using the create command... so I don't know what to say.

In general, the rule is to keep backward compatibiliy. Any exception should be explicitly raised and accepted on a case by case basis.

We don't have a separate backward compatiblity policy for towncrier ... so I go with the general Twisted org policy https://docs.twisted.org/en/stable/development/compatibility-policy.html

I am happy to have a separate policy for towncrier.
Since towncrie is a dev tool, I think we can be more releaxed.

At the same time, I don't like when a CI pipeline breaks after an update 

I guess that `towncrier create` is not used as part of any automation... so we should be ok.

If nobody else is -1 on this default, I will merge is at it is.

I am happy with the new defaults.

Thanks!

src/towncrier/newsfragments/482.feature.rst
adiroiban: Usualy backward incopatible changes are announced into `482.removal.rst`  This is the "Deprecation and removal" category.

I know that "removal" is a bad name... but this is what we have.

There are a log of bad defaults and bad naming conventions in towncrier :(

bad as in non intuitive

src/towncrier/newsfragments/482.feature.rst
adiroiban: OK. Thanks. Fine by me.

I am not sure that adding release notes in this way, in a single fragment, does what we want.

I remember that towncrier will create one list item per file

On my project, as a hack I have `482-1.feature.rst` as a way to add another item in the notes...but I think that we need better support here

I remember there were other disussions about this

adiroiban (approved):
Thanks.

All good.

The only think blocking this is moving the backward incomaptible change into a `.removal` fragment

I will wait a bit to see if anyone else has anything agains this backward incopatible change and all is ok, this can be merged.

thanks again!

adiroiban:
@SmileyChris I think that you are already in the top 5 contributors of this repo for the last few years, so I consider you one of the de-facto maintainers 


I sent you an invite to a team that has write access to the repo.
This will get you write access and issue triage access. You should be able to add tags and update issues

adiroiban:
I found a bit of time today, so I have triggered the auto-merge.

I am not sure when I will have time to check this.

If anyone is -1 on these changes, we can still revert them before the next public release.

Thanks for all your help!

PR #483, opened 2023-03-01
Default markdown template

SmileyChris:
# Description
Provides a default markdown template.

Towncrier will use this as the default template if the configured `filename` ends with `.md`. For any other filename, `default.rst` will continue to be used. For example:
```toml
filename = "CHANGES.md"
```

This is a simplification of earlier work in #439.

## Changes to `load.py`
In `load.py`, the Config dataclass now has defaults, rather than them being dispersed elsewhere in logic.

The `parse_toml` function has been reorganized for clarity (keeping the same functionality).

## Better configuration docs
The documentation for configuration settings was rewritten for completeness, better accuracy of defaults, and missing features.

It also better highlights different required settings for python vs non-python projects.

## Reference style link support
The only thing different in the markdown default template than the restructured text one is the ability to use markdown reference style links if the config is set to `issue_format = "[{issue}]: https://<bug tracker>/{issue}"`.

For example, with `issue_format = "[{issue}]: https://github.com/twisted/towncrier/issues/{issue}"`:

```markdown
...
### Features

- Foo added. ([2], [9], [72])
- More stuff ([3])
- Stuff! ([4])

[2]: https://github.com/twisted/towncrier/issues/2
[3]: https://github.com/twisted/towncrier/issues/3
[4]: https://github.com/twisted/towncrier/issues/4
[9]: https://github.com/twisted/towncrier/issues/9
[72]: https://github.com/twisted/towncrier/issues/72
```

This is why there's a new `issues_by_category` template context variable built and passed to the template.

# Checklist

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -34,22 +34,24 @@ Top level keys
 ~~~~~~~~~~~~~~
 
 ``name``
-    The name of your project. If empty and the ``package`` key is provided, the name will be automatically determined.
+    The name of your project.
+
+    For Python projects that provide a ``package`` key, if left empty then the name will be automatically determined.
 
     ``""`` by default.
 
 ``version``
     The version of your project.
 
-    Python projects that provide the ``package`` key can have the version to be automatically determined from a ``__version__`` variable in the package's module (or the version if the package is installed).
+    Python projects that provide the ``package`` key can have the version to be automatically determined from a ``__version__`` variable in the package's module.
 
     If not provided or able to be determined, the version must be passed explicitly by the command line argument ``--version``.
 
 ``directory``
     The directory storing your news fragments.
 
-    For Python projects, the default is a ``newsfragments`` directory within the package.
-    For non-Python projects, the default is a ``newsfragments`` directory relative to the configuration file.
+    For Python projects that provide a ``package`` key, the default is a ``newsfragments`` directory within the package.
+    Otherwise the default is a ``newsfragments`` directory relative to the configuration file.
 
 ``filename``
     The filename of your news file.
@@ -121,7 +123,7 @@ Extra top level keys for Python projects
 ``package``
     The Python package name of your project.
 
-    Allows ``version`` to be automatically determined from the Python package version.
+    Allows ``name`` and ``version`` to be automatically determined from the Python package.
     Changes the default ``directory`` to be a ``newsfragments`` directory within this package.
 
 ``package_dir``
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -43,6 +43,9 @@ Top level keys
   For non-Python projects, the default is a ``newsfragments`` directory relative to the configuration file.
 - **``filename``** -- The filename of your news file.
   ``"NEWS.rst"`` by default.
+- **``template``** -- Path to the template for generating the news file.
+  If the path looks like ``<some.package>:<filename.ext>``, it is interpreted as a template bundled with an installed Python package.
+  ``"towncrier:default.rst"`` by default unless ``filename`` ends with ``.md``, in which case the default is ``"towncrier:default.md"``.
 - **``start_string``** -- The magic string that ``towncrier`` looks for when considering where the release notes should start.
   ``".. towncrier release notes start"`` by default.
 - **``title_format``** -- A format string for the title of your project.
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -2,57 +2,116 @@ Configuration Reference
 =======================
 
 ``towncrier`` has many knobs and switches you can use, to customize it to your project's needs.
-The setup in the `Quick Start <quickstart.html>`_ doesn't touch on many, but this document will detail each of these options for you!
+The setup in the `Tutorial <tutorial.html>`_ doesn't touch on many, but this document will detail each of these options for you!
 
 For how to perform common customization tasks, see `Customization <customization/index.html>`_.
 
 ``[tool.towncrier]``
 --------------------
 
-All configuration for ``towncrier`` sits inside ``pyproject.toml``, under the ``tool.towncrier`` namespace.
+All configuration for ``towncrier`` sits inside ``towncrier.toml`` or ``pyproject.toml``, under the ``tool.towncrier`` namespace.
 Please see https://toml.io/ for how to write TOML.
 
+A minimal configuration for a Python project looks like this:
+
+.. code-block:: toml
+
+   # pyproject.toml
+
+   [tool.towncrier]
+   package = "myproject"
+
+A minimal configuration for a non-Python project looks like this:
+
+.. code-block:: toml
+
+   # towncrier.toml
+
+   [tool.towncrier]
+   name = "My Project"
+   version = "1.0.0"
+   directory = "newsfragments" 
 
 Top level keys
 ~~~~~~~~~~~~~~
 
-- ``directory`` -- If you are not storing your news fragments in your Python package, or aren't using Python, this is the path to where your newsfragments will be put.
-- ``filename`` -- The filename of your news file.
-  ``NEWS.rst`` by default.
-- ``package`` -- The package name of your project.
-  (Python projects only)
-- ``package_dir`` -- The folder your package lives. ``./`` by default, some projects might need to use ``src``.
-  (Python projects only)
-- ``template`` -- Path to an alternate template for generating the news file, if you have one.
-- ``start_string`` -- The magic string that ``towncrier`` looks for when considering where the release notes should start.
-  ``.. towncrier release notes start`` by default.
-- ``title_format`` -- A format string for the title of your project.
-  ``{name} {version} ({project_date})`` by default.
-- ``issue_format`` -- A format string for rendering the issue/ticket number in newsfiles.
-  ``#{issue}`` by default.
-- ``underlines`` -- The characters used for underlining headers.
+- **``name``** -- The name of your project. If empty and the ``package`` key is provided, the name will be automatically determined.
+  ``""`` by default.
+- **``version``** -- The version of your project. Mandatory except for Python projects that provide the ``package`` key and want the version to be automatically determined from a ``__version__`` variable in the package's module.
+- **``directory``** -- The directory storing you news fragments. Mandatory except for Python projects (where the default is a ``newsfragments`` directory within the package).
+- **``filename``** -- The filename of your news file.
+  ``"NEWS.rst"`` by default.
+- **``template``** -- Path to the template for generating the news file. If the path starts with ``towncrier:``, it is interpreted as a template bundled with ``towncrier``.
+  ``"towncrier:default.rst"`` by default (unless ``filename`` ends with ``.md``, in which case the default is ``"towncrier:default.md"``).
+- **``start_string``** -- The magic string that ``towncrier`` looks for when considering where the release notes should start.
+  ``".. towncrier release notes start"`` by default.
+- **``title_format``** -- A format string for the title of your project. The explicit value of ``False`` will disable the title entirely. Any other empty value means the template should render the title (the bundled templates use ``<name> <version> (<date>)``). Strings should use the following keys to render the title dynamically: ``{name}``, ``{version}``, and ``{project_date}``.
+  ``""`` by default.
+- **``issue_format``** -- A format string for rendering the issue/ticket number in newsfiles. If none, the issues are rendered as ``#<issue>`` if for issues that are integers, or just ``<issue>`` otherwise. Use the ``{issue}`` key in your string render the issue number, for example Markdown projects may want to use ``"[{issue}]: https://<your bug tracker>/{issue}"``.
+  ``None`` by default.
+- **``underlines``** -- The characters used for underlining headers. Not used in the bundled Markdown template.
   ``["=", "-", "~"]`` by default.
+- **``wrap``** -- Boolean value indicating whether to wrap news fragments to a width of 79.
+  ``false`` by default.
+- **``all_bullets``** -- Boolean value indicating whether the template uses bullets for each news fragment.
+  ``true`` by default.
+- **``single_file``** -- Boolean value indicating whether to write all news fragments to a single file. If false, the ``filename`` should use the following keys to render the filenames dynamically: ``{name}``, ``{version}``, and ``{project_date}``.
+  ``true`` by default.
+- **``orphan_prefix``** -- The prefix used for orphaned news fragments.
+  ``"+"`` by default.
+
+Extra top level keys for Python projects
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **``package``** -- The package name of your project.
+- **``package_dir``** -- The folder your package lives.
+  ``"."`` by default, some projects might need to use ``"src"``.
+
+
+Sections
+--------
+
+``towncrier`` supports splitting fragments into multiple sections, each with its own news of fragment types.
+
+Add an array of tables your ``.toml`` configuration file named **``[[tool.towncrier.section]]``**.
+
+Each table within this array has the following mandatory keys:
+
+- **``name``** -- The name of the section.
+- **``path``** -- The path to the directory containing the news fragments for this section, relative to the configured ``directory``. Use ``""`` for the root directory.
+
+For example:
+
+.. code-block:: toml
+
+   [[tool.towncrier.section]]
+   name = "Main Platform"
+   path = ""
+
+   [[tool.towncrier.section]]
+   name = "Secondary"
+   path = "secondary"
 
 
 Custom fragment types
 ---------------------
-``towncrier`` allows defining custom fragment types.
-Custom fragment types will be used instead ``towncrier`` default ones, they are not combined.
 
-There are two ways to add custom fragment types.
+``towncrier`` hase the following default fragment types: ``feature``, ``bugfix``, ``doc``, ``removal``, and ``misc``.
+
+You can use either of the two following method to define custom types instead (you will need to redefine any of the default types you want to use).
 
 
-Defining Custom Fragment Types With a TOML Mapping
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Use TOML tables (alphabetical order)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Users can configure each of their own custom fragment types by adding tables to
-the pyproject.toml named ``[tool.towncrier.fragment.<a custom fragment type>]``.
+Adding tables to your ``.toml`` configuration file named **``[tool.towncrier.fragment.<a custom fragment type>]``**.
 
-These tables may include the following optional keys:
+These may include the following optional keys:
 
- * ``name``: The description of the fragment type, as it must be included in the news file.
-   If omitted, it defaults to  its  fragment type, but capitalized.
- * ``showcontent``: Whether if the fragment contents should be included in the news file. If omitted, it defaults to ``true``
+- **``name``** -- The description of the fragment type, as it must be included in the news file.
+  Defaults to its fragment type, but capitalized.
+- **``showcontent``** -- A boolean value indicating whether the fragment contents should be included in the news file.
+  ``true`` by default.
 
 For example, if you want your custom fragment types to be ``["feat", "fix", "chore",]`` and you want all of them to use the default configuration except ``"chore"`` you can do it as follows:
 
@@ -70,25 +129,23 @@ For example, if you want your custom fragment types to be ``["feat", "fix", "cho
 
 .. warning::
 
-   Since TOML mappings aren't ordered, the sections are always rendered alphabetically.
+   Since TOML mappings aren't ordered, types defined using this method are always rendered alphabetically.
 
 
-Defining Custom Fragment Types With an Array of TOML Tables
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Use a TOML Array (defined order)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Users can create their own custom fragment types by adding an array of
-tables to the pyproject.toml named ``[[tool.towncrier.type]]``.
+Add an array of tables to your ``.toml`` configuration file named **``[[tool.towncrier.type]]``**.
 
-If you use this way to configure custom fragment types, please note that ``fragment_types`` must be empty or not provided.
+If you use this way to configure custom fragment types, ensure there is no ``tool.towncrier.fragment`` table.
 
-Each custom type (``[[tool.towncrier.type]]``) has the following
-mandatory keys:
+Each table within this array has the following mandatory keys:
 
-* ``directory``: The type / category of the fragment.
-* ``name``: The description of the fragment type, as it must be included
+- **``directory``** -- The type / category of the fragment.
+- **``name``** -- The description of the fragment type, as it must be included
   in the news file.
-* ``showcontent``: Whether if the fragment contents should be included in the
-  news file.
+- **``showcontent``** -- A boolean value indicating whether the fragment contents should be included in the news file.
+  ``true`` by default.
 
 For example:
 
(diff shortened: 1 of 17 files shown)


docs/tutorial.rst
```diff
@@ -1,7 +1,7 @@
 Tutorial
 ========
 
-This tutorial assumes you have a Python project with a *reStructuredText* (rst) news file (also known as changelog) file that you wish to use ``towncrier`` on, to generate its news file.
+This tutorial assumes you have a Python project with a *reStructuredText* (rst) or *markdown* (md) news file (also known as changelog) file that you wish to use ``towncrier`` on, to generate its news file.
```
hynek:
```suggestion
This tutorial assumes you have a Python project with a *reStructuredText* (rst) or *Markdown* (md) news file (also known as changelog) that you wish to use ``towncrier`` on, to generate its news file.
```

Let's capitalize Markdown, the double file is from me, but doesn't have to stick around. ;)

docs/tutorial.rst
```diff
@@ -144,6 +144,8 @@ You should get an output similar to this::
 
    - #1, #2
 
+Note: if you configure a markdown file (for example, ``filename = "CHANGES.md"`` in your configuration file), the titles will be output in markdown format instead.
```
hynek:
```suggestion
Note: if you configure a Markdown file (for example, ``filename = "CHANGES.md"``) in your configuration file, the titles will be output in markdown format instead.
```

I think it reads better this way.

Genuine question tho: is "to output" really a proper verb??

src/towncrier/_settings/load.py
```diff
@@ -29,25 +28,25 @@
...
+    package_dir: str = "."
+    single_file: bool = True
+    filename: str = "NEWS.rst"
+    directory: str | None = None
+    version: str | None = None
+    name: str = ""
+    start_string: str = ".. towncrier release notes start\n"
+    title_format: str | Literal[False] = ""
+    issue_format: str | None = None
+    underlines: list[str] = dataclasses.field(default_factory=lambda: ["=", "-", "~"])
```
hynek: Would it be a lot of work to change them to a tuple? It being a mutable list kinda makes everything worse.

src/towncrier/newsfragments/483.feature.rst
```diff
@@ -0,0 +1,3 @@
+Provide a default markdown template if the configured filename ends with `.md`.
```
hynek:
```suggestion
Provide a default Markdown template if the configured filename ends with `.md`.
```

src/towncrier/newsfragments/483.feature.rst
```diff
@@ -0,0 +1,3 @@
...
+
+The markdown template uses the same rendered format as the default ReStructuredText template, but with a markdown syntax.
```
hynek:
```suggestion
The Markdown template uses the same rendered format as the default _reStructuredText_ template, but with a Markdown syntax.
```

src/towncrier/test/test_format.py
```diff
@@ -179,6 +181,158 @@ def test_basic(self):
         )
         self.assertEqual(output, expected_output_weird_underlines)
 
+    def test_markdown(self):
+        """
+        Check formating of default markdown template.
+        """
+        fragments = OrderedDict(
```
hynek: Any reason for OrderedDicts here?

src/towncrier/test/test_settings.py
```diff
@@ -35,6 +37,53 @@ def test_base(self):
...
+        )
+        self.assertEqual(config.template, expected_template)
+
+    def test_explicit_template_extension(self):
+        """
+        If the filename references an .md file and the builtin template has an
+        extension, don't change it.
+        """
+        temp = self.mktemp()
+        os.makedirs(temp)
```
hynek: According to Twisted docs `mktemp` creates the dir, so this line shouldn't be necessary?

```suggestion
```

src/towncrier/_settings/load.py
```diff
@@ -107,52 +100,70 @@ def load_config_from_file(directory: str, config_file: str) -> Config:
...
+            )
 
-    sections = OrderedDict()
+    # Process options.
+    for field in dataclasses.fields(Config):
+        if field.name in ("sections", "types", "template"):
+            # Skip these options, they are processed later.
+            continue
+        if field.name in config:
+            #  Interestingly, the __future__ annotation turns the type into a string.
```
hynek, line 141: ah yeah, attrs has https://www.attrs.org/en/latest/api.html#attrs.resolve_types for that. I thought DCs had something too…

src/towncrier/_settings/load.py
```diff
@@ -162,23 +173,7 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
...
-        start_string=config.get("start_string", _start_string),
-        title_format=config.get("title_format", _title_format),
-        issue_format=config.get("issue_format"),
-        underlines=config.get("underlines", _underlines),
-        wrap=wrap,
-        all_bullets=all_bullets,
-        orphan_prefix=config.get("orphan_prefix", "+"),
-    )
+    # Return the parsed config.
+    return Config(**parsed_data)
```
hynek, line 198: I'm gonna be honest and say I'm not a fan of this function, but I guess to make it nicer, we'd have to move to something like jsonschema so I'm not gonna hold this PR over it. It's difficult to gauge it's correctness, due to many changes, but I'm gonna trust the tests…

src/towncrier/test/test_settings.py
```diff
@@ -35,6 +37,53 @@ def test_base(self):
         self.assertEqual(config.underlines, ["=", "-", "~"])
         self.assertEqual(config.orphan_prefix, "~")
 
+    def test_markdown(self):
+        """
+        If the filename references an .md file and the builtin template doesn't have an
+        extension, add .md rather than .rst.
+        """
+        temp = self.mktemp()
+        os.makedirs(temp)
```
hynek: Same as below, I believe this is not necessary?


```suggestion
```

src/towncrier/test/test_settings.py
```diff
@@ -35,6 +37,53 @@ def test_base(self):
...
+            f.write(
+                """[tool.towncrier]
+package = "foobar"
+filename = "NEWS.md"
+template = "towncrier:default.rst"
+"""
+            )
+
+        config = load_config(temp)
+        self.assertEqual(config.filename, "NEWS.md")
```
hynek: let's separate act and assert.

```suggestion
        config = load_config(temp)

        self.assertEqual(config.filename, "NEWS.md")
```

src/towncrier/test/test_settings.py
```diff
@@ -35,6 +37,53 @@ def test_base(self):
...
+filename = "NEWS.md"
+"""
+            )
+
+        config = load_config(temp)
+        self.assertEqual(config.filename, "NEWS.md")
+        expected_template = pkg_resources.resource_filename(
+            "towncrier", "templates/default.md"
+        )
+        self.assertEqual(config.template, expected_template)
```
hynek: let's separate act and assert.

```suggestion
        )

        self.assertEqual(config.template, expected_template)
```

src/towncrier/test/test_settings.py
```diff
@@ -35,6 +37,53 @@ def test_base(self):
...
+        with open(os.path.join(temp, "pyproject.toml"), "w") as f:
+            f.write(
+                """[tool.towncrier]
+package = "foobar"
+filename = "NEWS.md"
+"""
+            )
+
+        config = load_config(temp)
+        self.assertEqual(config.filename, "NEWS.md")
```
hynek: let's separate act and assert.

```suggestion
        config = load_config(temp)

        self.assertEqual(config.filename, "NEWS.md")
```

src/towncrier/test/test_format.py
```diff
@@ -179,6 +181,158 @@ def test_basic(self):
...
+        output = render_fragments(
+            template,
+            "[{issue}]: https://github.com/twisted/towncrier/issues/{issue}",
+            fragments,
+            definitions,
+            ["-", "~"],
+            wrap=True,
+            versiondata={"name": "MyProject", "version": "1.0", "date": "never"},
+        )
+        self.assertEqual(output, expected_output)
```
hynek: let's separate act and assert.

```suggestion
        )

        self.assertEqual(output, expected_output)
```

hynek (changes requested):
oof this is a big boy!

I'm really a bit anxious about it (not by your fault mind you, the code is what it is), but I guess we should trust the tests?

src/towncrier/_settings/load.py
SmileyChris: I mean, we *could*. It's technically an unrelated change to make it immutable, but we can slip it in if you like :P

src/towncrier/_settings/load.py
SmileyChris: Yeah, sorry -- I know it's a big change but it is at least in a more logical and readable order now. Also, it's far better that we're using Config defaults more correctly.

src/towncrier/test/test_format.py
SmileyChris: No, totally unnecessary. was probably copy pasting from other tests.

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+   path = "secondary"
 
 
 Custom fragment types
 ---------------------
-``towncrier`` allows defining custom fragment types.
-Custom fragment types will be used instead ``towncrier`` default ones, they are not combined.
 
-There are two ways to add custom fragment types.
+``towncrier`` hase the following default fragment types: ``feature``, ``bugfix``, ``doc``, ``removal``, and ``misc``.
```
hynek:
```suggestion
``towncrier`` has the following default fragment types: ``feature``, ``bugfix``, ``doc``, ``removal``, and ``misc``.
```

src/towncrier/newsfragments/483.feature.rst
```diff
@@ -0,0 +1,3 @@
+Provide a default Markdown template if the configured filename ends with `.md`.
+
+The Markdown template uses the same rendered format as the default _reStructuredText_ template, but with a Markdown syntax.
```
hynek:
```suggestion
The Markdown template uses the same rendered format as the default *reStructuredText* template, but with a Markdown syntax.
```

_cough rST_

---

Also towncrier doesn't use `rst` suffixes for news fragments. Not a fan myself, but it is what it is.

src/towncrier/test/test_format.py
hynek: you've left them in 😇

src/towncrier/test/test_settings.py
hynek: ping :)

src/towncrier/test/test_settings.py
hynek: ping :)

src/towncrier/_settings/load.py
hynek: It's strictly optional if it's not too much churn / work.

docs/tutorial.rst
```diff
@@ -144,6 +144,8 @@ You should get an output similar to this::
 
    - #1, #2
 
+Note: if you configure a Markdown file (for example, ``filename = "CHANGES.md"``) in your configuration file, the titles will be output in markdown format instead.
```
hynek:
```suggestion
Note: if you configure a Markdown file (for example, ``filename = "CHANGES.md"``) in your configuration file, the titles will be output in Markdown format instead.
```

caught one more

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+  ``".. towncrier release notes start"`` by default.
+- **``title_format``** -- A format string for the title of your project. The explicit value of ``False`` will disable the title entirely. Any other empty value means the template should render the title (the bundled templates use ``<name> <version> (<date>)``). Strings should use the following keys to render the title dynamically: ``{name}``, ``{version}``, and ``{project_date}``.
+  ``""`` by default.
+- **``issue_format``** -- A format string for rendering the issue/ticket number in newsfiles. If none, the issues are rendered as ``#<issue>`` if for issues that are integers, or just ``<issue>`` otherwise. Use the ``{issue}`` key in your string render the issue number, for example Markdown projects may want to use ``"[{issue}]: https://<your bug tracker>/{issue}"``.
+  ``None`` by default.
+- **``underlines``** -- The characters used for underlining headers. Not used in the bundled Markdown template.
   ``["=", "-", "~"]`` by default.
+- **``wrap``** -- Boolean value indicating whether to wrap news fragments to a width of 79.
```
hynek:
```suggestion
- **``wrap``** -- Boolean value indicating whether to wrap news fragments to a line length of 79.
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+
+A minimal configuration for a non-Python project looks like this:
+
+.. code-block:: toml
+
+   # towncrier.toml
+
+   [tool.towncrier]
+   name = "My Project"
+   version = "1.0.0"
```
hynek: Maybe worth it to mention that the version is optional & can be passed on command line? To me it doesn't make much sense to keep the project version separately in a config file.

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
 =======================
 
 ``towncrier`` has many knobs and switches you can use, to customize it to your project's needs.
-The setup in the `Quick Start <quickstart.html>`_ doesn't touch on many, but this document will detail each of these options for you!
+The setup in the `Tutorial <tutorial.html>`_ doesn't touch on many, but this document will detail each of these options for you!
```
hynek: the hell are these links with html suffixes – mind changing them to proper Sphinx links? In this case it should be enough to use
```rst
:doc:`tutorial`
```

With customization below you may want to add an explicit name if you want to, but I think

```rst
:doc:`customization`
```

should be fine

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
-- ``start_string`` -- The magic string that ``towncrier`` looks for when considering where the release notes should start.
-  ``.. towncrier release notes start`` by default.
-- ``title_format`` -- A format string for the title of your project.
-  ``{name} {version} ({project_date})`` by default.
-- ``issue_format`` -- A format string for rendering the issue/ticket number in newsfiles.
-  ``#{issue}`` by default.
-- ``underlines`` -- The characters used for underlining headers.
+- **``name``** -- The name of your project. If empty and the ``package`` key is provided, the name will be automatically determined.
+  ``""`` by default.
+- **``version``** -- The version of your project. Mandatory except for Python projects that provide the ``package`` key and want the version to be automatically determined from a ``__version__`` variable in the package's module.
```
hynek: It's not really mandatory – you just have to pass the version explicitly on the command-line as [we do in attrs](https://github.com/python-attrs/attrs/blob/56cbf826df6adf04dd176fdd5391cf8829a14aa5/tox.ini#L52) for instance.

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+- **``directory``** -- The directory storing you news fragments. Mandatory except for Python projects (where the default is a ``newsfragments`` directory within the package).
```
hynek:
```suggestion
- **``directory``** -- The directory storing your news fragments. Mandatory except for Python projects (where the default is a ``newsfragments`` directory within the package).
```

docs/configuration.rst
hynek:
```suggestion
- **``version``** -- The version of your project.
  Python projects that provide the ``package`` key can have the version to be automatically determined from a ``__version__`` variable in the package's module.
  The version can also be passed explicitly by command line argument ``--version``.
```

docs/configuration.rst
hynek:
```suggestion
- **``directory``** -- The directory storing you news fragments.
  Mandatory except for Python projects (where the default is a ``newsfragments`` directory within the package).
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+- **``filename``** -- The filename of your news file.
+  ``"NEWS.rst"`` by default.
+- **``template``** -- Path to the template for generating the news file. If the path starts with ``towncrier:``, it is interpreted as a template bundled with ``towncrier``.
```
hynek:
```suggestion
- **``template``** -- Path to the template for generating the news file.
  If the path starts with ``towncrier:``, it is interpreted as a template bundled with ``towncrier``.
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+  ``"towncrier:default.rst"`` by default (unless ``filename`` ends with ``.md``, in which case the default is ``"towncrier:default.md"``).
+- **``start_string``** -- The magic string that ``towncrier`` looks for when considering where the release notes should start.
+  ``".. towncrier release notes start"`` by default.
+- **``title_format``** -- A format string for the title of your project. The explicit value of ``False`` will disable the title entirely. Any other empty value means the template should render the title (the bundled templates use ``<name> <version> (<date>)``). Strings should use the following keys to render the title dynamically: ``{name}``, ``{version}``, and ``{project_date}``.
```
hynek:
```suggestion
- **``title_format``** -- A format string for the title of your project.
  The explicit value of ``False`` will disable the title entirely.
  Any other empty value means the template should render the title (the bundled templates use ``<name> <version> (<date>)``).
  Strings should use the following keys to render the title dynamically: ``{name}``, ``{version}``, and ``{project_date}``.
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+  ``""`` by default.
+- **``issue_format``** -- A format string for rendering the issue/ticket number in newsfiles. If none, the issues are rendered as ``#<issue>`` if for issues that are integers, or just ``<issue>`` otherwise. Use the ``{issue}`` key in your string render the issue number, for example Markdown projects may want to use ``"[{issue}]: https://<your bug tracker>/{issue}"``.
```
hynek:
```suggestion
- **``issue_format``** -- A format string for rendering the issue/ticket number in newsfiles.
  If none, the issues are rendered as ``#<issue>`` if for issues that are integers, or just ``<issue>`` otherwise.
  Use the ``{issue}`` key in your string render the issue number, for example Markdown projects may want to use ``"[{issue}]: https://<your bug tracker>/{issue}"``.
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+  ``None`` by default.
+- **``underlines``** -- The characters used for underlining headers. Not used in the bundled Markdown template.
```
hynek:
```suggestion
- **``underlines``** -- The characters used for underlining headers.
  Not used in the bundled Markdown template.
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
   ``["=", "-", "~"]`` by default.
+- **``wrap``** -- Boolean value indicating whether to wrap news fragments to a width of 79.
+  ``false`` by default.
+- **``all_bullets``** -- Boolean value indicating whether the template uses bullets for each news fragment.
+  ``true`` by default.
+- **``single_file``** -- Boolean value indicating whether to write all news fragments to a single file. If false, the ``filename`` should use the following keys to render the filenames dynamically: ``{name}``, ``{version}``, and ``{project_date}``.
```
hynek:
```suggestion
- **``single_file``** -- Boolean value indicating whether to write all news fragments to a single file.
  If false, the ``filename`` should use the following keys to render the filenames dynamically:
  ``{name}``, ``{version}``, and ``{project_date}``.
```

docs/configuration.rst
```diff
@@ -2,57 +2,116 @@ Configuration Reference
...
+--------
+
+``towncrier`` supports splitting fragments into multiple sections, each with its own news of fragment types.
+
+Add an array of tables your ``.toml`` configuration file named **``[[tool.towncrier.section]]``**.
+
+Each table within this array has the following mandatory keys:
+
+- **``name``** -- The name of the section.
+- **``path``** -- The path to the directory containing the news fragments for this section, relative to the configured ``directory``. Use ``""`` for the root directory.
```
hynek:
```suggestion
- **``path``** -- The path to the directory containing the news fragments for this section, relative to the configured ``directory``.
  Use ``""`` for the root directory.
```

hynek (changes requested):
Greetings from the PyCon US sprints!

I've realized that we never documented the fact we use [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) so I've opened #500

docs/configuration.rst
SmileyChris: This says "minimal configuration", so I'll just take it out :P

src/towncrier/test/test_settings.py
SmileyChris: Ok, I removed this from the entire codebase for consistency

src/towncrier/test/test_format.py
SmileyChris: Fine, I purged OrderedDict from the entire codebase 💣

src/towncrier/_builder.py
```diff
@@ -66,15 +66,15 @@ def parse_newfragment_basename(
 
 # Returns a structure like:
 #
-# OrderedDict([
+# {[
```
hynek: The instantiation of OrderedDicts works a bit different than those of normal dicts.

What you've described here (and used in the tests) is creating a set with a list inside (which doesn't work).

The transformation is a bit tedious (you have to transform `[("foo", 42)]` into `{"foo": 42}`) so feel free to roll back that changes if you don't feel like it.

hynek (changes requested):
I've fixed the doc link that I suggested incorrectly, but for some reason CI didn't run. It fails tho due to the ordereddict stuff i mention inline, so I've reverted that commit for now and it passes for me locally.

Either way you'll have to resolve the conflicts that have accumulated. :-/ (on the bright side the revert simplified the conflict somewhat)

src/towncrier/test/test_format.py
```diff
@@ -179,6 +181,159 @@ def test_basic(self):
...
+
+- Multi-line bulleted
+  - fix
+  - here
+
+  (#2)
+- Web fixed. (#3)
+"""
+
+        template = pkg_resources.resource_string(
```
hynek: Ah this needs to use the helper function introduced in #496

src/towncrier/newsfragments/483.feature
```diff
@@ -0,0 +1,3 @@
+Provide a default Markdown template if the configured filename ends with `.md`.
```
hynek:
```suggestion
Provide a default Markdown template if the configured filename ends with ``.md``.
```

rST 🙈

src/towncrier/_settings/load.py
```diff
@@ -29,25 +28,25 @@
...
-    all_bullets: bool
-    orphan_prefix: str
+    package: str = ""
+    package_dir: str = "."
+    single_file: bool = True
+    filename: str = "NEWS.rst"
+    directory: str | None = None
+    version: str | None = None
+    name: str = ""
+    start_string: str = ".. towncrier release notes start\n"
```
hynek: I have just realized, that this makes no sense with Markdown (you'd have to use an HTML comment like `<!-- towncrier release notes start -->\n`) and that we don't have a test for the case where an existing news file is amended – correct?

If things are supposed to work out of the box, I guess you'll have to special-case this too (and add a test for it).

SmileyChris:
Oh windows, why are you being a jerk?

SmileyChris:
> I have just realized, that this makes no sense with Markdown (you'd have to use an HTML comment like `<!-- towncrier release notes start -->\n`) and that we don't have a test for the case where an existing news file is amended – correct?

If things are supposed to work out of the box, I guess you'll have to special-case this too (and add a test for it). 

Ok, special-cased and added tests.

docs/customization/newsfile.rst
```diff
@@ -25,3 +25,10 @@ In an existing news file, it'll look something like this::
   - Fixed, etc...
 
 ``towncrier`` will not alter content above the comment.
+
+Markdown
+~~~~~~~~
+
+If your news file is in Markdown (e.g. ``NEWS.md``), use the following comment instead::
+
+  <!--- towncrier release notes start -->
```
hynek: If you don't explicitly add a type, it's interpreted as Python, but it's HTML: https://towncrier--483.org.readthedocs.build/en/483/customization/newsfile.html

`.. code-block:: html`

Bonus points for adding the other code blocks to `rat`. 🤓

src/towncrier/test/test_build.py
```diff
@@ -1163,6 +1163,107 @@ def test_start_string(self):
...
+    @with_isolated_runner
+    def test_default_start_string(self, runner):
+        """
+        The default start string is ``.. towncrier release notes start``.
+        """
+        setup_simple_project()
+
+        with open("foo/newsfragments/123.feature", "w") as f:
+            f.write("Adds levitation")
+        with open("NEWS.rst", "w") as f:
```
hynek: Please use helpers.read/write in new code.

src/towncrier/test/test_build.py
```diff
@@ -1163,6 +1163,107 @@ def test_start_string(self):
...
+    def test_default_start_string_markdown(self, runner):
+        """
+        The default start string is ``<!-- towncrier release notes start -->`` for
+        Markdown.
+        """
+        setup_simple_project(extra_config='filename = "NEWS.md"')
+
+        with open("foo/newsfragments/123.feature", "w") as f:
+            f.write("Adds levitation")
+        with open("NEWS.md", "w") as f:
```
hynek: same

hynek (changes requested):
this is not an exhaustive review, because i'm on a road trip in the middle of nowhere and being dragged off but i'm trying to keep things moving 😅

hynek (approved):
OK I think this should be good to merge, I don't think another round of a full review by me would do much good and we'd just accumulate more conflicts.

It should probably be like three PRs, but let's fix problems in post. 😅

PR #485, opened 2023-03-11
ci: add 'workflow_dispatch' trigger

umarcor:
# Description

See https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatch.

# Checklist

### DIFF
--- a/src/towncrier/newsfragments/485.misc
+++ b/src/towncrier/newsfragments/485.misc
@@ -1,0 +1,1 @@
+Add workflow_dispatch CI trigger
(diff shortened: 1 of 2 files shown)


src/towncrier/newsfragments/485.misc
```diff
@@ -0,0 +1 @@
+Add workflow_dispatch CI trigger
```
hynek:
```suggestion
```

Misc like these don't need a message (and I believe they aren't rendered either).

hynek (approved):
I wasn't sure we needed this, but I might've _just_ run into a use-case. thanks!

PR #486, opened 2023-03-11
build: if draft, return early

umarcor:
# Description

This is a small change to make the source easier to read. The condition that evaluates `draft` can return early, because there is no content after the else branch. As a result, ~25 lines can be unindented.

# Checklist

### DIFF
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -245,38 +245,39 @@ def __main(
             err=to_err,
         )
         click.echo(content)
-    else:
-        click.echo("Writing to newsfile...", err=to_err)
-        news_file = config.filename
+        return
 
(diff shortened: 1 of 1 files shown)


hynek (changes requested):
The changes look fine, but it needs a newsfragment.

Running `towncrier create -c "" 486.misc` should be fine.

Thanks!

umarcor:
@hynek, done!

hynek:
I've just noticed that #483 is touching the same lines, so I'm gonna hold this back, until it lands soon, because the conflicts should be much easier to fix here than there. (if it gets protracted, we can move this on forward).

hynek (approved):
OK looks like Chris is busy, let's get over with this.

PR #487, opened 2023-03-11
build: do not echo 'Removing...' if the files are to be kept

umarcor:
# Description

Message "Removing news fragments..." is printed before knowing whether the fragments are to be removed.  As a result, when `--keep` is used, the following log is shown:

```
# https://github.com/VUnit/vunit/actions/runs/4390933073/jobs/7689738665#step:5:37
Staging newsfile...
Removing news fragments...
Keeping the following files:
/home/runner/work/vunit/vunit/docs/news.d/921.misc.rst
```

This PR moved the echo command after the condition es evaluated, so the result is:

```
https://github.com/dbhi/vunit/actions/runs/4393378969/jobs/7693737661#step:5:36
Staging newsfile...
Keeping the following files:
/home/runner/work/vunit/vunit/docs/news.d/921.misc.rst
Done!
```

# Checklist


src/towncrier/build.py
```diff
@@ -263,12 +263,12 @@ def __main(
         click.echo("Staging newsfile...", err=to_err)
         _git.stage_newsfile(base_directory, news_file)
 
-        click.echo("Removing news fragments...", err=to_err)
         if should_remove_fragment_files(
```
SmileyChris, line 266: A good additional fix would be changing this line
```suggestion
        if fragment_filenames and should_remove_fragment_files(
```
Currently you get the following process if there are no files to remove:
```
Removing news fragments...
Removing the following files:
Is it okay if I remove those files? [Y/n]
```

SmileyChris (changes requested):

src/towncrier/build.py
umarcor: Nice! I moved the check inside `should_remove_fragment_files`, instead of doing it outside.

src/towncrier/build.py
```diff
@@ -279,6 +279,9 @@ def should_remove_fragment_files(
     answer_yes: bool,
     answer_keep: bool,
 ) -> bool:
+    if not fragment_filenames:
+        click.echo("No news fragments to remove. Skipping...")
```
SmileyChris: Heh, this was my first take too and then I told you the other way because it seemed simpler. This is fine though.
If I was nit-picking, I wouldn't finish with '...' as that is used everywhere else in the process to indicate something is about to be done.

SmileyChris (approved):

src/towncrier/build.py
umarcor: Fair enough! Replaced `...` with `!` and force-pushed.

hynek (approved):
lgtm, thanks!

PR #491, opened 2023-03-27
Remove incremental

### LINKED ISSUE #308: Remove dependency on incremental in favor of general compatible solution

SmileyChris:
# Description
Remove Incremental as a requirement (but still support its use for project versions/names)

Fixes #308 and #398

# Checklist


src/towncrier/test/test_project.py
```diff
@@ -48,25 +41,61 @@ def test_tuple(self):
...
-            DeprecationWarning, "Accessing towncrier.__version__ is deprecated.*"
-        ):
-            name = get_project_name(pkg, "towncrier")
 
-        self.assertEqual(metadata_version("towncrier"), version)
-        self.assertEqual("towncrier", name)
+__version__ = WeirdVersion()
+"""
+            )
+        with self.assertRaises(Exception) as e:
```
chrysle, line 90: Which exception does this raise?

SmileyChris: Literally [`Exception`](https://github.com/twisted/towncrier/pull/491/files#diff-ea234c597bb51dfad3d154ebdeb24af6c75e95ac9a922267f99e7704f1fbde42R72) -- the definition of that exception is existing code, but we can of course tighten this up if it makes sense now.

adiroiban:
Thanks, Chris for working on this.

I am working to make a new release of towncrier, and the plan is to drop 3.7 support.

So this PR should be even simpler.

I am happy to drop the `incremental` dependency from towncrier.
I think that managing it via a simple string is good enough.
The version should only be modified by the release manager and the release manager should be familiar with the versioning scheme.

I think that using `Version('towncrier', 19, 9, 0, release_candidate=1)` instead of just `19.9.0.rc1` is overengineering

SmileyChris:
@adiroiban cool! I'd say merge this first then work on dropping the 3.7 support, otherwise I can update this after you've changed the automated tests so this doesn't fail when I remove the import try/excepts.

(also a review of #482 would be nice, it would be a good addition before the new release ;) )

src/towncrier/_shell.py
```diff
@@ -13,14 +13,25 @@
 
 from click_default_group import DefaultGroup
 
-from ._version import __version__
 from .build import _main as _build_cmd
 from .check import _main as _check_cmd
 from .create import _main as _create_cmd
 
 
+try:
```
adiroiban, line 23: Maybe we can still keep the `_version.py` file but this time just just a `_version` string variable.

I this way, we don't have to repeat the `importlib` code each time we want to fetch the version

src/towncrier/_shell.py
```diff
@@ -13,14 +13,25 @@
...
+    from importlib.metadata import PackageNotFoundError, version
+except ImportError:  # < Python 3.8
+    from importlib_metadata import PackageNotFoundError, version  # type: ignore
+
+
+try:
```
adiroiban: this code can be included in the updated `_version.py` code

src/towncrier/test/test_packaging.py
```diff
@@ -1,22 +1,16 @@
...
+import towncrier
 
 
 class TestPackaging(TestCase):
-    def test_version_warning(self):
+    def no_version_attr(self):
         """
-        Import __version__ from towncrier returns an Incremental version object
-        and raises a warning.
+        towncrier.__version__ was deprecated, now no longer exists.
```
adiroiban, line 12: not sure if `towncrier.__version__` was considered public or private API.

For private API, raising AttributeError is ok.
If it's public API, we need to raise a separate error with an info message.

src/towncrier/test/test_project.py
```diff
@@ -46,25 +39,61 @@ def test_tuple(self):
         version = get_version(temp, "mytestproja")
         self.assertEqual(version, "1.3.12")
 
-    @skipIf(metadata_version is None, "Needs importlib.metadata.")
     def test_incremental(self):
         """
-        An incremental Version __version__  is picked up.
+        An incremental-like Version __version__  is picked up.
```
adiroiban, line 44: what does `incremental-like` version means ?

I guess that this means a `__version__` object/instance that has the same public API/interface as incremental.
That is, it doesn't have to be `incremental` it only needs to behave like it.

src/towncrier/test/test_project.py
```diff
@@ -46,25 +39,61 @@ def test_tuple(self):
...
         """
-        pkg = "../src"
+        temp = self.mktemp()
+        os.makedirs(temp)
+        os.makedirs(os.path.join(temp, "mytestprojinc"))
+
+        with open(os.path.join(temp, "mytestprojinc", "__init__.py"), "w") as f:
+            f.write(
+                """
+class Version:
```
adiroiban, line 53: maybe a docstring can be added to infom that this tries to "emulate" an incremntal class.

src/towncrier/test/test_project.py
adiroiban: I think that it's ok to keep Exception for now.

Later we can look at raising ClickException or something like this for better CLI integration... but in a separate PR.

adiroiban (changes requested):
The changes look good.

Only minor comments.

I think  that it would be nice to be able to still do `from towncrier._version import version` or something similar, without having to worry about importlib each time.

Now, I don't know if it should raise an exception or return a dummy string, when the version was not found... I would go with a "dummy/placeholder" string.

adiroiban:
@SmileyChris #482 it's on my todo list... but we can trigger a release anytime.

src/towncrier/test/test_project.py
SmileyChris: Correct. The code doesn't check the instance of the __version__, just that it quacks like an incremental version.

hynek:
JFTR: if you'd like to entertain the idea of using `hatch-vcs` to deduplicate version information, I'm happy to help. I used to be skeptical to the idea, but now I've moved all my projects to it. If configured correctly, it also allows for continuous uploads to Test PyPI: https://test.pypi.org/project/structlog/#history

I guess the "you" here is mostly towards @adiroiban since he's doing all the release work and @glyph because he's The Twisted King.

adiroiban:
To me `hatch-vcs` is over-engineering.
I am not sure what problem it tries to solve.

Is it that hard to keep the version as a text inside a .toml file? :)

hynek:
Well, it's not. :) But it deduplicates the version information to one place (we do git tags too) and enables the continuous uploads I've mentioned. But obviously, if you don't want it, it's not happening. Just wanted to offer. :)

hynek:
this needs a rebase / conflict resolution

SmileyChris:
Removed the python 3.7 related fallbacks, updated the version in the toml, and rebased against trunk.
(and pushing again with no change because the `checknewsfragment` test failed for no good reason 😖 )

src/towncrier/__init__.py
```diff
@@ -4,29 +4,3 @@
...
-
-    warnings.warn(
-        "Accessing towncrier.__version__ is deprecated and will be "
-        "removed in a future release. Use importlib.metadata directly "
-        "to query for towncrier's packaging metadata.",
-        DeprecationWarning,
-        stacklevel=2,
-    )
-
-    return __version__
```
hynek, line 32: I'm not sure this should be removed. It seems like there's still a bit of software relying on `pkg.__version__` such that I've starte de-deprecating it in my own packages. See also https://github.com/python-attrs/attrs/issues/1136

(to be clear: it has to be ported, not just left around ;))

src/towncrier/_project.py
```diff
@@ -38,47 +38,45 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
 
 
 def get_version(package_dir: str, package: str) -> str:
-    module = _get_package(package_dir, package)
+    version: Any
```
hynek, line 41: Why Any? Any opts `version` completely out of type checking, is there a rescue for that?

src/towncrier/_project.py
```diff
@@ -38,47 +38,45 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
...
 
 def get_project_name(package_dir: str, package: str) -> str:
     module = _get_package(package_dir, package)
-
     version = getattr(module, "__version__", None)
+    # Incremental has support for package names, try duck-typing it.
+    try:
+        return str(version.package)  # type: ignore
+    except AttributeError:
+        pass
```
hynek, line 80:
```suggestion
    with contextlib.suppress(AttributeError):
        return str(version.package)  # type: ignore
```

is more idiomatic, I think

src/towncrier/_shell.py
hynek: We can do this lazily using `towncrier.__version__` if the block that I've commented on earlier is ported and kill two birds with one stone.

src/towncrier/test/test_packaging.py
```diff
@@ -1,22 +1,16 @@
...
-from incremental import Version
 from twisted.trial.unittest import TestCase
 
-from towncrier._version import _hatchling_version
+import towncrier
 
 
 class TestPackaging(TestCase):
-    def test_version_warning(self):
+    def no_version_attr(self):
```
hynek, line 10: I don't think this runs if it doesn't have a `test_` prefix.

hynek (changes requested):
I've added a few notes, but ultimately I'm unconditionally yielding the decisions to @adiroiban since this has been his dream for years. ;)

src/towncrier/_project.py
```diff
@@ -38,47 +38,45 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
...
 def get_version(package_dir: str, package: str) -> str:
-    module = _get_package(package_dir, package)
+    version: Any
 
+    module = _get_package(package_dir, package)
     version = getattr(module, "__version__", None)
-
     if not version:
-        raise Exception("No __version__, I don't know how else to look")
+        try:
```
adiroiban, line 46: it looks like this is not only about removing incremental, but also about using `importlib_metadata` to get the version of  the target project

I think is best to do this as part of a separate PR ..l like https://github.com/twisted/towncrier/pull/502

sigma67:
Hi @adiroiban @SmileyChris is this PR stale and ready to be picked up by someone else? Or are you planning to continue work?

src/towncrier/_project.py
```diff
@@ -38,47 +38,45 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
...
     if isinstance(version, tuple):
         return ".".join(map(str, version)).strip()
 
+    # Try duck-typing as an Incremental version.
+    if hasattr(version, "base"):
+        try:
+            version = str(version.base()).strip()
+            # Incremental uses `X.Y.rcN`.
+            # Standardize on importlib (and PEP440) use of `X.YrcN`:
+            return version.replace(".rc", "rc")  # type: ignore
```
twm, line 63: FYI this has been fixed on the Incremental side.

adiroiban:
Hi @sigma67 

This PR is stale.

If you have time, feel free to continue working on it.

Thanks

src/towncrier/__init__.py
sigma67: re-added #627

src/towncrier/_project.py
sigma67: changed to str #627

src/towncrier/_project.py
sigma67: done  #627

src/towncrier/test/test_packaging.py
sigma67: changed test to check for deprecationwarning instead #627

src/towncrier/_project.py
sigma67: Stale comment as linked PR has been merged. Updated and integrated latest trunk in  #627

src/towncrier/_project.py
sigma67: Not sure what you are referring to. Should anything in this PR be changed as a result?

src/towncrier/_shell.py
sigma67: `_version.py` restored with contents

```
__version__ = "23.11.1.dev0"
```

sigma67:
Hi @adiroiban, thanks, I've picked it up in #627

adiroiban:
I am closing this as the current effort is with https://github.com/twisted/towncrier/pull/627

PR #496, opened 2023-04-12
Replace pkg_resources by importlib.resources

### LINKED ISSUE #494: [TODO] Stop importing `pkg_resources`
It's deprecated and causes the following warning:
```console
E   DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('sphinxcontrib')`.
E   Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

hynek:
This trades our dependency on setuptools to a conditional one on the importlib-resources backport package.

Fixes #494

### DIFF
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -112,6 +112,11 @@ def load_config_from_file(directory: str, config_file: str) -> Config:
     return parse_toml(directory, config)
 
 
+# Clean up possible temporary files on exit.
+_file_manager = ExitStack()
+atexit.register(_file_manager.close)
+
+
 def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
     if "tool" not in config:
         raise ConfigError("No [tool.towncrier] section.", failing_option="all")
@@ -152,12 +157,10 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
 
     template = config.get("template", _template_fname)
     if template.startswith("towncrier:"):
-        file_manager = ExitStack()
-        atexit.register(file_manager.close)
-
         resource_name = f"templates/{template.split('towncrier:', 1)[1]}.rst"
-        resource_path = resources.files("towncrier") / resource_name
-        file_manager.enter_context(resources.as_file(resource_path))
+        resource_path = _file_manager.enter_context(
+            resources.as_file(resources.files("towncrier") / resource_name)
+        )
 
         if not resource_path.is_file():
             raise ConfigError(
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -3,10 +3,12 @@
 
 from __future__ import annotations
 
+import atexit
 import os
 import sys
 
 from collections import OrderedDict
+from contextlib import ExitStack
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Mapping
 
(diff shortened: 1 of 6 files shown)


src/towncrier/test/helpers.py
```diff
@@ -20,6 +28,13 @@ def write(path: str | Path, contents: str) -> None:
     p.write_text(contents)
 
 
+def read_pkg_resource(path: str) -> str:
+    """
+    Read *path* from the towncrier package.
+    """
+    return (resources.files("towncrier") / path).read_text()
```
graingert:
```suggestion
    return (resources.files("towncrier") / path).read_text("utf8")
```

graingert (dismissed):

src/towncrier/_settings/load.py
```diff
@@ -147,13 +151,14 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
     template = config.get("template", _template_fname)
     if template.startswith("towncrier:"):
         resource_name = "templates/" + template.split("towncrier:", 1)[1] + ".rst"
-        if not pkg_resources.resource_exists("towncrier", resource_name):
+        resource_path = resources.files("towncrier") / resource_name
+        if not resource_path.exists():
```
graingert: Traversable objects are not guaranteed to have a `.exists()` and `.resolve()` method, eg if you're running towncrier from a .whl file, I'm not sure why mypy isn't catching this for you

graingert: see here for a porting guide for resource_filename: https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename

hynek: I was vaguely aware but is that a regression and/or relevant in our use case?

hynek: Particularly because we don't consume the file but only pass its path further. as_file makes no sense to me?

graingert: yeah the previous resource_filename would extract this template from towncrier zips and wheels and place it in a temporary directory

hynek: And then just leave it there??

graingert: yeah it would just leave it there, the new api forces a context manager so the file gets cleaned up

graingert: mypy didn't catch this because `reveal_type(resource_name) # Any`

src/towncrier/_settings/load.py
```diff
@@ -146,14 +152,20 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
 
     template = config.get("template", _template_fname)
     if template.startswith("towncrier:"):
-        resource_name = "templates/" + template.split("towncrier:", 1)[1] + ".rst"
-        if not pkg_resources.resource_exists("towncrier", resource_name):
+        file_manager = ExitStack()
+        atexit.register(file_manager.close)
```
graingert: You're supposed to register this once at module import not every time in some function

src/towncrier/_settings/load.py
```diff
@@ -146,14 +152,20 @@ def parse_toml(base_path: str, config: Mapping[str, Any]) -> Config:
...
+
+        resource_name = f"templates/{template.split('towncrier:', 1)[1]}.rst"
+        resource_path = resources.files("towncrier") / resource_name
+        file_manager.enter_context(resources.as_file(resource_path))
```
graingert: you need something like resource_path = file_manager.enter_context(resources.as_file(resource_path))

graingert (dismissed):

hynek:
I think I've addressed everything; CI will be breaking until someone reviews #497 (or fixes it otherwisely).

hynek:
yay green CI

glyph (approved):

PR #497, opened 2023-04-12
Replace codecov with something simpler for now to fix CI

hynek:
The codecov package has been yanked from PyPI: https://community.codecov.com/t/codecov-yanked-from-pypi-all-versions/4259

If someone wants to figure out their new uploader, be my guest of course.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,6 +237,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
       with:
         # Use latest Python, so it understands all syntax.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,15 +92,18 @@ jobs:
         python-version: ${{ matrix.python.action }}
 
     - name: Install dependencies
-      run: python -m pip install --upgrade pip nox codecov coverage[toml]
+      run: python -m pip install --upgrade pip nox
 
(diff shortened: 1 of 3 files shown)


hynek:
Please note that the codecov/patch check can't pass, but I didn't got forward to remove it before this is agreed on.

adiroiban:
There is this PR that was not reverted, since it doesn't work with non-contributors. - https://github.com/twisted/towncrier/pull/447

---------

From what I know, codecov wants you to use the ["universal" uploader](https://github.com/codecov/uploader) or their dedicated GHA action (which uses bash)


If is the ["standard" Python example](https://github.com/codecov/python-standard/blob/master/.github/workflows/python-standard.yml) from codecov

adiroiban:
I think that we can also keep only the current task and stop using codecov.

I like codecov and I use it for work, but at the same time, the current job is good enough, and we don't have to depend on external software.

hynek:
> I think that we can also keep only the current task and stop using codecov.
> 
> I like codecov and I use it for work, but at the same time, the current job is good enough, and we don't have to depend on external software.

Can we merge then?

adiroiban (approved):
Let's merge.

It would have been nice to have the command from ci.yaml defined inside nox ... but this should not be a blocker.

I try to have as little as possible inside the GHA YAML file.

It makes it easier for dev to troubleshoot, reuse, re-run a command on their local dev env.

---------

I have disabled codecov branch protection from here https://github.com/twisted/towncrier/settings/branch_protection_rules/1000593

hynek:
Thanks Adi!

PR #500, opened 2023-04-26
Document semantic linefeeds in CONTRIBUTING.rst

hynek:
While reviewing #483, I've noticed we never mention that docs should be written one sentence per line.

This fixes it.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -145,5 +145,5 @@ fails with our dummy test commits.
 .. _issues:  https://github.com/twisted/towncrier/issues
 .. _pull request: https://github.com/twisted/towncrier/pulls
 .. _nox: https://nox.thea.codes/
-.. _`one sentence per line`_: https://rhodesmill.org/brandon/2012/one-sentence-per-line/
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -68,10 +65,11 @@ We recommend the following workflow:
    b. Write your test cases and run the complete test suite, see the section
       *Running the test suite* for details.
 
-
    c. Document any user-facing changes in one of the ``/docs/`` files.
+      Please use `one sentence per line`_.
```
adiroiban, line 69: The URL is named "one sentence per line", but I think that the rule is "one sentence **or** clause per line".

```suggestion
      Please use `semantic linefeeds`_.
```

adiroiban (approved):
Thanks for the update.

I think that is best to call it "semantic linefeeds" .

But this is not a blocker.

Feel free to merge as it is, or update and merge.

Thanks again

hynek:
I think I'll leave it because while Clause is better, I think it's somewhat less ambiguous — thanks for the review!

PR #501, opened 2023-04-29
No traceback on missing configuration with `towncrier check`

### LINKED ISSUE #336: Avoid exception traceback with "towncrier check"
# Situation

When calling `towncrier check` in a directory without any Towncrier config files, you get this exception:

```
$ mkdir /tmp/towncrier-test; cd /tmp/towncrier-test
$ python3 -m venv .env
$ source .env/bin/activate
$ .env/bin/pip install towncrier
$ pip show towncrier
Name: towncrier
Version: 21.3.0
[...]
Location: /tmp/towncrier-test/.env/lib/python3.6/site-packages
$ .env/bin/towncrier check
  File ".env/bin/towncrier", line 8, in <module>
    sys.exit(cli())
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/click/core.py", line 829, in __call__
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/click/core.py", line 1066, in invoke
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/click/core.py", line 610, in invoke
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/towncrier/check.py", line 27, in _main
    return __main(compare_with, directory, config)
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/towncrier/check.py", line 32, in __main
    base_directory, config = load_config_from_options(directory, config)
  File "/tmp/towncrier-test/.env/lib64/python3.6/site-packages/towncrier/_settings.py", line 49, in load_config_from_options
    "No configuration file found.\nLooked in: %s" % (base_directory,)
towncrier._settings.ConfigError: No configuration file found.
Looked in: /tmp/towncrier-test
```

# Proposed solution

The traceback shouldn't be displayed. Don't frighten our users. :wink: IMHO it's completely enough if we catch it and output only the error message:

```
$ .env/bin/towncrier check
towncrier._settings.ConfigError: No configuration file found.
Looked in: /tmp/towncrier-test
$ echo $?
1
```

# Possible dependencies

Maybe related to #306

chrysle:
# Description

Fixes #336 

# Checklist

### DIFF
--- a/src/towncrier/newsfragments/501.feature
+++ b/src/towncrier/newsfragments/501.feature
@@ -1,1 +1,3 @@
-Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was generated instead of the error message.
+Calling ``towncrier check`` without an existing configuration, will just show only an error message.
+
+In previous versions, a traceback was generated instead of the error message.
--- a/src/towncrier/newsfragments/501.feature
+++ b/src/towncrier/newsfragments/501.feature
@@ -1,1 +1,1 @@
-Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was also generated.
+Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was generated instead of the error message.
--- a/src/towncrier/newsfragments/501.feature
+++ b/src/towncrier/newsfragments/501.feature
@@ -1,1 +1,1 @@
-Don't raise an exception with ``towncrier check`` on missing configuration.
+Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was also generated.
--- a/src/towncrier/newsfragments/501.feature
+++ b/src/towncrier/newsfragments/501.feature
@@ -1,0 +1,1 @@
+Don't raise an exception with ``towncrier check`` on missing configuration.
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -8,8 +8,7 @@ from twisted.trial.unittest import TestCase
 
 from .._settings import ConfigError, load_config
 from .._shell import cli
-from .helpers import write, with_isolated_runner
-
+from .helpers import with_isolated_runner, write
 
 
 class TomlSettingsTests(TestCase):
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -7,9 +7,8 @@ from click.testing import CliRunner
 from twisted.trial.unittest import TestCase
 
 from .._settings import ConfigError, load_config
-
-from .helpers import write
 from .._shell import cli
+from .helpers import write
 
 
 class TomlSettingsTests(TestCase):
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -157,7 +157,9 @@ orphan_prefix = "~"
 
     def test_load_no_config(self):
         """
-        Check that no exception is raised if no config is found in the base directory.
+        Calling the root CLI without an existing configuration file in the base directory,
+        will exit with code 1 and an info message is sent to stdout.
+        No tracebeback is generated.
         """
         temp = self.mktemp()
         os.makedirs(temp)
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -8,7 +8,8 @@ from twisted.trial.unittest import TestCase
 
 from .._settings import ConfigError, load_config
 from .._shell import cli
-from .helpers import write
+from .helpers import write, with_isolated_runner
+
 
 
 class TomlSettingsTests(TestCase):
@@ -194,7 +195,8 @@ class TomlSettingsTests(TestCase):
         config = load_config(project_dir)
         self.assertEqual(config.package, "a")
 
-    def test_load_no_config(self):
+    @with_isolated_runner
+    def test_load_no_config(self, runner: CliRunner):
         """
         Calling the root CLI without an existing configuration file in the base directory,
         will exit with code 1 and an informative message is sent to standard output.
@@ -202,8 +204,8 @@ class TomlSettingsTests(TestCase):
         temp = self.mktemp()
         os.makedirs(temp)
 
-        runner = CliRunner()
-        result = runner.invoke(cli, ["--dir", temp])
+        result = runner.invoke(cli, ("--dir", temp))
+
         self.assertEqual(
             result.output,
             f"No configuration file found.\nLooked in: {os.path.abspath(temp)}\n",
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -158,8 +158,7 @@ orphan_prefix = "~"
     def test_load_no_config(self):
         """
         Calling the root CLI without an existing configuration file in the base directory,
-        will exit with code 1 and an info message is sent to stdout.
-        No tracebeback is generated.
+        will exit with ConfigError.
         """
         temp = self.mktemp()
         os.makedirs(temp)
--- a/src/towncrier/test/test_settings.py
+++ b/src/towncrier/test/test_settings.py
@@ -7,9 +7,11 @@ import textwrap
 
 from textwrap import dedent
 
+from click.testing import CliRunner
 from twisted.trial.unittest import TestCase
 
 from .._settings import ConfigError, load_config
+from .._shell import cli
 
 
 class TomlSettingsTests(TestCase):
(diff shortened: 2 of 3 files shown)


src/towncrier/newsfragments/501.feature
```diff
@@ -0,0 +1 @@
+Don't raise an exception with ``towncrier check`` on missing configuration.
```
adiroiban: very minor comment.

The release notes are for end users.

The current message sounds like a commit message, and less like a release note. maybe something like this.

And is this about about `towncrier check` or `towncrier` CLI in general?

```suggestion
Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was also generated.
```

src/towncrier/test/test_settings.py
```diff
@@ -153,6 +155,21 @@ def test_towncrier_toml_preferred(self):
         config = load_config(temp)
         self.assertEqual(config.package, "a")
 
+    def test_load_no_config(self):
+        """
+        Check that no exception is raised if no config is found in the base directory.
```
adiroiban: Eveything inside the tests is about checking.

The docstring of the test should explain the specific condition and the expected result.

In this way, if this tests fails during a future refactor,
it should be easier to know if there is a bug in the test or a bug in the refactored code.

```suggestion
        Calling the root CLI without an existing configuration file the base directory,
        will exit with code 1 and an info message is sent to stdout.
        No tracebeback is generated.
```

adiroiban (commented):
Thanks for the changes.

Changes looks good, and I guess that they work for this specific use case.

I am not a big fan us using sys.exit

At the same time, I see that ConfigError is raised in many other places.
Does `towncrier` already handle those cases without a traceback?

For example, if you have a config with `single_file = Non-Boolean-Value` , do you see a clean error? or also the traceback.

I wish we could just do a generic handling of the ConfigError instead of replacing it with sys.exit case by case.

I am leaving this a bit longer for review to see that others are saying.

If nobody else is against this, I guess that we can merge it, and look at creating a general fix in a separate PR.

This PR has a new test and that is already an improvement :)

src/towncrier/newsfragments/501.feature
chrysle: > In previous versions, a traceback was also generated.

Should the last version containing this and type of traceback (`ConfigError`) also be specified?

src/towncrier/test/test_settings.py
chrysle: Thanks, I updated the docstring.

SmileyChris:
> I am not a big fan us using sys.exit

> At the same time, I see that ConfigError is raised in many other places.
> Does `towncrier` already handle those cases without a traceback?

The `_main` function (wrapped as the build command) in `build.py` wraps the call in a `try/except ConfigError` and sys.exits already so the precident is set. This PR should just follow the same pattern.

SmileyChris:
> I wish we could just do a generic handling of the ConfigError instead of replacing it with sys.exit case by case.

Well actually, these could be changed to a `click.ClickException` and then it'd be handled automagically...

I added this to `check.py`s `_main` as an example:
```python
    try:
        __main(compare_with, directory, config)
    except ConfigError as e:
        raise click.ClickException(str(e))
```

Which results in `towncrier check` in a no-configuration file location outputting (with an exit code of 1):
```
Error: No configuration file found.
Looked in: /home/chris/sandbox/abc
```

adiroiban:
Thanks Chris for the feedback.
I think that `raise click.ClickException(str(e))` is the way to go.

No need to reinvent the wheel :)

adiroiban (changes requested):
Thanks @chrysle for the initial PR

Can you please consider using the dedicated `ClickException` ?

I guess that one option is to go with something like this

```diff
-class ConfigError(Exception):
+class ConfigError(click.ClickException):
```

I don't know if we need multiple exception types in `towncrier` 
and I can't see any side-effect for this change.

chrysle:
Thanks for the review! I updated the code. 

However, there seems to be an issue with urllib3 after dropping an OpenSSL version which breaks the doctests. See urllib3/urllib3#2168

src/towncrier/test/test_settings.py
```diff
@@ -192,6 +194,22 @@ def test_towncrier_toml_preferred(self):
         config = load_config(project_dir)
         self.assertEqual(config.package, "a")
 
+    def test_load_no_config(self):
+        """
+        Calling the root CLI without an existing configuration file in the base directory,
+        will exit with ConfigError.
```
adiroiban: I think that for this test, the ConfigError  is not really important.

this looks to me like an end-to-end test.
So the important part is that the error message is sent to stdout (and not stderr) and process will exit with code 1 (non-zero code)

src/towncrier/newsfragments/501.feature
```diff
@@ -0,0 +1 @@
+Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was also generated.
```
adiroiban:
```suggestion
Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was generated instead of the error message.
```

adiroiban (approved):
I think this looks good.
Only minor comments.
Many thanks for the test.

--------

I hope the CI will be green and we can merge this.

Thanks again

chrysle:
Updated! Thanks for reviewing.

src/towncrier/newsfragments/501.feature
```diff
@@ -0,0 +1 @@
+Calling ``towncrier check`` without an existing configuration, will just show only an error message. In previous versions, a traceback was generated instead of the error message.
```
hynek: Semantic line feeds plssss

src/towncrier/test/test_settings.py
```diff
@@ -192,6 +194,22 @@ def test_towncrier_toml_preferred(self):
         config = load_config(project_dir)
         self.assertEqual(config.package, "a")
 
+    def test_load_no_config(self):
+        """
+        Calling the root CLI without an existing configuration file in the base directory,
+        will exit with code 1 and an informative message is sent to standard output.
+        """
+        temp = self.mktemp()
+        os.makedirs(temp)
```
hynek, line 204: Do we need this?

src/towncrier/test/test_settings.py
```diff
@@ -192,6 +194,22 @@ def test_towncrier_toml_preferred(self):
...
+
+        runner = CliRunner()
```
hynek: Empty lines between arrange, act, and assert plssss

hynek (changes requested):
Posted from a phone on a car ;)

src/towncrier/test/test_settings.py
SmileyChris: Yes, but it could use the `@with_isolated_runner` decorator found in `towncrier.test.helpers` instead to simplify things (and the runner can be invoked without `["--dir", temp]`)

chrysle: > (and the runner can be invoked without `["--dir", temp]`)

This doesn't seem to work? The runner looks in a subdirectory below `temp`, for example in `/tmp/tmpjsgwk7qd/towncrier.test.test_settings/TomlSettingsTests/test_load_no_config/t5o5_2b0/temp` instead of `/tmp/tmpjsgwk7qd`

chrysle: > (and the runner can be invoked without `["--dir", temp]`)

This doesn't seem to work? The runner looks in a subdirectory below `temp`, for example in `/tmp/tmpjsgwk7qd/towncrier.test.test_settings/TomlSettingsTests/test_load_no_config/t5o5_2b0/temp` instead of `/tmp/tmpjsgwk7qd`

chrysle:
@hynek Thanks for review! Updated.

src/towncrier/test/test_settings.py
hynek: Are y'all sure about the mktemp? The docs of it claim that it creates the directory and that's what I'd expect!? The name is literally "make temp"?? 😵‍💫

chrysle: As far as I've understood, this only creates a name for the temporary dir. 
Whoops, looks like this is [deprecated](https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables)

hynek: Ugh you're right https://docs.twistedmatrix.com/en/stable/api/twisted.trial.unittest.SynchronousTestCase.html#mktemp I was sure I checked this before. I knew the OG mktemp is deprecated because of race condition problems but I was under the wrong impression that trial did it right. :( Pls disregard. _mumble pytest mumble_

hynek (approved):

chrysle:
> I knew the OG mktemp is deprecated because of race condition problems but I was under the wrong impression that trial did it right.

@hynek So let's deprecate that thing, add a modern solution and do a great clean-out at towncrier!

Where can I volunteer?

hynek:
That code lives in Twisted proper.

PR #502, opened 2023-04-29
#432 Use `importlib.metadata.version` to get the version

### LINKED ISSUE #432: Get the version from pyproject.toml [tool.poetry] section
Currently using towncrier with a project that uses poetry as package manager. The docs says that you can add towncrier's config to `pyproject.toml`, and poetry has metadata version like this

```
[tool.poetry]
...
version = "0.1.0"
```

if adding towncrier metadata to the toml like this
```
[tool.towncrier]
directory = "changes"
package_dir = "."
filename = "NEWS.rst"
```
it complains about lacking the version. Can't towncrier parse the toml looking for version instead of repeating version variables?

agriyakhetarpal:
# Description
This PR updates the `get_version()` method in `src/towncrier/_project.py` to use `importlib.metadata.version()` [Fixes #432]

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ requires-python = ">=3.8"
 dependencies = [
     "click",
     "importlib-resources>=5; python_version<'3.10'",
-    "importlib-metadata==4.6.*; python_version<'3.10'",
+    "importlib-metadata>=4.6; python_version<'3.10'",
     "incremental",
     "jinja2",
     "tomli; python_version<'3.11'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ requires-python = ">=3.8"
 dependencies = [
     "click",
     "importlib-resources>=5; python_version<'3.10'",
+    "importlib-metadata=4.6.*; python_version<'3.10'",
     "incremental",
     "jinja2",
     "tomli; python_version<'3.11'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ requires-python = ">=3.8"
 dependencies = [
     "click",
     "importlib-resources>=5; python_version<'3.10'",
-    "importlib-metadata=4.6.*; python_version<'3.10'",
+    "importlib-metadata==4.6.*; python_version<'3.10'",
     "incremental",
     "jinja2",
     "tomli; python_version<'3.11'",
--- a/src/towncrier/_project.py
+++ b/src/towncrier/_project.py
@@ -11,8 +11,8 @@ from __future__ import annotations
 import sys
 
 from importlib import import_module
-from importlib.metadata import version as metadata_version
 from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as metadata_version
 from types import ModuleType
 
 from incremental import Version as IncrementalVersion
--- a/src/towncrier/_project.py
+++ b/src/towncrier/_project.py
@@ -13,6 +13,7 @@ from importlib import import_module
 from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version as metadata_version
 from types import ModuleType
+from typing import Any
 
 from incremental import Version as IncrementalVersion
 
@@ -20,7 +21,7 @@ from incremental import Version as IncrementalVersion
 try:
     from importlib.metadata import packages_distributions
 except ImportError:
-    from importlib_metadata import packages_distributions
+    from importlib_metadata import packages_distributions  # type: ignore
 
 
 def _get_package(package_dir: str, package: str) -> ModuleType:
@@ -57,6 +58,8 @@ def _get_metadata_version(package: str) -> str | None:
 
 
 def get_version(package_dir: str, package: str) -> str:
+    version: Any
+
     # First try to get the version from the package metadata.
     if version := _get_metadata_version(package):
         return version
--- a/src/towncrier/_project.py
+++ b/src/towncrier/_project.py
@@ -45,6 +45,9 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
 
 
 def _get_metadata_version(package: str) -> str | None:
+    """
+    Try to get the version from the package metadata.
+    """
     distributions = packages_distributions()
     distribution_names = distributions.get(package)
     if not distribution_names or len(distribution_names) != 1:
@@ -54,12 +57,19 @@ def _get_metadata_version(package: str) -> str | None:
 
 
 def get_version(package_dir: str, package: str) -> str:
+    """
+    Get the version of a package.
+
+    Try to extract the version from the distribution version metadata that matches
+    `package`, then fall back to looking for the package in `package_dir`.
+    """
     version: Any
 
     # First try to get the version from the package metadata.
     if version := _get_metadata_version(package):
         return version
 
+    # When no version if found, fall back to looking for the package in `package_dir`.
     module = _get_package(package_dir, package)
 
     version = getattr(module, "__version__", None)
--- a/src/towncrier/_project.py
+++ b/src/towncrier/_project.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 import sys
 
 from importlib import import_module
-from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version as metadata_version
 from types import ModuleType
 from typing import Any
@@ -51,10 +50,7 @@ def _get_metadata_version(package: str) -> str | None:
     if not distribution_names or len(distribution_names) != 1:
         # We can only determine the version if there is exactly one matching distribution.
         return None
-    try:
-        return metadata_version(distribution_names[0])
-    except PackageNotFoundError:
-        return None
+    return metadata_version(distribution_names[0])
 
 
 def get_version(package_dir: str, package: str) -> str:
--- a/src/towncrier/_project.py
+++ b/src/towncrier/_project.py
@@ -47,8 +47,8 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
 def _get_metadata_version(package: str) -> str | None:
     distributions = packages_distributions()
     distribution_names = distributions.get(package)
-    if not len(distribution_names) == 1:
-        # We can't determine the version if there are multiple distributions.
+    if not distribution_names or not len(distribution_names) == 1:
+        # We can only determine the version if there is exactly one matching distribution.
         return None
     try:
         return metadata_version(distribution_names[0])
--- a/src/towncrier/_project.py
+++ b/src/towncrier/_project.py
@@ -5,7 +5,6 @@
 Responsible for getting the version and name from a project.
 """
 
-
 from __future__ import annotations
 
 import sys
@@ -16,6 +15,12 @@ from importlib.metadata import version as metadata_version
 from types import ModuleType
 
 from incremental import Version as IncrementalVersion
+
+
+try:
+    from importlib.metadata import packages_distributions
+except ImportError:
+    from importlib_metadata import packages_distributions
 
 
 def _get_package(package_dir: str, package: str) -> ModuleType:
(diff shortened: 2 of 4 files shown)


agriyakhetarpal:
Some tests in `VersionFetchingTests` are currently failing locally because `importlib.metadata.version()` requires a package to be either installed or editable-installed, and the tests do not do so currently when creating a temporary project. How may I proceed? I have caught `PackageNotFoundError` in a try-except block for now so that the functionality is retained

Alternatives are to try to use `importlib.metadata.distribution` to get the version, or to just parse `pyproject.toml` — the latter is not in consideration since it is neither the current standard nor Pythonic.

SmileyChris:
See also: #491

hynek:
I'm not sure how I feel about this whole"install towncrier along with your package" -- it kinda seems like a misfeature / attractive nuisance for the reasons we've just seen. :-/ 

I would argue for deprecation and adding ways to extract the version from the _outside_ (pyproject.toml / extract using regexp from a file / run a command [e.g. hatch]).

agriyakhetarpal:
It would be much easier to use a `toml` parser library to extract package versions from `pyproject.toml`. With hatch as the build backend, it is just as easy to get it just as it is to update it, –from `hatch version`, which also might be extracting it from `[tool.hatch.version]` or `[project]`, I am not sure.

agriyakhetarpal:
It would be much easier to use a `toml` parser library to extract package versions from `pyproject.toml`. With hatch as the build backend, it is just as easy to get it just as it is to update it – from `hatch version`, which also might be extracting it from `[tool.hatch.version]` or `[project]`, I am not sure.

hynek:
It’s not that simple: hatch supports dynamic versions using plugins like `hatch-vcs` that use git metadata. Running `hatch` would be probably the most correct but also the most invasive approach.

I for one would claim that the user can run `towncrier build --version $
(hatch version)` themselves just fine?

Don’t y’all feel too that these features are children of a bygone Python packaging era when things were simpler and more homogenous?

adiroiban:
I think that `towncrier build --version $(hatch version)` already works.

At the same time, I think that we should make simple things simple.

So if the version is already defined as a simple text value inside the .toml file, I think that it makes sense to accept a PR in which the version is automatically extracted.

We can document that this only works for limited use cases.

----------

I prefer to have all the common commands stored in file.
So maybe `towncrier build --version $(hatch version)` can be implemented as

If towncrier `version` starts with `shell`, then we will call an external command.
I think that we already are calling `git` commands

```
[tool.towncrier]
version = "shell:hatch version" 
```

or if someone already has the package installed, they can use something like

```
[tool.towncrier]
version = "importlib.metadata" 
```


----------

These options will not work for any case... but I don't think that we should reject a PR adding support for a specific use case.

I expect that the majority of project will try to keep things simple, and have the version as a string inside the `.toml` file.


----------

Just my feedback

hynek:
Hmm either way I'm sorry, I think the approach in this PR is the more correct one even though I disagree with the feature as a whole.

It will need something like [this](https://github.com/hynek/structlog/blob/ade1f4b10f361a7ba126e72111d455c1c50779e6/src/structlog/__init__.py#L107-L110) and a conditional dependency `'importlib_metadata; python_version<'3.8'"` to work on 3.7, though.

Please consider this un-bikeshedded otherwise.

adiroiban:
So for this PR to be merged  it needs:

* newsfragment / release notes to advertise this feature
* (nice to have) Update the existing documentation to mention that if you have your project installed, towncrier can extract the version
* automated tests to check the new feature - for this, we might want automatically create a super simple package as part of the test suite, install that package and check that version is detected.

I think there are already some tests generating `setup.cfg/pyproject.toml` on the fly and executing towncrier.

With the automated tests, there is one big problem, if you start a python interpreter/process and while the process is running you install packages... the metadata of the new pacage might not be available.

For my code sometimes I am calling `pip` like this... and I get a warning that this usage is not supported:

```
from pip import main
pip_arguments = ['install', 'some-test-package']
result = main(args=pip_arguments)
```

It works... if in the same process I call `uninstall` pip complains that the package doesn't exist.

So just something to take into consideration when testing.

adiroiban:
I triggered the current test suite to see if there are any regression that might be introduced by this change.

To me, this approach is harmeless ... if you don't have the package install it should work as before.

Also, I don't know if we should worry about python 3.7 ... It will be EOL soon.

hynek:
How about we (read: you 😇) cut a release and drop 3.7 afterwards?

(Pls someone merge my typo PR first tho 😅)

agriyakhetarpal:
Dropping Python 3.7 before this makes a lot of sense, though the backport of `importlib.metadata` works fine as well. I was waiting on a consensus before working further and it shall hopefully be fine using a `toml` parser rather than using regex, just for the sake of convenience

adiroiban:
Well...  I think that the  overall feedback is that people are expecting from `towncrier` to auto-magically detect the version of their project.

Using `importlib.metadata_version` should be the most `standard` way of doing this.

So the direction of this PR is good.

But in order to merge this PR, we need all tests to be green.

agriyakhetarpal:
Oh yes, I just saw that. I meant that I could wait until the next release until 3.7 support is dropped as @hynek suggests, and then work on it

src/towncrier/_project.py
```diff
@@ -39,8 +41,10 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
 
 def get_version(package_dir: str, package: str) -> str:
     module = _get_package(package_dir, package)
```
RazerM, line 73: I think the import via `_get_package` can be avoided unless `importlib.metadata.version` cannot find the package.

Some packages I work on are quite slow to import and it is quite noticeable when we run towncrier. This would be a nice speedup.

src/towncrier/_project.py
```diff
@@ -39,8 +41,10 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
...
-
-    version = getattr(module, "__version__", None)
+    try:
+        version = metadata_version(f"{module}")
```
RazerM: The `str` of a module is not something you can pass to [`importlib.metadata.version`](https://docs.python.org/3/library/importlib.metadata.html#distribution-versions) because you get a repr-like string:

```python
>>> module = importlib.import_module("towncrier")
>>> print(f"{module}")
<module 'towncrier' from '/Users/.../site-packages/towncrier/__init__.py'>
```

which is probably just a small mistake here, but it reminds me that in fact `version()` does not accept an [Import Package](https://packaging.python.org/en/latest/glossary/#term-Import-Package) name at all, but rather a [Distribution Package](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package). They're commonly the same but not necessarily so, to use some popular examples:

| Import Package | Distribution Package |
|----------------|----------------------|
| attr[^1]       | attrs                |
| yaml           | PyYAML               |

[^1]: attrs also has its newer `attrs` namespace of course :-)

As an example, let's say I'm the PyYAML developer, and I'm using towncrier with a `__version__` variable via `package = "yaml"`. If I want to drop that variable and specify my version in `pyproject.toml` instead, then we can't pass `yaml` to `importlib.metadata.version`. We have to load a mapping from one to the other:

```python
>>> import importlib.metadata
>>> importlib.metadata.packages_distributions()['yaml']
['PyYAML']
>>> importlib.metadata.version('PyYAML')
'6.0.1'
```

which in the context of `get_version` would be something like:

```python
dist_packages = distributions_packages()
try:
    dist_package = dist_packages[package]
except KeyError:
    # do fallback to package.__version__ here
else:
    if len(dist_package) > 1:
        raise Exception("I found more than one package")
    version = metadata_version(dist_package[0])
```

So that's how we'd make towncrier use importlib.metadata in a backwards-compatible way (user doesn't have to change their `tool.towncrier` config).

It's a *little* unfortunate, though, that I have to do this:

```toml
[project]
name = "PyYAML"
version = "1.2.3"

[tool.towncrier]
package = "yaml"
```

The version associated with the distribution package is right there, but then I have to tell towncrier one of the import packages I have so that it can look up the name of my distribution package.

PEP 621 says that the `project.name` is required to be statically defined, which means towncrier could try to read `project.name` by default if `package` is not defined and `--version` is not passed.

Notably Poetry doesn't yet support PEP 621 but the `package` would still be there for Poetry users.

agriyakhetarpal:
I realise that this PR has been lost in the hallows of my GitHub account... and it has taken almost an year for me to get back to this, following the notification for the received review, thanks! 🙂

I will shall look to rebase this branch soon and finish this up, because it looks like other PRs are now depending on this (such as #491).

SmileyChris:
So something like my last commit in this PR, @RazerM ?

agriyakhetarpal:
I think the message in the exception raised needs to be updated as well for the test to pass

src/towncrier/_project.py
```diff
@@ -39,15 +44,31 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
     return module
 
 
-def get_version(package_dir: str, package: str) -> str:
-    module = _get_package(package_dir, package)
+def _get_metadata_version(package: str) -> str | None:
+    distributions = packages_distributions()
+    distribution_names = distributions.get(package)
+    if not len(distribution_names) == 1:
```
RazerM: nit: `len(distribution_names) != 1`

src/towncrier/_project.py
```diff
@@ -39,15 +44,31 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
     return module
 
 
-def get_version(package_dir: str, package: str) -> str:
-    module = _get_package(package_dir, package)
+def _get_metadata_version(package: str) -> str | None:
+    distributions = packages_distributions()
+    distribution_names = distributions.get(package)
```
RazerM, line 52: if `distribution_names` is `None` then the `len()` will raise a `TypeError`. Maybe raise a better error in this case?

src/towncrier/_project.py
SmileyChris: Fair enough, changed! 🔬

src/towncrier/_project.py
SmileyChris: It's now `not distribution_names or len(distribution_names) != 1`, returning `None`. That seems fit for purpose...

pyproject.toml
```diff
@@ -31,6 +31,7 @@ requires-python = ">=3.8"
 dependencies = [
     "click",
     "importlib-resources>=5; python_version<'3.10'",
+    "importlib-metadata==4.6.*; python_version<'3.10'",
```
adiroiban: I think that this needs a comment to explain why we are pinning on 4.6 version.

Why not use the latest version ?

src/towncrier/_project.py
```diff
@@ -5,17 +5,24 @@
...
 from types import ModuleType
+from typing import Any
 
 from incremental import Version as IncrementalVersion
 
 
+try:
+    from importlib.metadata import packages_distributions
+except ImportError:
+    from importlib_metadata import packages_distributions  # type: ignore
```
adiroiban, line 23: Maybe add a comment.
This should help with a future cleanup, once 3.9 is no longer supported

```suggestion
    # Fallback on non-stdlib version for
    # Python 3.9 or older.
    from importlib_metadata import packages_distributions  # type: ignore
```

src/towncrier/_project.py
```diff
@@ -37,13 +44,30 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
...
+def _get_metadata_version(package: str) -> str | None:
+    distributions = packages_distributions()
+    distribution_names = distributions.get(package)
+    if not distribution_names or len(distribution_names) != 1:
+        # We can only determine the version if there is exactly one matching distribution.
+        return None
+    return metadata_version(distribution_names[0])
+
+
 def get_version(package_dir: str, package: str) -> str:
```
adiroiban, line 59: Just a minor comment.

Can we have a comment for this method ?

Maybe like

```
Try to extract the versiong for package by first looking at the current runtime.
When no version if found it fallbacks and tries to look for the package in `package_dir`. 
```

src/towncrier/_project.py
```diff
@@ -37,13 +44,30 @@ def _get_package(package_dir: str, package: str) -> ModuleType:
...
+        return version
+
     module = _get_package(package_dir, package)
 
     version = getattr(module, "__version__", None)
 
     if not version:
-        raise Exception("No __version__, I don't know how else to look")
+        raise Exception(
+            f"No __version__ or metadata version info for the {package} package."
```
adiroiban: A minor comment to make the package name explicit


```suggestion
            f'No __version__ or metadata version info for the "{package}" package.'
```

The message `No __version__ or metadata version info for the missing package.` as a bit confusing at first.

src/towncrier/test/test_project.py
```diff
@@ -94,7 +103,8 @@ def test_missing_version(self):
             get_version(tmp_dir, "missing")
```
adiroiban, line 104: Minor comment.

Maybe this is a better name 

```suggestion
            get_version(tmp_dir, "non-existing-subdir")
```

adiroiban (approved):
It looks good. Thanks.

Only minor comments.

--------

There might be things that can be improved. but I think that we can do that in a separate PR.

Thanks again.

src/towncrier/_project.py
RazerM: `packages_distributions` is 3.10+, and since towncrier uses pyupgrade it seems a good idea to use a formation that pyupgrade can automatically remove in the future:

```suggestion
if sys.version_info >= (3, 10):
    from importlib.metadata import packages_distributions
else:
    from importlib_metadata import packages_distributions  # type: ignore
```

pyproject.toml
SmileyChris: A better pin would be >=4.6 I guess (this is the version that had parity with Python 3.10).

src/towncrier/_project.py
SmileyChris: Well, this method doesn't do the fallback, it specifically just looks in the metadata. I'll improve the commenting in get_version.

PR #504, opened 2023-05-02
Update configuration.rst to fix broken link to quickstart.html

michaelkaye:
# Description

https://towncrier.readthedocs.io/en/stable/configuration.html contains the current broken link to https://towncrier.readthedocs.io/en/stable/quickstart.html on the first paragraph, I believe this should likely point to the tutorial instead.

# Checklist

I don't believe i need to tick most of these as this is a documentation change:

### DIFF
--- a/src/towncrier/newsfragments/504.doc
+++ b/src/towncrier/newsfragments/504.doc
@@ -1,0 +1,1 @@
+Update Quick Start in configuration.html to link to tutorial instead.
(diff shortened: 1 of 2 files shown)


chrysle (approved):
Yes, that seems to have been renamed in #435

src/towncrier/newsfragments/504.doc
```diff
@@ -0,0 +1 @@
+Update Quick Start in configuration.html to link to tutorial instead.
```
chrysle:
```suggestion
Update link to Quick Start in configuration.html to point to Tutorial instead.
```

SmileyChris:
PS: this is also fixed in  #483

glyph (approved):
seems correct

PR #507, opened 2023-05-03
Provide a friendly message if version option is required

### LINKED ISSUE #445: Unfriendly error when the version can't be determined
Given this setup:

```console
$ ls
changes  towncrier.toml
$ cat towncrier.toml
[tool.towncrier]
directory = "changes"
```

running `towncrier build` yields this error message:

```
Loading template...
Finding news fragments...
Rendering news fragments...
  File "/home/dpb/.local/bin/towncrier", line 8, in <module>
    sys.exit(cli())
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/towncrier/build.py", line 89, in _main
    return __main(
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/towncrier/build.py", line 147, in __main
    project_version = get_version(
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/towncrier/_project.py", line 40, in get_version
  File "/home/dpb/.local/pipx/venvs/towncrier/lib/python3.10/site-packages/towncrier/_project.py", line 19, in _get_package
    module = import_module(package)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1047, in _gcd_import
  File "<frozen importlib._bootstrap>", line 981, in _sanity_check
ValueError: Empty module name
```

This in no way describes the actual problem, which is that towncrier couldn't determine the project version. It would be nice if it could display a friendlier message, e.g. "Unable to determine the project version. Use the --version option."

SmileyChris:
# Description
Fixes #445

For non-python projects that don't explicitly provide the version in the config, the `--version` option is required. So let the user know rather than raising an obscure exception.

# Checklist

### DIFF
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -153,15 +153,15 @@ def __main(
 
     if project_version is None:
         project_version = config.version
-        if project_version is None:
-            if not config.package:
-                raise UsageError(
-                    "'--version' is required since the config file does "
-                    "not contain 'version' or 'package'."
-                )
-            project_version = get_version(
-                os.path.join(base_directory, config.package_dir), config.package
-            ).strip()
+    if project_version is None:
+        if not config.package:
+            raise UsageError(
+                "'--version' is required since the config file does "
+                "not contain 'version' or 'package'."
+            )
+        project_version = get_version(
+            os.path.join(base_directory, config.package_dir), config.package
+        ).strip()
 
     click.echo("Loading template...", err=to_err)
     if isinstance(config.template, tuple):
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -15,7 +15,7 @@ from datetime import date
 
 import click
 
-from click import Context, Option
+from click import Context, Option, UsageError
 
 from towncrier import _git
 
@@ -151,6 +151,18 @@ def __main(
     base_directory, config = load_config_from_options(directory, config_file)
     to_err = draft
 
+    if project_version is None:
+        project_version = config.version
+        if project_version is None:
+            if not config.package:
+                raise UsageError(
+                    "'--version' is required since the config file does "
+                    "not contain 'version' or 'package'."
+                )
+            project_version = get_version(
+                os.path.join(base_directory, config.package_dir), config.package
+            ).strip()
+
     click.echo("Loading template...", err=to_err)
     if isinstance(config.template, tuple):
         template = resources.read_text(*config.template)
@@ -181,13 +193,6 @@ def __main(
     fragments = split_fragments(
         fragment_contents, config.types, all_bullets=config.all_bullets
     )
-
-    if project_version is None:
-        project_version = config.version
-        if project_version is None:
-            project_version = get_version(
-                os.path.join(base_directory, config.package_dir), config.package
-            ).strip()
 
     if project_name is None:
         project_name = config.name
--- a/src/towncrier/newsfragments/507.misc
+++ b/src/towncrier/newsfragments/507.misc
@@ -1,1 +1,1 @@
-Provide a friendly message if --version option is required (for non-Python projects that don't explicitly provide the version in the config)
+A friendly message is now provided, when it's necessary to pass the ``--version`` option explicitly.
--- a/src/towncrier/newsfragments/507.misc
+++ b/src/towncrier/newsfragments/507.misc
@@ -1,0 +1,1 @@
+Provide a friendly message if --version option is required (for non-Python projects that don't explicitly provide the version in the config)
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -517,9 +517,10 @@ class TestCli(TestCase):
         If the configuration file doesn't specify a version or a package, the version
         option is required.
         """
-        with open("towncrier.toml", "w") as f:
-            f.write("[tool.towncrier]")
+        write("towncrier.toml", "[tool.towncrier]")
+
         result = runner.invoke(_main, ["--draft"], catch_exceptions=False)
+
         self.assertEqual(2, result.exit_code)
         self.assertIn("Error: '--version' is required", result.output)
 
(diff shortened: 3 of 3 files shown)


src/towncrier/build.py
```diff
@@ -145,6 +145,18 @@ def __main(
     base_directory, config = load_config_from_options(directory, config_file)
     to_err = draft
 
+    if project_version is None:
+        project_version = config.version
+        if project_version is None:
+            if not config.package:
+                raise UsageError(
+                    "'--version' is required since the the config file does "
```
chrysle:
```suggestion
                raise UsageError(
                    "Passing '--version' is required since the config file does "
```

SmileyChris: Fixed the typo, but I think the "Passing" is unnecessary. The output looks like this for the usage error:
```
Usage: towncrier build [OPTIONS]
Try 'towncrier build --help' for help.

Error: '--version' is required since the config file does not contain 'version' or 'package'.
```

As an aside: it's pretty ugly that `build` uses a custom `--version` option, since it's overridden by the standard `--version` by click when accessed as the default command by the click `cli` group (e.g. `towncrier --version x.y` just returns `towncrier, version 22.12.1.dev0` rather than actually doing anything).

chrysle: > As an aside: it's pretty ugly that `build` uses a custom `--version` option, since it's overridden by the standard `--version` by click when accessed as the default command by the click `cli` group (e.g. `towncrier --version x.y` just returns `towncrier, version 22.12.1.dev0` rather than actually doing anything).

I think you can open an issue for that?

src/towncrier/newsfragments/507.misc
```diff
@@ -0,0 +1 @@
+Provide a friendly message if --version option is required (for non-Python projects that don't explicitly provide the version in the config)
```
hynek:
```suggestion
A friendly message is now provided, when it's necessary to pass the ``--version`` option explicitly.
```

Let's not over-explain, since it might grow stale / miss something.

src/towncrier/test/test_build.py
```diff
@@ -511,6 +511,18 @@ def test_needs_config(self):
...
         self.assertTrue(result.output.startswith("No configuration file found."))
 
+    @with_isolated_runner
+    def test_needs_version(self, runner: CliRunner):
+        """
+        If the configuration file doesn't specify a version or a package, the version
+        option is required.
+        """
+        with open("towncrier.toml", "w") as f:
+            f.write("[tool.towncrier]")
```
hynek: Please use the write helper.

src/towncrier/test/test_build.py
```diff
@@ -511,6 +511,18 @@ def test_needs_config(self):
...
+        result = runner.invoke(_main, ["--draft"], catch_exceptions=False)
```
hynek, line 522: Let's leave some space between arrange, act, and assert.

```suggestion

        result = runner.invoke(_main, ["--draft"], catch_exceptions=False)

```

src/towncrier/build.py
```diff
@@ -145,6 +145,18 @@ def __main(
...
+        project_version = config.version
+        if project_version is None:
+            if not config.package:
+                raise UsageError(
+                    "'--version' is required since the config file does "
+                    "not contain 'version' or 'package'."
+                )
+            project_version = get_version(
+                os.path.join(base_directory, config.package_dir), config.package
+            ).strip()
```
hynek: Can't run the code rn, but we should be able to unindent this block, no?

hynek (changes requested):

src/towncrier/build.py
SmileyChris: No. `project_verison` may be passed explicitly. If not, then it's set from `config.version`. If it's still blank, *then* we're doing this block.

src/towncrier/build.py
SmileyChris: Sure, done in #510

src/towncrier/build.py
hynek: I don't follow – that's exactly what would happen if you unindent?

Under what circumstances would the behavior differ if you unindented?

SmileyChris: The outer `if project_version is None` comes from the option passed to the `__main` click command.
The inner `if project_version is None` comes from `config.version`
We only want to try and get the version from the python package if both of these are null, *and* if there's a `config.package` actually set (otherwise we raise the new `UsageError`).

SmileyChris: The only difference between this code and the previous code is the new exception within the new `if not config.package:` condition. It's just moved to be in a more logical place in the command (i.e. before trying to load the template, etc).

hynek: Maybe I was unclear (sorry on phone): I'm not asking you to remove any logic but to unindent the whole block starting with the second `if project_version is None:`.

AFAICT, this is the classic "try to fill a variable from different sources until it's not None" pattern.

SmileyChris: I think I was just being dense. I get you now 🤣

hynek (approved):

PR #520, opened 2023-06-08
Don't depend on wheel for build

hynek:
I don't think we need this with hatchling.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
     "hatchling ~= 1.12.2",
-    "wheel ~= 0.38.4",
     "incremental == 22.10.0",
 ]
 build-backend = "hatchling.build"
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
Nice cleanup. Thanks!

It looks like it works

It looks like setuptools is already installed.
I was looking to see that wheel isn't alredy installed by default by the GHA Python distribution.

https://github.com/twisted/towncrier/actions/runs/5212124121/jobs/9405358869?pr=520#step:6:796

```
  Installed Packages
  ==================
  
  argcomplete==3.0.8
  colorlog==6.7.0
  distlib==0.3.6
  filelock==3.12.0
  importlib-metadata==6.6.0
  nox==2023.4.22
  packaging==23.1
  pip==23.1.2
  platformdirs==3.5.1
  setuptools==47.1.0
  typing_extensions==4.6.3
  virtualenv==20.23.0
  zipp==3.15.0
```

hynek:
To clarify: these global packages aren’t available with _isolated_ builds. I’m not even sure how to check which get installed in the build environment to be honest. :)

PR #522, opened 2023-06-12
Document news fragment type in contributing guide

hynek:
I keep hunting for these every time I need them.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -79,7 +79,7 @@ We recommend the following workflow:
       - ``doc`` for improvements to documentation
       - ``removal`` for deprecations and removals
       - ``misc`` for everything else that is linked but not shown in our ``NEWS.rst`` file.
-        Use this for PRs that don't affect end-users and leave them empty.
+        Use this for pull requests that don't affect end-users and leave them empty.
 
    e. Create a `pull request`_.
       Describe in the pull request what you did and why.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -68,8 +68,18 @@ We recommend the following workflow:
    c. Document any user-facing changes in one of the ``/docs/`` files.
       Please use `one sentence per line`_.
 
-   d. Create a newsfragment in ``src/towncrier/newsfragments/`` describing the changes and containing information that is of interest to end-users.
+   d. Create a news fragment in ``src/towncrier/newsfragments/`` describing the changes and containing information that is of interest to end-users.
       Please use `one sentence per line`_ here, too.
```
chrysle:
```suggestion
   d. Create a news fragment in ``src/towncrier/newsfragments/`` describing the changes and containing information that is of interest to end-users. That is a file which name consists out of the number of your PR, and the type of change that is introduced.
      Please use `one sentence per line`_ here, too.
```
I'd describe this even a little more.

hynek: You do realize that this _is_ the Towncrier project, right? 😇 I think this is one of the few places where we don't have to explain what a news fragment is. More text is not necessarily better.

adiroiban: I don't remember which company was describing/documenting the style for writing documentation. But I remember, that the suggestion was to avoid "please" in the docs 

```suggestion
      Use `one sentence per line`_ here, too.
```

adiroiban (approved):
All good. Thanks for the update.

I can see the  result here https://towncrier--522.org.readthedocs.build/en/522/contributing.html#modifying-the-code

it's nice that we can share this file between the repo root and the docs site

PR #523, opened 2023-06-17
Remove type-check dependency

hynek:
I believe we needed types-setuptools before #496, but we don't anymore.


adiroiban (approved):
Thanks

PR #525, opened 2023-06-23
prevent universal_newlines modifying --draft output

### LINKED ISSUE #524: Fragment files with \r\n line endings get written out as \r\r\n
I am using the https://github.com/sphinx-contrib/sphinxcontrib-towncrier Sphinx extension which embeds the latest `towncrier` draft changelog into the built changelog. This works nicely but there is an odd effect that when a changelog fragment has multiple lines separated by \r\n line endings, the draft changelog gets written with those line endings replaced by a double carriage return \r\r\n then reading that back in Python converts it to `\n\n` so that each line of the fragment gets its own paragraph in the rendered changelog.

This effect seems to be caused by mixing binary reading of fragments with text output, as alluded to in this issue #420 by @hynek. I should note that this is all on Windows. Python has a rather weird behaviour trying to handle newlines in text because it forces all newlines to be `\n` internally. So reading a file as text it will convert any `\r\n` newlines into simply `\n`. Correspondingly, when it writes out the string as text it will automatically revert the `\n` to a `\r\n` string. But if the file is read in binary then it will keep the `\r\n` form for all newlines, then when it is written out as text the `\n` is converted to `\r\n`, leaving the written bytes as `\r\r\n`.

You can reproduce this with this very simple example code:
```
with open("example.txt", "w") as fout:
    fout.write("\r\n")
with open("example.txt", "rb") as fin:
    print(fin.read())  # prints "\r\r\n"
with open("example.txt", "r") as fin:
    print(fin.read().encode("utf8"))  # prints "\n\n"
```

Of course, when the content is written straight to the console then you don't see this issue, so it is normally fine for `--draft`, but when the content is read back in to Python for further use then it is a problem (on Windows). The simplest solution to this problem would be to encode the draft output before echoing it here:
https://github.com/twisted/towncrier/blob/3cef9a18fabe34b897f1e6943ffb2ae3085557f4/src/towncrier/build.py#L261
`click` can handle writing the `bytes` just as easily as a `str`, but this preserves the newlines without adding the extra `\r` and makes the problem go away. I can't see any obvious side effects for any other use case of doing this encode step. Happy to put together a PR if desired, although this is only a 1 line change.

bennyrowland:
# Description

Fixes #524 

News fragments are read in binary mode, so on Windows they keep their `\r\n` newlines. When these are echoed to the console in `--draft` mode, Python's universal newlines support converts the `\n` to `os.linesep` (`\r\n`), leaving them as `\r\r\n`. If this output is then read by Python again (in text mode), that same newline support converts both `\r` and `\r\n` into `\n`, causing all newlines to be doubled up (`\n\n`). This commit echoes the `--draft` output as utf8 encoded `bytes` rather than a `str` to prevent this from happening. This also matches the behaviour of writing to a newsfile in non-draft mode.

@adiroiban please let me know what tests (if any) you think would be appropriate for this change and I will add them. I will also add a news fragment detailing the change if you are minded to accept the change.

# Checklist

### DIFF
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -258,7 +258,8 @@ def __main(
             "What is seen below is what would be written.\n",
             err=to_err,
         )
-        click.echo(content)
+        # output as bytes to prevent universal_newlines modifying content #453
+        click.echo(content.encode("utf8"))
         return
 
     click.echo("Writing to newsfile...", err=to_err)


SmileyChris:
Wouldn't the better thing to do here be just stop opening fragments in binary mode?

In `_builder.py`:
```diff
-            with open(full_filename, "rb") as f:
-                data = f.read().decode("utf8", "replace")
+            with open(full_filename, encoding="utf8", errors="replace") as f:
+                data = f.read()
```

bennyrowland:
I can confirm that that would solve my use case, but I assumed that there was some good reason for reading the fragments in binary mode in the first place. I knew that changing the output at this level would not have any side effects on the rest of `towncrier`, but changing the input might, which is why I proposed this change. If the maintainers are happy with changing the input side of things instead then that would be fine with me.

SmileyChris:
Well if it passes the test suite then it should be fine 😂
It does seem the better fix to do it right imo

adiroiban:
> Wouldn't the better thing to do here be just stop opening fragments in binary mode?

I think that the fragments should be just text, so is safe to open them in text mode.

As mentioned by Chris, if the change passes the test it should be good enough.

But it's very important to add a new test that would fail with the code from `trunk`  and that is green with the changes from this PR.

Otherwise, in the future, someone else might revert this change...and  the tests would still  be green and the PR could be approved.

I don't know what test to suggest ... try to get some input similar to what you have on Windows and follow the same  steps.

I can see that we run the tests on Windows, so we should be covered.

Fell free to skip the tests when running on Linux.

Thanks

src/towncrier/build.py
```diff
@@ -258,7 +258,8 @@ def __main(
             "What is seen below is what would be written.\n",
             err=to_err,
         )
-        click.echo(content)
+        # output as bytes to prevent universal_newlines modifying content #453
```
adiroiban, line 265: I prefer to have a self contained comment, that explains why this is needed without having to search the internet / GitHub :)

I still don't fully understand what went wrong here and what is the best way to fix it..

As mentioned by Chris, I think that a better idea is to fix this inside `render_fragments`

adiroiban (changes requested):
Thanks for the changes.

This needs a release notes fragment and at least one new test, which is executed  at least on Windows

PR #536, opened 2023-07-20
Remove a spurious ` in 'towncrier create' doc

dlax:


hynek (approved):
thanks!

adiroiban:
I think that as long as the changes are only doc related, we should be more flexible and don't require a news fragment... it's just a typo  / RST format.

I guess that for this PR,  I can just create a .misc file

But in the future, we might want to improve support for `towncrier check` to ignore a branch if for example it only has .rst file changes.

i will create the .misc file now

I have re-enabled auto-merge and I hope it will be accepted by the CI :)

PR #540, opened 2023-08-04
Vendor click_default_group to make us installable again

### LINKED ISSUE #492: Deprecation from `click-default-group` dependency
When installing `towncrier` in Python 3.11 using pip 23.0.1, I got the warning:

```
DEPRECATION: click-default-group is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change.
A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```

I am making the issue here because [`click-default-group`](https://github.com/click-contrib/click-default-group) hasn't really seen much activity recently. I have opened click-contrib/click-default-group#24 there in addition to this issue.

hynek:
Fixes #492

IANAL, but this should be OK since it keeps the copyright around.

I've blocked it from type checks and coverage because I'm not gonna start writing tests and type annotations that didn't exist before. If someone at some point wants to fix that (and maybe remove unneeded code); be my guest. :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,6 @@ classifiers = [
 requires-python = ">=3.8"
 dependencies = [
     "click",
-    "click-default-group",
     "importlib-resources>=5; python_version<'3.10'",
     "incremental",
     "jinja2",
@@ -146,9 +145,9 @@ strict = true
 exclude = '^src/towncrier/test/test_.*\.py$'
 
 [[tool.mypy.overrides]]
-module = 'click_default_group'
-# 2022-09-04: This library has no type annotations.
-ignore_missing_imports = true
+module = 'towncrier.click_default_group'
+# Vendored module without type annotations.
+ignore_errors = true
 
 [[tool.mypy.overrides]]
 module = 'incremental'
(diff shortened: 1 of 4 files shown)


glyph:
I don't know if we need to make it a whole Policy Thing, but I think it would be nice to submit the pip fix upstream to at least have a PR to point at, in the event that we could ever un-vendor the thing. Is the pip fix hard? I don't know what the pip issue is

glyph (approved):
That said, if this is blocking installs, it's a bit of an academic point about having the comment point to a PR rather than a date; we can discuss after things are working again.

hynek:
The technicalities aren't hard, but they seem to have problems to get access: https://github.com/click-contrib/click-default-group/issues/24

But then again, it's less then 200 LoC that haven't changed since 2019 so I think it's fine to leave it as it is.

webknjaz:
@hynek FYI yesterday I reached out to the author via email and he repackaged the thing today, in response. So maybe this could be reverted after all...

hynek:
Yeah, but why should we? Again: it's 200 LoC that never change and it's one dependency and potential problem less. If click ever breaks compat and we need to update, we can reconsider. But it's not like my ego hangs on that, if they get a release out, for all I care revert.

webknjaz:
I don't care enough either. Just wanted to document this properly.

PR #546, opened 2023-09-15
use markdown default template in markdown docs

### LINKED ISSUE #545: Markdown docs use 'custom' template and don't show/mention default available
Hi,

(PR incoming, but opening this first so I can get the news snippet right).

The [Markdown docs](https://towncrier.readthedocs.io/en/stable/markdown.html) specify how to set up a custom `changelog.d/changelog_template.jinja` file, but these docs predate #483 , which will use the `default.md` template for a `.md` changelog file by default - so for users, it would likely just be simpler to depend on a default template than to copy/paste the 'custom' one specified in the docs.

The `default.md` template seems to also add complexity to handle more cases, and has another couple of desirable features I found, so is likely to be preferred anyway:

* the 'custom' template from the docs will add trailing whitespace at the end of lines that are 'orphan snippets' without an issue
* the `default.md` template puts brackets around issues in a way that seems to match [Keep A Changelog](https://github.com/olivierlacan/keep-a-changelog/blob/main/CHANGELOG.md) more closely, and matching Keep A Changelog is (part of) the objective of that page in the docs.

I think it would be useful for it to be clear to users who want to use markdown that they don't need to configure a template themselves to get towcrier working with markdown - either by switching the docs to just use the default template, or by explicit discussion of this option in the docs.

I suspect the current situation is just a historical consequence of #436 coming a good while before #483 rather than a deliberate choice, but please correct me if I'm wrong!

While I'm writing, thanks for towncrier, I've found it very useful!

EFord36:
# Description
Closes #545

Uses the default markdown templates introduced in #483 in the markdown docs rather than specifying a custom template.

One option for resolving #545 - thought I'd provide a PR as it was simple, but happy if you want to resolve in a different way (e.g. an explicit block in the docs discussing the default markdown template vs the custom one currently used).

# Checklist

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,7 +13,6 @@ build:
     python: "3.11"
 
 python:
-  system_packages: False
   install:
     - method: pip
       path: .
--- a/docs/markdown.rst
+++ b/docs/markdown.rst
@@ -21,7 +21,6 @@ Put the following into your ``pyproject.toml`` or ``towncrier.toml``:
    filename = "CHANGELOG.md"
    start_string = "<!-- towncrier release notes start -->\n"
    underlines = ["", "", ""]
-   template = "changelog.d/changelog_template.jinja"
    title_format = "## [{version}](https://github.com/twisted/my-project/tree/{version}) - {project_date}"
    issue_format = "[#{issue}](https://github.com/twisted/my-project/issues/{issue})"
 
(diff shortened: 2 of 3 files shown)


.readthedocs.yaml
```diff
@@ -13,7 +13,6 @@ build:
     python: "3.11"
 
 python:
-  system_packages: False
```
adiroiban, line 16: this was removed from RTD 2023-08-29 

https://blog.readthedocs.com/drop-support-system-packages/

We need  this change to build the docs

EFord36: thanks - I'll add a new commit

EFord36:
I looked into my problem with `check_newsfragment`, and it was that I had edited it with vim and left a `.545.doc.un~` hanging around, which apparently confuses `towncrier.check`. Maybe that should be a new issue? In any case, it isn't a problem for this PR (that file was never committed anyway).

I removed the issue I was having from the description for clarity - preserving below for transparency:

<details><summary>check_newsfragment issue</summary>
<p>
One problem at the moment - when I run `nox -e pre_commit check_newsfragment`

I get a failure on `check_newsfragment` that I don't understand: `ValueError: multiple files for 545.doc in /Users/ktdb353/workplace/towncrier/src/towncrier/newsfragments` . I'll debug this but wanted to get this PR out close to #545 being opened, and if it's just something wrong with the news snippet, this approach in general could be considered.


</p>
</details>

adiroiban (approved):
Thanks. It looks very good.

https://towncrier--546.org.readthedocs.build/en/546/markdown.html

I hope that in the future we can also have a standard directory name for these files.

I see now that towncrier uses `newsfragments` but this documentation is suggesting `changelog.d`

I prefer something like `release-notes/123.fix` :) 

---------

For the other issue, is best to open a separate ticket, if  you want o work on it.

The idea is that `towncrie` is expected to be used by developers, so there isn't much effort into making the errors "user friendly"

but of course, we can and should improve the error reporting.

PR #548, opened 2023-09-26
Config location separate from directory containing news file and fragments

iliakur:
# Description

This adds official support for the scenario where one towncrier config is shared between multiple projects with independent news files and fragments.

This is useful in monorepo-style setups.

# Checklist

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -12,7 +12,8 @@ The following options can be passed to all of the commands that explained below:
 
 .. option:: --dir PATH
 
-   Build fragment in ``PATH``.
+   The command is executed relative to ``PATH``.
+   For instance with the default config news fragments are checked and added in ``PATH/newsfragments`` and the news file is built in ``PATH/NEWS.rst``.
 
    Default: current directory.
 
--- a/docs/monorepo.rst
+++ b/docs/monorepo.rst
@@ -1,0 +1,52 @@
+Multiple Projects Share One Config (Monorepo)
+=============================================
+
+Several projects may have independent release notes with the same format.
+For instance packages in a monorepo.
+Here's how you can use towncrier to set this up.
+
+Below is a minimal example:
+
+.. code-block:: text
+
+  repo
+  ├── project_a
+  │   ├── newsfragments
+  │   │   └── 123.added
+  │   ├── project_a
+  │   │   └── __init__.py
+  │   └── NEWS.rst
+  ├── project_b
+  │   ├── newsfragments
+  │   │   └── 120.bugfix
+  │   ├── project_b
+  │   │   └── __init__.py
+  │   └── NEWS.rst
+  └── towncrier.toml
+
+The ``towncrier.toml`` looks like this:
+
+.. code-block:: toml
+
+  [tool.towncrier]
+  # It's important to keep these config fields empty
+  # because we have more than one package/name to manage.
+  package = ""
+  name = ""
+
+Now to add a fragment:
+
+.. code-block:: console
+
+   towncrier create --config towncrier.toml --dir project_a  124.added
+
+This should create a file at ``project_a/newsfragments/124.added``.
+
+To build the news file for the same project:
+
+.. code-block:: console
+
+   towncrier build --config towncrier.toml --dir project_a --version 1.5
+
+Note that we must explicitly pass ``--version``, there is no other way to get the version number.
+The ``towncrier.toml`` can only contain one version number and the ``package`` field is of no use for the same reason.
--- a/src/towncrier/newsfragments/548.feature
+++ b/src/towncrier/newsfragments/548.feature
@@ -1,2 +1,2 @@
-Full support for monorepo-style setup.
-One project with multiple independent news files that share the same towncrier config.
+Initial support was added for monorepo-style setup.
+One project with multiple independent news files stored in separate sub-directories, that share the same towncrier config.
--- a/src/towncrier/newsfragments/548.feature
+++ b/src/towncrier/newsfragments/548.feature
@@ -1,0 +1,2 @@
+Full support for monorepo-style setup.
+One project with multiple independent news files that share the same towncrier config.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -144,6 +144,32 @@ class TestCli(TestCase):
         self.assertTrue((project_dir / "NEWS.rst").exists())
 
     @with_isolated_runner
+    def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
+        """
+        Config location differs from the base directory for news file and fragments.
+
+        This is useful when multiple projects share one towncrier configuration.
+        The default `newsfragments` setting already supports this scenario so here
+        we test that custom settings also do.
+        """
+        Path("pyproject.toml").write_text(
+            "[tool.towncrier]\n" + 'directory = "changelog.d"\n'
+        )
+        Path("foo/foo").mkdir(parents=True)
+        Path("foo/foo/__init__.py").write_text("")
+        Path("foo/changelog.d").mkdir()
+        Path("foo/changelog.d/123.feature").write_text("Adds levitation")
+        self.assertFalse(Path("foo/NEWS.rst").exists())
+
+        result = runner.invoke(
+            cli,
+            ("--yes", "--config", "pyproject.toml", "--dir", "foo", "--version", "1.0"),
+        )
+
+        self.assertEqual(0, result.exit_code)
+        self.assertTrue(Path("foo/NEWS.rst").exists())
+
+    @with_isolated_runner
     def test_no_newsfragment_directory(self, runner):
         """
         A missing newsfragment directory acts as if there are no changes.
@@ -1344,40 +1370,3 @@ Deprecations and Removals
 
         self.assertEqual(0, result.exit_code, result.output)
         self.assertEqual(expected_output, result.output)
-
-    @with_isolated_runner
-    def test_projects_share_one_config_with_nondefault_directory(self, runner):
-        """
-        Multiple projects with independent changelogs share one towncrier
-        configuration.
-
-        For this to work:
-        1. We need to leave `config.package` empty.
-        2. We need to pass `--dir` to `create` and `build` explicitly.
-           It must point to the project folder.
-        3. We need to pass `--config` pointing at the global configuration.
-        4. We need to make sure `config.directory` and `config.filename` are resolved
-           relative to what we passed as `--dir`.
-        """
-        # We don't want to specify the package because we have multiple ones.
-        Path("pyproject.toml").write_text(
-            # Important to customize `config.directory` because the default
-            # already supports this scenario.
-            "[tool.towncrier]\n"
-            + 'directory = "changelog.d"\n'
-        )
-        # Each subproject contains the source code...
-        Path("foo/foo").mkdir(parents=True)
-        Path("foo/foo/__init__.py").write_text("")
-        # ... and the changelog machinery.
-        Path("foo/changelog.d").mkdir()
-        Path("foo/changelog.d/123.feature").write_text("Adds levitation")
-        self.assertFalse(Path("foo/NEWS.rst").exists())
-
-        result = runner.invoke(
-            cli,
-            ("--yes", "--config", "pyproject.toml", "--dir", "foo", "--version", "1.0"),
-        )
-
-        self.assertEqual(0, result.exit_code)
-        self.assertTrue(Path("foo/NEWS.rst").exists())
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -146,11 +146,12 @@ class TestCli(TestCase):
     @with_isolated_runner
     def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
         """
-        Config location differs from the base directory for news file and fragments.
-
-        This is useful when multiple projects share one towncrier configuration.
-        The default `newsfragments` setting already supports this scenario so here
-        we test that custom settings also do.
+        Using the `--dir` CLI argument, the NEWS file can
+        be generated in a sub-directory from fragments
+        that are relatives to that sub-directory.
+        
+        The path passed to `--dir` becomes the
+        working directory.
         """
         Path("pyproject.toml").write_text(
             "[tool.towncrier]\n" + 'directory = "changelog.d"\n'
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -1363,7 +1363,8 @@ Deprecations and Removals
         Path("pyproject.toml").write_text(
             # Important to customize `config.directory` because the default
             # already supports this scenario.
-            "[tool.towncrier]\n" + 'directory = "changelog.d"\n'
+            "[tool.towncrier]\n"
+            + 'directory = "changelog.d"\n'
         )
         # Each subproject contains the source code...
         Path("foo/foo").mkdir(parents=True)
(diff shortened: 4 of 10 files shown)


iliakur:
Gave it some more thought: this is looking even more like a Bugfix because `build` with the default `config.directory` supports the usecase as well. This PR only extends that support to custom `config.directory` values.

adiroiban:
Many thanks for the PR.

The thing is that normally an issue is created first.
As part of that issue, we can discuss if the current behaviour is the expect one, or a new feature is required or this is a bugfix.

with that, we can have a PR created and referencing the original issue.

As there is no issue, you can use 548 as the newsfragment / release notes id.

----

Is there already documentation for this functionality ?

If there is no documentation, this can be considered a new feature.

If there is documentation, but the sub-command don't work as documented, then we have a bug.

iliakur:
@adiroiban Thanks for the quick response.

There is currently no documentation, **but** `towncrier add` supports the equivalent usecase. Moreover, `towncrier build` also supports this **as long as** `config.directory` has the default value. 

That's why it seems more like a bug to me than a missing feature. However, I'll defer to you for the final call, I'm also totally fine adding a feature-typed fragment.

adiroiban:
Can you add some documentation as part of this PR.

The documentation should be the ultimate oracle telling what is a bug or a feature.

If something is not documented, then any change is a feature.
If something is already documented, but not working as documented, then we have a bug.

------

I guess that you try to use towncrier with a `monorepo` 

So I think that it make sense to add some information in the docs about how is the monorepo support implemented in `towncrier`.

Regards

iliakur:
@adiroiban Thanks for the feedback, I addressed it. In the process of adding the docs I realized that the `towncrier check` command also needed to be adjusted. With this scope this definitely feels more like a feature!

This PR is ready for review.

docs/cli.rst
```diff
@@ -12,7 +12,8 @@ The following options can be passed to all of the commands that explained below:
 
 .. option:: --dir PATH
 
-   Build fragment in ``PATH``.
+   The command is executed relative to ``PATH``.
```
adiroiban, line 15: Can we also have this information as part of the built-in CLI help system ?

this would need updating the `@click.options` help text for build.

https://github.com/twisted/towncrier/pull/548/files#diff-3576f35f5faa5ded14cd2b3db52e19829e34270f7957ebdf765870dc6f305903R68-R74

-------

In an ideal case, we would only write the docs for `click` and have it automatically exported in RST format.

--------

Same comment is valid for `create` and `check` sub-commands

docs/monorepo.rst
```diff
@@ -0,0 +1,52 @@
...
+
+This should create a file at ``project_a/newsfragments/124.added``.
+
+To build the news file for the same project:
+
+.. code-block:: console
+
+   towncrier build --config towncrier.toml --dir project_a --version 1.5
+
+Note that we must explicitly pass ``--version``, there is no other way to get the version number.
```
adiroiban, line 51: Maybe create a GitHub issue for this.
I think that if we want to support mono-repos, this should be a 1st class feature... so in the ideal case, the version could be extracted

I don't know how to do it... as we are missing the package name info... so it's not easy to know for which package to extract the version.

But maybe it's best to just have the issue created and linked from here.
If someone else wants this feature, they can send suggestion or PRs for how to fix this.

The sentence from below, reads like a bug description :)

src/towncrier/newsfragments/548.feature
```diff
@@ -0,0 +1,2 @@
+Full support for monorepo-style setup.
+One project with multiple independent news files that share the same towncrier config.
```
adiroiban: Just a minor comment to make it read like a release notes.

```suggestion
Initial support was added for monorepo-style setup.
One project with multiple independent news files stored in separate sub-directories, that share the same towncrier config.
```

src/towncrier/test/test_build.py
```diff
@@ -143,6 +143,32 @@ def test_in_different_dir_config_option(self, runner):
...
+        )
+        Path("foo/foo").mkdir(parents=True)
+        Path("foo/foo/__init__.py").write_text("")
+        Path("foo/changelog.d").mkdir()
+        Path("foo/changelog.d/123.feature").write_text("Adds levitation")
+        self.assertFalse(Path("foo/NEWS.rst").exists())
+
+        result = runner.invoke(
+            cli,
+            ("--yes", "--config", "pyproject.toml", "--dir", "foo", "--version", "1.0"),
```
adiroiban, line 169: I see normpath was previously used.

Maybe for the test, we can go wild, and add something like this, to make sure `dir` still works when it is passed as a non-normalized path.

```suggestion
            ("--yes", "--config", "pyproject.toml", "--dir", "./foo/../foo/", "--version", "1.0"),
```

also, do we need to pass `--config=pyproject.toml` here ? I was thinking that this is already the default value.

src/towncrier/test/test_build.py
```diff
@@ -143,6 +143,32 @@ def test_in_different_dir_config_option(self, runner):
...
+    def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
+        """
+        Config location differs from the base directory for news file and fragments.
+
+        This is useful when multiple projects share one towncrier configuration.
+        The default `newsfragments` setting already supports this scenario so here
+        we test that custom settings also do.
+        """
+        Path("pyproject.toml").write_text(
+            "[tool.towncrier]\n" + 'directory = "changelog.d"\n'
```
adiroiban, line 159: Maybe also add an explicit empty `package` config here.
My understanding, is that this is required for this usecase.

src/towncrier/test/test_build.py
```diff
@@ -143,6 +143,32 @@ def test_in_different_dir_config_option(self, runner):
...
         self.assertTrue((project_dir / "NEWS.rst").exists())
 
+    @with_isolated_runner
+    def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
+        """
+        Config location differs from the base directory for news file and fragments.
+
+        This is useful when multiple projects share one towncrier configuration.
+        The default `newsfragments` setting already supports this scenario so here
+        we test that custom settings also do.
```
adiroiban: I think is best to just focus on the core functionality that is tested here.

```suggestion
        Using the `--dir` CLI argument, the NEWS file can
        be generated in a sub-directory from fragments
        that are relatives to that sub-directory.
        
        The path passed to `--dir` becomes the
        working directory.
```

src/towncrier/test/test_check.py
```diff
@@ -297,3 +297,102 @@ def test_get_default_compare_branch_fallback(self):
...
+    @with_isolated_runner
+    def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
+        """
+        Config location differs from the base directory for news file and fragments.
+
+        This is useful when multiple projects share one towncrier configuration.
+        """
+        main_branch = "main"
+        Path("pyproject.toml").write_text(
+            # Important to customize `config.directory` because the default
```
adiroiban, line 309: I don't understand this comment.

Maybe something like

For monorepos, the `directory` is used not as relative to the configuration file, but as relative to the `--dir` CLI argument.

src/towncrier/test/test_check.py
```diff
@@ -297,3 +297,102 @@ def test_get_default_compare_branch_fallback(self):
...
+        subproject1 = Path("foo")
+        (subproject1 / "foo").mkdir(parents=True)
+        (subproject1 / "foo/__init__.py").write_text("")
+        (subproject1 / "changelog.d").mkdir(parents=True)
+        (subproject1 / "changelog.d/123.feature").write_text("Adds levitation")
+        initial_commit(branch=main_branch)
+        call(["git", "checkout", "-b", "otherbranch"])
+
+        # We add a code change but forget to add a news fragment.
+        write(subproject1 / "foo/somefile.py", "import os")
```
adiroiban, line 323: maybe also write a file like `./changelog.d/234.feature` that is relative to the config file.

I think that the main goal of this tests it to make sure that `--dir` is used as the "working directory"

src/towncrier/test/test_check.py
```diff
@@ -297,3 +297,102 @@ def test_get_default_compare_branch_fallback(self):
 
         self.assertEqual("origin/master", branch)
         self.assertTrue(w[0].message.args[0].startswith('Using "origin/master'))
+
+    @with_isolated_runner
+    def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
+        """
+        Config location differs from the base directory for news file and fragments.
+
+        This is useful when multiple projects share one towncrier configuration.
```
adiroiban:
```suggestion
        It can check the fragments located in a sub-directory
        that is specified using the `--dir` CLI argument.
```

src/towncrier/test/test_create.py
```diff
@@ -249,3 +249,35 @@ def test_create_orphan_fragment_custom_prefix(self, runner: CliRunner):
         self.assertEqual(len(change.stem), 11)
         # Check the remainder are all hex characters.
         self.assertTrue(all(c in string.hexdigits for c in change.stem[3:]))
+
+    @with_isolated_runner
+    def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
+        """
+        Config location differs from the base directory for news file and fragments.
+
+        This is useful when multiple projects share one towncrier configuration.
```
adiroiban: Maybe just focus on describing the core functionality that is tested here.


```suggestion
        When the `--dir` CLI argument is passed,
        it will create a new file in directory that is
        created by combining the `--dir` value
        with the `directory` option from the configuration
        file.
```

src/towncrier/test/test_create.py
```diff
@@ -249,3 +249,35 @@ def test_create_orphan_fragment_custom_prefix(self, runner: CliRunner):
...
+        """
+        Path("pyproject.toml").write_text(
+            # Important to customize `config.directory` because the default
```
adiroiban, line 263: I don't understand this comment here.

Maybe, it's important to note that this is a configured using a sub-directory that is later combined with the CLI argument.

Also, maybe it's also important to pass an empty `project` value for this test

adiroiban (commented):
Hi,

Many thanks for this great PR.

I left a few comments.

The changes looks good.

I have a few comments in terms of the docstrings for the tests.

The `--dir` option can be used for mono-repos, but I think that the core features, is that the `--dir` path is joined with the `directory` configuration option to generate the final path.

docs/cli.rst
iliakur: I'll add this to the options. Have you looked into using [this sphinx plugin](https://sphinx-click.readthedocs.io/en/latest/) to extract the documentation automatically?

src/towncrier/test/test_build.py
iliakur: Thanks for "wild" suggestion, I'll try it out!

We must explicitly pass `--config` here because otherwise we will search for `pyproject.toml` inside the directory provided by `--dir` :)

src/towncrier/test/test_check.py
iliakur: Yea, hmm, this must really be complicated cuz we're having to go over it a 3rd time now (I mentioned this in the PR description and in a comment about bug vs feature criteria).

Let me know if the following is clearer.

1. We compute the fragments base dir path **differently** depending on whether `config.directory` is set or not. References: [build](https://github.com/twisted/towncrier/blob/7f37ab5daa72f5a761db1bf14cba1e2d3d87422d/src/towncrier/build.py#L177-L184), [check](https://github.com/twisted/towncrier/blob/7f37ab5daa72f5a761db1bf14cba1e2d3d87422d/src/towncrier/check.py#L111-L118) and [create](https://github.com/twisted/towncrier/blob/7f37ab5daa72f5a761db1bf14cba1e2d3d87422d/src/towncrier/check.py#L111-L118).
2. The implementation for the case when `config.directory` is not set is the same across all 3 commands. This implementation **supports the usecase we care about in this PR**.
3. The problems start in the other case, however. `create` implementation already supports the usecase in this PR whereas `build` and `check` have to be adjusted.
4. Following the logic of testing "wilder" scenarios, and also to make sure I'm testing my changes, I'm setting `config.directory`.

Longer-term I think it makes sense to figure out how to merge these implementations for ease of maintenance. I didn't have time to do so for this PR.

src/towncrier/test/test_check.py
iliakur: Do I understand correctly that we don't want `./changelog.d/234.feature` to influence the changelog generation?

docs/cli.rst
adiroiban: That's the plan. It just needs someone to implement it into the current build system :)

adiroiban (approved):
Thanks for the PR :)

I think the main thing here is that I am doing reviews at 1AM or during the day in 5 minutes intervals between the test runs of my main work :(

---------

But I think we should merge this.

If anyone else is intersted into monorepo part, they can help improve the docs, the test or the configuration.

-------

Please consider updating the tests to check for non-normalized path.

After that, I think that we can merge this.

Otherwise I don't know when I will have to review it or if someone else will have time soon to review this PR.

iliakur:
> Thanks for the PR :)
> 
> I think the main thing here is that I am doing reviews at 1AM or during the day in 5 minutes intervals between the test runs of my main work :(
> 
> But I think we should merge this.
> 
> If anyone else is intersted into monorepo part, they can help improve the docs, the test or the configuration.
> 
> Please consider updating the tests to check for non-normalized path.
> 
> After that, I think that we can merge this.
> 
> Otherwise I don't know when I will have to review it or if someone else will have time soon to review this PR.

Let me know if you have a feeling that you'd want more time to consider this. We can also focus on 1 PR at a time and take a break with #553. 

I appreciate you carving out time to review and comment. 

I'll be addressing the part about non-normalized paths and CLI later today, so we should be good to merge over the weekend.

iliakur:
@adiroiban thanks once more for bearing with this.

I'm a bit tired and sick today, I'll get to it tomorrow for sure. I'm thinking of moving all monorepo-related tests into one file so that it's easier to share setup code and comments for it. I want to test some more combinations of settings. Too bad `trial` doesn't support parametrized tests the way `pytest` does.

adiroiban:
Hi. I hope you will get well soon.

We can merge this PR as it is. No need to have a perfect PR. As long as it's better that we have in `trunk` and doesn't introduce any regressions, it should be good to merge.

My comments were more of a "nice to have" category.

Let me know if you want to have this PR merge as it is ? (I will take care of updating/rebase with trunk before merge)

----------

I have never worked on mono-repos so I don't have experince with them.
So my feedback from this PR related to mono-repo might not be best.

This is why we can merge this and if someone else wants to use towncrier with monorepo, they can add more feedback or suggest changes.

---------

We can consider adding support for `parametrize` in trial.

Is this the feature ?

```
@pytest.mark.parametrize("test_input,expected", [("3+5", 8), ("2+4", 6), ("6*9", 42)])
def test_eval(test_input, expected):
    assert eval(test_input) == expected
```

I know that instead of simple `parametrize` in twisted/twisted we use https://hypothesis.readthedocs.io/en/latest/ ... not sure if it's the same thing.

iliakur:
Alrighty, let's merge then :)

I owe you some docs and some tests.

---

To be precise:

- [ ] make sure cli and web docs are the same
- [ ] extend and refactor tests
- [ ] open issue base on one of your comments.

---

Your example with `pytest.parametrize` is correct. The concept is similar to `hypothesis.example`.

---

I have just one favor to ask of you: would it be possible to cut a release with this in the next couple days? That way I can incorporated it properly in our infrastructure (referencing branches isn't allowed in "prod").

adiroiban:
> I owe you some docs and some tests.

You don't owe me anything :)

towncrier is just a project that I find useful, and was not mantained, so in my spare time I look after it.

we are standing on the shoulder of giants as many other people have contributed significant time for the project.

--------

> I have just one favor to ask of you: would it be possible to cut a release

Sure. Will do.

-------

I have updated the branch and enabled auto-merge . I hope this will merge soon :)

Issue #552, opened 2023-10-17
Make whitespace between releases configurable


iliakur:
# Status Quo

`towncrier build` will generate the following example changelog entry

```
<!-- towncrier release notes start -->

## <New Release>

### Changed

- Something changed


## Old Release
```

Notice how there are 2 newlines between `Something changed` and `## Old Release`.

It may be useful in some contexts to customize it. For instance I'm on a team that prefers **not** to have the second newline between releases.

I think the most flexible solution would be to add a configuration field called something like `release_separator`. It could have a default value of either `\n` or `\n\n` depending on what we consider as an integral part of one release section.

adiroiban:
I think that this can be configured by providing a custom Jinja2 template.

Would that work for you?

Cheers

iliakur:
I tried removing all kinds of whitespace from this template with no success:

```
{% if sections[""] %}
{% for category, val in definitions.items() if category in sections[""] %}
***{{ definitions[category]['name'] }}***:

{% for text, values in sections[""][category].items() %}
* {{ text }} {{ values|join(', ') }}
{% endfor %}

{% endfor %}
{% else %}
No significant changes.


{% endif %}
```

lmk if you need any more setup from me to repro this.

adiroiban:
I guess that this is an of Jinja whitespace control.

https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control

Just a very quick and brief suggestion.

Have you tried something like ?

```
{% if sections[""] -%}
{%- for category, val in definitions.items() if category in sections[""] %}
```

iliakur:
It's a good call with the jinja whitespace control, I forgot about that.

After trying to remove all  possible whitespace in the jinja template without success, I dove into the towncrier code and I think [this line](https://github.com/twisted/towncrier/blob/239f84102c374e7395cdc0a50f7b4171cc8c3e83/src/towncrier/_writer.py#L50) is the culprit.

No matter what my template looks like, there are always 2 newlines inserted before the previous releases...

I don't really know of a backwards-compatible way to solve this without exposing a config variable and setting it to `\n\n` by default.

wdyt?

adiroiban:
Thanks for looking into this.

Would it work if the hardcoded newlines are removed from the code and the default template is updated to compensate for that?

If we somehow could get full functionality via the template, I would prefer that, over adding many configuratin flags.

But if this is big issue for you and submit a PR with a new configuration option that is well documented and tested, I would not reject it.

-------

[...]

PR #553, opened 2023-10-19
Whitespace between releases can be configured in the template.

iliakur:
It used to be baked into the rendering code.

# Description

Implementation of #552 

# Checklist


src/towncrier/_writer.py
```diff
@@ -44,10 +44,7 @@ def append_to_newsfile(
         if header:
             f.write(header)
 
-        f.write(content)
-
-        if prev_body:
-            f.write(f"\n\n{prev_body}")
+        f.write(content + prev_body if prev_body else content.rstrip() + "\n")
```
adiroiban, line 48: maybe add a comment why the `content.rstrip() + '\n'` vs `content + prev_body` 

I guess that this is here to support having separate news files for each release or for draft output.

------

Also, I was expecting for the writer, to just do the write,  without updating the content.

Generating the content should be the job of the builder.

src/towncrier/build.py
```diff
@@ -260,7 +260,7 @@ def __main(
             "What is seen below is what would be written.\n",
             err=to_err,
         )
-        click.echo(content)
+        click.echo(content.rstrip() + "\n")
```
adiroiban: Just a minor comment ... and just asking :)

why strip the content here ?

I think it's ok to have one extra newline, but leave the other newlines as they are defined in the template.

This can help debug / preview the actual fragment.

src/towncrier/newsfragments/552.feature
```diff
@@ -0,0 +1,3 @@
+Make whitespace between releases configurable with the template.
```
adiroiban:
```suggestion
Two newlines are no longer always added between the current release notes and the previous content.
The newlines are now defined only inside the template.
```

src/towncrier/templates/default.md
```diff
@@ -58,4 +58,5 @@ No significant changes.
 No significant changes.
 
 {% endif %}
-{% endfor %}
+{% endfor +%}
+{# This comment adds one more newline at the end of the rendered newsfile content. #}
```
adiroiban: Maybe add a few more info about the design decision here.

```suggestion
{#- This comment adds one more newline at the end of the rendered newsfile content.
In this way the there are 2 newlines between the latest release and the previous release content.
#}
```

adiroiban (approved):
Thanks for the changes.

Changes looks good.

I left a few minor comments.

--------

A note for another PR

I now see that we don't run `towncrier draft` as part of the CI process.

I guess that it could help to have it executed and the output sent as part of the job output... 

similar to the output that we have for build and coverage

https://github.com/twisted/towncrier/actions/runs/6580860357#summar

src/towncrier/build.py
iliakur: I was on the fence about this one, but I think you're right, it makes more sense for debugging to keep the whitespace as-is.

I'll need to adjust quite a few tests though :)

src/towncrier/_writer.py
iliakur: It's an interesting point about the builder vs writer responsibilities. Maybe the names aren't great, but my impresson is that `builder` is responsible for generating new content, whereas the writer is responsible for incorporating this new content in an existing news file.

src/towncrier/build.py
adiroiban: My understanding is that with this change, you can't configure some newlines in the template, as they are stripped :) ... and this is contrary to the purpose of this PR.

adiroiban: But we can leave it like this, if this will cause a big change.

iliakur: This particular change is only for the draft command though ;)

So the newlines are preserved when generating the actual file, but stripped for the "preview".

I do, however, like the debugging argument, that's worth adjusting the tests.

src/towncrier/_writer.py
adiroiban: I have the same view :) `builder` renders new content. `writers` puts together the new rendered contents with the existing (already rendered) content :)

iliakur:
@adiroiban this is ready for review, thanks!

adiroiban:
Thanks for the extra comment. Looks good. I will merge.

PR #557, opened 2023-10-24
Make consistent the expectation around non-existent fragments directory

### LINKED ISSUE #538: FileNotFoundError, Failed to list the news fragment files
After [adopting towncrier](https://github.com/jaraco/skeleton/pull/84) for the hundreds of projects that depend on [jaraco/skeleton](/jaraco/skeleton), I've been pleased to say that it has helped streamline the changelog generation.

I've tried as much as possible to minimize the amount of config required (relying on towncrier to provide best practices by default). Unfortunately, if I try to cut a release when there are no newsfragments, the "[finalize](https://github.com/jaraco/jaraco.develop/blob/9061f3bfc77c39fa2c3b1e06ea0c7695edb753ef/jaraco/develop/finalize.py#L9-L13)" step fails:

```
 pytest-checkdocs main @ tox -e finalize
finalize: commands[0]> python -m jaraco.develop.finalize
Loading template...
Finding news fragments...
Failed to list the news fragment files.
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jaraco/code/jaraco/pytest-checkdocs/newsfragments'
```

I can work around the issue by first running `mkdir -p newsfragments`, but I'd rather not have to add that step. I know I can create a permanent `newsfragments` directory in every project, but that requires adding a sentinel file to keep it present. And given that towncrier already works well to create that directory when it doesn't exist, I'd like it to fall back to the degenerate behavior when running `build`.

Can you update towncrier build to treat a missing fragments directory as the same as an empty directory, same as towcrier create?

jaraco:
# Description

In the first commit, I update the documentation to reflect the current behavior, highlighting that a missing fragments directory is normal for a "create" operation and caused by the "build" operation.

In the second commit, I expand the documentation to include the case where "build" is invoked and the fragments directory is missing to be consistent with an empty directory or no directory at all. I also draft a likely way to achieve this behavior. Once tests pass, this approach closes #538.

Note that this change aligns the behavior with the [currently prescribed expectation](https://github.com/twisted/towncrier/blob/0b023fa95926470482418d4e11f9ae9f0ada7d56/src/towncrier/test/test_build.py#L178).

# Checklist

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -24,6 +24,12 @@ The following options can be passed to all of the commands that explained below:
 Build the combined news file from news fragments.
 ``build`` is also assumed if no command is passed.
 
+If there are no news fragments (including an empty fragments directory), a
+notice of "no significant changes" will be added to the news file.
+
+By default, the processed news fragments are removed using ``git``, which will
+also remove the fragments directory if now empty.
+
 .. option:: --draft
 
    Only render news fragments to standard output.
@@ -66,6 +72,8 @@ Create a news fragment in the directory that ``towncrier`` is configured to look
    $ towncrier create 123.bugfix.rst
 
 ``towncrier create`` will enforce that the passed type (e.g. ``bugfix``) is valid.
+
+If the fragments directory does not exist, it will be created.
 
 If the filename exists already, ``towncrier create`` will add (and then increment) a number after the fragment type until it finds a filename that does not exist yet.
 In the above example, it will generate ``123.bugfix.1.rst`` if ``123.bugfix.rst`` already exists.
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -24,8 +24,9 @@ The following options can be passed to all of the commands that explained below:
 Build the combined news file from news fragments.
 ``build`` is also assumed if no command is passed.
 
-If there are no news fragments (including an empty fragments directory), a
-notice of "no significant changes" will be added to the news file.
+If there are no news fragments (including an empty fragments directory or a
+non-existent directory), a notice of "no significant changes" will be added to
+the news file.
 
 By default, the processed news fragments are removed using ``git``, which will
 also remove the fragments directory if now empty.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -182,8 +182,8 @@ class TestCli(TestCase):
 
         result = runner.invoke(_main, ["--draft", "--date", "01-01-2001"])
 
-        self.assertEqual(1, result.exit_code, result.output)
-        self.assertIn("Failed to list the news fragment files.\n", result.output)
+        self.assertEqual(0, result.exit_code)
+        self.assertIn("No significant changes.\n", result.output)
 
     def test_no_newsfragments_draft(self):
         """
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -6,14 +6,11 @@ from __future__ import annotations
 
 import os
 import textwrap
-import traceback
 
 from collections import defaultdict
 from typing import Any, DefaultDict, Iterable, Iterator, Mapping, Sequence
 
 from jinja2 import Template
-
-from ._settings import ConfigError
 
 
 def strip_if_integer_string(s: str) -> str:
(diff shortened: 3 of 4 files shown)


src/towncrier/test/test_build.py
```diff
@@ -182,8 +182,8 @@ def test_no_newsfragment_directory(self, runner):
 
         result = runner.invoke(_main, ["--draft", "--date", "01-01-2001"])
 
-        self.assertEqual(1, result.exit_code, result.output)
-        self.assertIn("Failed to list the news fragment files.\n", result.output)
+        self.assertEqual(0, result.exit_code)
+        self.assertIn("No significant changes.\n", result.output)
```
adiroiban, line 186: Thanks. It looks like this was a bug.

The test docstring states that it is expected to handle it as  no changes, but in fact it was failing :)

docs/cli.rst
```diff
@@ -24,6 +24,13 @@ The following options can be passed to all of the commands that explained below:
 Build the combined news file from news fragments.
 ``build`` is also assumed if no command is passed.
 
+If there are no news fragments (including an empty fragments directory or a
+non-existent directory), a notice of "no significant changes" will be added to
+the news file.
+
+By default, the processed news fragments are removed using ``git``, which will
+also remove the fragments directory if now empty.
```
adiroiban, line 32:
```suggestion
By default, the processed news fragments are removed using ``git rm``.
```

I don't think that `towncrier` (or git) explicitly removes the empty dir.

src/towncrier/newsfragments/538.feature.rst
```diff
@@ -0,0 +1 @@
+``build`` now treats a missing fragments directory the same as an empty one, consistent with other operations.
```
adiroiban: Is this a feature or a bug ?  Just asking.

Based on the test docstring .

adiroiban (approved):
Looks very good. Many thanks.

Only minor comments.

docs/cli.rst
jaraco: I tested this expectation in one of my projects. For example, you can checkout /jaraco/skeleton, which currently has one fragment, and run towncrier build 0.0.0, and it will remove the newsfragmenta dir.

src/towncrier/newsfragments/538.feature.rst
jaraco: My rationale, even if it was a subtly indicated behavior, is that it's for most users a change in the expected behavior. I also could see it considered a bugfix.

It could also theoretically be considered a breaking change for users reliant on the current behavior (seems unlikely). I'll go with whatever scope the maintainers wish.

Since the latest recommendation is for bugfix, I'll switch to that, but with openness to change.

docs/cli.rst
jaraco: Here's where I did just that yesterday:

```
 @ work-on skeleton
 skeleton main @ towncrier build --version 0.0.0
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
/Users/jaraco/code/jaraco/skeleton/newsfragments/+drop-py37.feature.rst
Is it okay if I remove those files? [Y/n]: y
Removing news fragments...
Done!
 skeleton main @ cat NEWS.rst
0.0.0
=====

Features
--------

- Require Python 3.8 or later.
 skeleton main @ towncrier build --version 0.0.1
Loading template...
Finding news fragments...
Failed to list the news fragment files.
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jaraco/code/jaraco/skeleton/newsfragments'

 skeleton main @ mkdir newsfragments
 skeleton main @ towncrier build --version 0.0.1
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
No news fragments to remove. Skipping!
Done!
```

adiroiban:
Thanks. Sorry for the extra trouble. I have merged it.

iliakur:
@adiroiban you're gonna laugh, but I think we ran into this situation as well. I think we're gonna change our process for other reasons in a way that this isn't a problem. However, just in case, would you mind cutting a release with this in the next few days? Thanks a bunch!!

adiroiban:
@iliakur here is the release PR https://github.com/twisted/towncrier/pull/566

PR #558, opened 2023-10-24
Update the documentation for the release process.

adiroiban:
# Description

Update the releaes process docs.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -38,8 +38,8 @@ Create a new release candidate using `GitHub New release UI <https://github.com/
 * *Choose a tag*: Type `19.9.0rc1` and select `Create new tag on publish.`
 * *Target*: Search for the release branch and select it.
 * *Title*: "Towncrier 19.9.0rc1".
-* Set the content based on the NEWS file.
-* Make sure to mark **This is a pre-release**.
+* Set the content based on the NEWS file (for now in RST format).
+* Make sure to check **This is a pre-release**.
 * Click `Publish release`
 
 This will trigger the PyPI release candidate.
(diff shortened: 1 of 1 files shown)


adiroiban:
@webknjaz Let me know if the instruction make sense.
I added a note about creating the discussion.

----------

> As for signing, I can look into it sometime, if there's any interest.

I think that signing would be great.

I didn't had time to check how this is done. 

Right now, the PyPi packages are generated by GitHub Actions ... so no signing is involved.

If you want to see towncrier packages signed, feel free to send a PR. But only if you have time. Really :)

Much appreciated

twm (approved):
Looks reasonable to me. Thanks Adi!

adiroiban:
Thanks for the review.

hynek:
sorry for coming late, but I've been secretly doing the rst → md transformation of the releases and it's really easy so it might be worth adding?

The command line is `pandoc --wrap=none --from rst --to gfm` with stdin/stdout. So to simply transform the current clipboard:

- on macOS it's `pbpaste | pandoc --wrap=none --from rst --to gfm | pbcopy`
- on Linux `xclip -selection clipboard -o | pandoc --wrap=none --from rst --to gfm | xclip -selection clipboard` or `xsel --clipboard --output | pandoc --wrap=none --from rst --to gfm | xsel --clipboard --input`
- on Windows `Get-Clipboard | pandoc --wrap=none --from rst --to gfm | Set-Clipboard`

(Linux & Windows untested)

adiroiban:
I will keep this in mind for the next release, and we can update the docs.

----------

Rather than doing pandoc, maybe it is worth also having a NEWS.md file

This can help as a showcase for MD support in towncrier.

I guess that in these days, MD is more popular than RST so we should consider .MD as first class format in `towncrier`

hynek:
That's a great idea – there seems still to be the impression that Towncrier is rst-only.

PR #559, opened 2023-10-27
Update build-and-inspect-python-package to 1.5.2

hynek:
1.5 is broken as GitHub Actions switched 3.x to 3.12 by default


adiroiban (approved):
all good. thanks!

PR #564, opened 2023-11-07
Fix orphan fragments with numbers

### LINKED ISSUE #562: orphan fragments that contain a number in the filename are linked to issues.
e.g. calling the fragment `CHANGES/+compat.somelib_1.12.feature` will link the supposed orphan snippet to the issue `#12`.

mdellweg:
# Description
Fixes #562

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@
 .vscode
 Justfile
 *egg-info/
-towncrier.test.test_*/
 _trial_temp*/
 apidocs/
 dist/
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@
 .vs/
 .vscode
 Justfile
+*egg-info/
+towncrier.test.test_*/
 _trial_temp*/
 apidocs/
 dist/
--- a/src/towncrier/newsfragments/562.bugfix
+++ b/src/towncrier/newsfragments/562.bugfix
@@ -1,0 +1,1 @@
+Orphan newsfragments containing numeric values are no longer accidentally associated to tickets. In previous versions the orphan marker was ignored and the newsfragment was associated to a ticket having the last numerical value from the filename.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -60,7 +60,7 @@ class TestCli(TestCase):
 
             result = runner.invoke(command, ["--draft", "--date", "01-01-2001"])
 
-        self.assertEqual(0, result.exit_code)
+        self.assertEqual(0, result.exit_code, result.output)
         self.assertEqual(
             result.output,
             dedent(
@@ -78,7 +78,7 @@ class TestCli(TestCase):
                 --------
 
                 - Baz levitation (baz)
-                - Baz fix levitation (#2)
+                - Baz fix levitation (fix-1.2)
                 - Adds levitation (#123)
                 - Extends levitation (#124)
                 - An orphaned feature ending with a dotted number
@@ -417,6 +417,7 @@ class TestCli(TestCase):
             call(["git", "init"])
             call(["git", "config", "user.name", "user"])
             call(["git", "config", "user.email", "user@example.com"])
+            call(["git", "config", "commit.gpgSign", "false"])
             call(["git", "add", "."])
             call(["git", "commit", "-m", "Initial Commit"])
 
@@ -441,6 +442,7 @@ class TestCli(TestCase):
             call(["git", "init"])
             call(["git", "config", "user.name", "user"])
             call(["git", "config", "user.email", "user@example.com"])
+            call(["git", "config", "commit.gpgSign", "false"])
             call(["git", "add", "."])
             call(["git", "commit", "-m", "Initial Commit"])
 
@@ -470,6 +472,7 @@ class TestCli(TestCase):
         call(["git", "init"])
         call(["git", "config", "user.name", "user"])
         call(["git", "config", "user.email", "user@example.com"])
+        call(["git", "config", "commit.gpgSign", "false"])
         call(["git", "add", "."])
         call(["git", "commit", "-m", "Initial Commit"])
 
@@ -503,6 +506,7 @@ class TestCli(TestCase):
         call(["git", "init"])
         call(["git", "config", "user.name", "user"])
         call(["git", "config", "user.email", "user@example.com"])
+        call(["git", "config", "commit.gpgSign", "false"])
         call(["git", "add", "."])
         call(["git", "commit", "-m", "Initial Commit"])
 
@@ -531,6 +535,7 @@ class TestCli(TestCase):
             call(["git", "init"])
             call(["git", "config", "user.name", "user"])
             call(["git", "config", "user.email", "user@example.com"])
+            call(["git", "config", "commit.gpgSign", "false"])
             call(["git", "add", "."])
             call(["git", "commit", "-m", "Initial Commit"])
 
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -43,6 +43,14 @@ class TestCli(TestCase):
                 f.write("Orphaned feature")
             with open("foo/newsfragments/+xxx.feature", "w") as f:
                 f.write("Another orphaned feature")
+            with open("foo/newsfragments/+123_orphaned.feature", "w") as f:
+                f.write("An orphaned feature starting with a number")
+            with open("foo/newsfragments/+12.3_orphaned.feature", "w") as f:
+                f.write("An orphaned feature starting with a dotted number")
+            with open("foo/newsfragments/+orphaned_123.feature", "w") as f:
+                f.write("An orphaned feature ending with a number")
+            with open("foo/newsfragments/+orphaned_12.3.feature", "w") as f:
+                f.write("An orphaned feature ending with a dotted number")
             # Towncrier ignores files that don't have a dot
             with open("foo/newsfragments/README", "w") as f:
                 f.write("Blah blah")
@@ -73,6 +81,10 @@ class TestCli(TestCase):
                 - Baz fix levitation (#2)
                 - Adds levitation (#123)
                 - Extends levitation (#124)
+                - An orphaned feature ending with a dotted number
+                - An orphaned feature ending with a number
+                - An orphaned feature starting with a dotted number
+                - An orphaned feature starting with a number
                 - Another orphaned feature
                 - Orphaned feature
 
--- a/src/towncrier/test/test_builder.py
+++ b/src/towncrier/test/test_builder.py
@@ -8,63 +8,73 @@ from .._builder import parse_newfragment_basename
 
 class TestParseNewsfragmentBasename(TestCase):
     def test_simple(self):
+        """<number>.<category> generates a counter value of 0."""
         self.assertEqual(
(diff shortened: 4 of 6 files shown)


adiroiban:
We can discuss the changes here.

The test changes looks great. It should be enought to update the existing tests.

No need for a new tests... a new test would be nice... but the current one is also ok.

I see the updated tests failing. THat's excellent.

~~Not sure why 3.9 wasn't failing and also mypy 3.11~~ (it was the build job, not the test one)

.gitignore
```diff
@@ -17,6 +17,8 @@
 .vs/
 .vscode
 Justfile
+*egg-info/
```
adiroiban, line 20: I guess we can have this. I have this rule in my global gitignore file

.gitignore
```diff
@@ -17,6 +17,8 @@
...
+towncrier.test.test_*/
```
adiroiban: not sure why this is needed ... we can have it if it helps you. just asking :)

.gitignore
mdellweg: good call.
Also this is a python project, so i'd expect that directory to exist eventually.

.gitignore
mdellweg: I can remove it.
These dirs popped up after calling trial.

src/towncrier/test/test_build.py
```diff
@@ -70,9 +78,13 @@ def _test_command(self, command):
                 --------
 
                 - Baz levitation (baz)
-                - Baz fix levitation (#2)
+                - Baz fix levitation (fix-1.2)
```
mdellweg, line 81: I'm not so sure about this one. But i'd argue, that is what was actually meant by:
https://github.com/twisted/towncrier/pull/564/files#diff-36db088409efe07783d49792a18cec6ae3778ace338bf1cd08daa8a3fce29e93R45-R47
```
# something-cool.feature.ext for projects that don't use ticket
```

adiroiban, line 81: I think that this need a better comment.

I don't really understant what is the use case here.

This is the bad part of bundling all kind of scenarios in a single test case.

```
            # Towncrier supports files that have a dot in the name of the
            # newsfragment
            with open("foo/newsfragments/fix-1.2.feature", "w") as f:
                f.write("Baz fix levitation")
```

since it has no orphan marker, I would consider it an error, and the file should be rejected.

src/towncrier/test/test_build.py
```diff
@@ -405,6 +417,7 @@ def test_draft_no_date(self):
             call(["git", "init"])
             call(["git", "config", "user.name", "user"])
             call(["git", "config", "user.email", "user@example.com"])
+            call(["git", "config", "commit.gpgSign", "false"])
```
adiroiban, line 420: We might need to refactor this code into an helper function that does the git init for us.

I expect that this is needed in other tests.

We can leave this as it is for now.


There are numerous bad practices in the towncrier tests.
It would be nice to have them fix, but it should not be a burder for new contributors.

src/towncrier/test/test_builder.py
```diff
@@ -52,7 +52,7 @@ def test_non_numeric_ticket_with_extension(self):
     def test_dots_in_ticket_name(self):
         self.assertEqual(
             parse_newfragment_basename("baz.1.2.feature", ["feature"]),
-            ("2", "feature", 0),
+            ("baz.1.2", "feature", 0),
```
adiroiban, line 63: Again a bad test practice.
A test without a docstring.
But this is a nice test, as this specific feature is tested in isolation.

To me, this looks like a regression and this line should not be change.

But the docstring should be changed

```python
def test_dots_in_ticket_name(self):
    """
    The ticket ID can also be extracted from the last
    integer value found in the file name.
    """
```

src/towncrier/test/test_builder.py
```diff
@@ -81,3 +81,29 @@ def test_strip_with_counter(self):
             parse_newfragment_basename("  007.feature.3", ["feature"]),
             ("7", "feature", 3),
         )
+
+    def test_orphan(self):
```
adiroiban, line 98: many thanks for the new tests.
I think they are very useful.

I think that for the new tests we should follow best practice and alwasy add a docstring explaining the purpose/scope/scenario of the test.

Can you please add docstring to these new tests? Thanks

adiroiban (changes requested):
Thanks for the PR.
I did a review now as I had some time.

This needs changes due to `test_dots_in_ticket_name`

This change introduces a regression.

If we decide that `test_dots_in_ticket_name` was a bug, we need to add a `.removal` news fragment for this PR to let others know that this use case is no longer supported and that it was removed.

But, it looks like `test_dots_in_ticket_name` is indeed a feature... it's just that it was not documented.

So we need to decide if we want to keep the undocumented feature, add documentation for it, or plan to remove this feature.

I hope that we can keep this feature.

src/towncrier/test/test_build.py
mdellweg: As per my comment above, I think this snippet was parsed wrongly. But I cannot find anything in the docs about it the only reference is the comment (_builder.py):
```
# something-cool.feature.ext for projects that don't use ticket
```

And I do not think that in that case "3" was meant to be an issue number.

mdellweg: Also I read the comment you cited as the "name of the newsfragment" being `fix-1.2`.

.gitignore
mdellweg: I guess it was before me reading the CONTRIBUTOR docs. I tried running the tests with pytest.

src/towncrier/test/test_build.py
mdellweg: How about this?

src/towncrier/test/test_builder.py
```diff
@@ -8,63 +8,73 @@
 
 class TestParseNewsfragmentBasename(TestCase):
     def test_simple(self):
+        """Plain <number>.<category> should be valid with counter=0."""
```
adiroiban: We can leave these comments.

Just a note, if you have time to make them better



I recommend this read ... is not log https://jml.io/test-docstrings/

Maybe
```suggestion
        """
        <number>.<category> generates a counter value of 0.
        """
```

src/towncrier/test/helpers.py
```diff
@@ -69,3 +70,43 @@ def setup_simple_project(
 
     if mkdir_newsfragments:
         Path("foo/newsfragments").mkdir()
+
+
+def with_isolated_simple_project(
```
adiroiban: Thanks for the refactoring.

I know that we already have `setup_simple_project`

The test decorators are all about isolation.

And not sure if "simple" has value.

Do we also have "with_isolated_complex_project" ?

```suggestion
def with_project(
```

Can we have a docstring for this function?

I think that in the docstring we can have more details about what is considered "simple".

My main goal for the docstring is to describe the intended use cases for each argument.

With that, you can decide if you need to extend this function, or create a new one (that for example is more complex)

src/towncrier/test/test_build.py
```diff
@@ -15,44 +15,50 @@
...
-            with open("foo/newsfragments/README", "w") as f:
-                f.write("Blah blah")
-            # And files that don't have a valid category
-            with open("foo/newsfragments/README.rst", "w") as f:
-                f.write("**Blah blah**")
-
-            result = runner.invoke(command, ["--draft", "--date", "01-01-2001"])
+    @with_isolated_simple_project()
+    def _test_command(self, command, runner):
+        # Off the shelf newsfragment
```
adiroiban: What does this comment mean ?

src/towncrier/test/helpers.py
```diff
@@ -69,3 +70,43 @@ def setup_simple_project(
...
+    *,
+    with_git=False,
```
adiroiban: I am not a big fan of flag arguments.

I prefer something like `@with_project` and `@with_git_project` . but we can have this like this, as long as the intended usage of `with_git` is documented.

src/towncrier/test/helpers.py
```diff
@@ -69,3 +70,43 @@ def setup_simple_project(
...
+    config: str | None = None,
```
adiroiban: maybe config can be dedented by default, so that we don't have to always call dedent.

By forcing the usage of dedent, it should make the tests easier to read.
We no longer have to do list concatenation

src/towncrier/test/helpers.py
```diff
@@ -69,3 +70,43 @@ def setup_simple_project(
...
+    extra_config: str = "",
```
adiroiban: Just a comment. Not sure if it make sense to keep this.

I find it easier to read a test, if it has the full configuration.

adiroiban (changes requested):
Changes look very good. Many thanks.

I think that the only major comment is the missing release note for the `fix.1.2.3.bugfix` change.

We should add a note about this in the relese note, so that people who were accidentally relying on this bug, should know about it and switch to orphan marker.

The others are just minor comments. Nothing to block the merge.

Thanks again!

mdellweg:
I'm going to separate the test rewrite into a different PR for easier review.

adiroiban:
> I'm going to separate the test rewrite into a different PR for easier review.

That would be awesome. Much appreciated.

src/towncrier/test/test_build.py
mdellweg: I consider this the archetype of a newsfragment. (One line down; this is not a docstring...) How would you call it?

src/towncrier/newsfragments/562.removal
```diff
@@ -0,0 +1 @@
+Previously, fragment filenames like ``fix-1.2.3.feature`` identified as closing issue ``3``. Now the identifier of newsfragment is presented as ``fix-1.2.3``.
```
adiroiban: Since this "feature" was not documented,
I think that we can consider it a bugfix.
So this filename can be `562.1.bugfix`

I think is best to start with how it works now.

Not sure about ticket vs issue.

Just a suggestion. Feel free to keep your content.

```suggestion
Fragments with filenames like `fix-1.2.3.feature` are now associated with the ticket `fix-1.2.3`.
In previous versions they were incorectly associated to ticket `3`.
```

src/towncrier/newsfragments/562.bugfix
```diff
@@ -0,0 +1 @@
+Parsing news fragments is now more robust in the face of non numeric fragment names containing numbers and dots.
```
adiroiban: Maybe something like this.

```suggestion
Orphan newsfragments containing numberic values are no longer accidentally associated to tickets. In previous verions they orphan marker was ignored and the newsfragment was associated to a ticket having the last numerical value from the filename.
```

adiroiban (approved):
Looks good. Thanks.

I am not sure we have a removal here.
I think there are just 2 bugfixes.

src/towncrier/newsfragments/562.removal
mdellweg: > So this filename can be `562.1.bugfix`

Pretty sure it's `562.bugfix.1` .

mdellweg:
I believe i updated the fragments as you wanted (tried to fix any typo found...).

adiroiban:
Many thanks. Great work. I have merged it and I will try to get a new release

mdellweg:
Thank you for the quick reviews!

PR #565, opened 2023-11-07
Refactored test_build.py

mdellweg:
Made use of more generic with_[git_]project decorators.

# Description

# Checklist

### DIFF
--- a/src/towncrier/test/helpers.py
+++ b/src/towncrier/test/helpers.py
@@ -5,6 +5,7 @@ import textwrap
 
 from functools import wraps
 from pathlib import Path
+from subprocess import call
 from typing import Any, Callable
 
 from click.testing import CliRunner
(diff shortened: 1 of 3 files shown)


src/towncrier/test/helpers.py
```diff
@@ -62,10 +63,78 @@ def setup_simple_project(
...
+
+def with_project(
+    *,
+    config: str | None = None,
+    pyproject_path: str = "pyproject.toml",
+) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
+    """Decorator to run a test with an isolated directory containing a simple
+    project.
+    Use `config` to tweak the content of the config file.
+    Use `pyproject_path` to tweak the location of the config file.
```
adiroiban: Maybe

very minor comment.

```suggestion
    """Decorator to run a test with an isolated directory containing a simple
    project.

    The files are not managed by git.

    `config` is the content of the config file.
    It will be automatically dedented.
    
    `pyproject_path` is the path where to store the config file.
```

src/towncrier/test/test_build.py
```diff
@@ -15,50 +14,48 @@
...
-                f.write("An orphaned feature ending with a dotted number")
-            # Towncrier ignores files that don't have a dot
-            with open("foo/newsfragments/README", "w") as f:
-                f.write("Blah blah")
-            # And files that don't have a valid category
-            with open("foo/newsfragments/README.rst", "w") as f:
-                f.write("**Blah blah**")
-
-            result = runner.invoke(command, ["--draft", "--date", "01-01-2001"])
+    @with_project()
```
adiroiban, line 23: happy to see the reduction in code indentation/nesting :)

src/towncrier/test/test_build.py
```diff
@@ -157,7 +152,12 @@ def test_in_different_dir_config_option(self, runner):
         self.assertEqual(0, result.exit_code)
         self.assertTrue((project_dir / "NEWS.rst").exists())
 
-    @with_isolated_runner
+    @with_project(
```
adiroiban, line 155: this is so much easier to read. thanks a lot!

src/towncrier/test/test_build.py
```diff
@@ -958,123 +893,115 @@ def do_build_once_with(version, fragment_file, fragment):
...
-            output = read("NEWS.rst")
+        self.assertEqual(0, result.exit_code, result.output)
+        output = read("NEWS.rst")
 
         self.assertEqual(
             output,
-            """
-foo 7.8.9 (01-01-2001)
-======================
+            dedent(
```
adiroiban, line 938: just asking. Was this change needed? was the test failing with the old code?

For assertion, I prefer to have as little processing as possible.

But we can also have this code... no problem.

adiroiban (approved):
Looks good. Only minor comments. Much appreciated.

I did a quick review between a meeting and a windows server core image build :)

src/towncrier/test/test_build.py
mdellweg: Not specifically needed. But this way it aligns better with the other tests doing assertions the same way.

adiroiban: Ok. We can merge this. No problem.

adiroiban:
It looks good. I don't have time to look into the details... but I think that we are fine.
It's better than we have in trunk, the tests pass, so it shoudl be good to merge...  the bar is low, I know, but this is what we can do now.

PR #568, opened 2023-11-11
Add nox as development dependency

iliakur:
# Description

When I set up my development env I had to manually add `nox`. This should install it automatically.

# Checklist


adiroiban (approved):
Thanks. Good catch. I have created the missing fragment. Should be ready to merge.

PR #577, opened 2024-02-10
Add explicit encoding to read_text.

jaraco:
Closes #561

# Description

# Checklist

* [x] Make sure changes are covered by existing or new tests.
* [x] For at least one Python version, make sure local test run is green.
* [x] Create a file in `src/towncrier/newsfragments/`. Describe your
  change and include important information. Your change will be included in the public release notes.
* [x] Make sure all GitHub Actions checks are green (they are automatically checking all of the above).
* [x] Ensure `docs/tutorial.rst` is still up-to-date.
* [x] If you add new **CLI arguments** (or change the meaning of existing ones), make sure `docs/cli.rst` reflects those changes.
* [x] If you add new **configuration options** (or change the meaning of existing ones), make sure `docs/configuration.rst` reflects those changes.

Issue #561, opened 2023-11-04
EncodingWarning when building


jaraco:
While building using towncrier with PYTHONWARNDEFAULTENCODING=1, I see this warning:

```
/Users/jaraco/code/jaraco/jaraco.functools/.tox/finalize/lib/python3.11/site-packages/towncrier/build.py:169: EncodingWarning: 'encoding' argument not specified
  resources.files(config.template[0]).joinpath(config.template[1]).read_text()
```

One needs to explicitly pass `encoding='utf-8'` to `.read_text()` to avoid these warnings. See [PEP 597](https://peps.python.org/pep-0597/) for details.

adiroiban:
Thanks. Good call.

I think that `trial` is missing a helper to catch generic warnings... but maybe we can have a quick hack and run our CI with the warnings enabled and just fail somehow is a warning is raised.

But event a PR with just the explicit encoding, and without any extra warning detection, is good enough.

Happy to review a PR.

PR #578, opened 2024-02-10
Enable reporting of encoding warnings.

jaraco:
# Description
As suggested in #561 and #577, enable reporting of EncodingWarnings per PEP 597.

This PR is proposed separately from #577, because #577 only fixes the warnings in the main codebase and not in the tests. 80-some warnings remain to be addressed, so I leave it to the maintainers to determine when to add this check.

# Checklist


adiroiban:
Thansk for the PR... sorry for missing it.

Feel free to ping me if a PR is not reviewed.

I sent invitation to you and @SmileyChris  to towncrier contributors team... if a PR is approved, you can trigger the merge.

adiroiban (approved):
Thanks.

We can have it merged.

The current change will not do much in our CI... but I hope that it will make problem more visible to the current dev team :)

I think that this is something that we might to have in trial.

Trial should capture the warnings and fail on warnings

adiroiban:
I have enabled auto-merge for this. I hope it will land soon. Thanks again!

PR #587, opened 2024-04-15
fix: Also render title in markdown template without name

infinisil:
# Description
Previously, if there was no project name defined, the default markdown template wouldn't render the title properly due to a missing `#`. 

# Checklist

### DIFF
--- a/src/towncrier/newsfragments/587.bugfix
+++ b/src/towncrier/newsfragments/587.bugfix
@@ -1,0 +1,1 @@
+The default Markdown template now renders a title containing the release version and date, even when the `name` configuration is left empty.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -1251,6 +1251,53 @@ class TestCli(TestCase):
     @with_project(
         config="""
         [tool.towncrier]
+        name = ""
+        directory = "changes"
+        filename = "NEWS.md"
+        version = "1.2.3"
+        """
+    )
+    def test_markdown_no_name_title(self, runner):
+        """
+        When configured with an empty `name` option,
+        the default template used for Markdown
+        renders the title of the release note with just
+        the version number and release date.
+        """
+        write("changes/123.feature", "Adds levitation")
+        write(
+            "NEWS.md",
+            contents="""
+                A line
+
+                <!-- towncrier release notes start -->
+            """,
+            dedent=True,
+        )
+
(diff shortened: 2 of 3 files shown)


src/towncrier/test/test_build.py
```diff
@@ -1248,6 +1248,49 @@ def test_default_start_string_markdown(self, runner):
 
         self.assertEqual(expected_output, output)
 
+    @with_project(
+        config="""
+        [tool.towncrier]
+        directory = "changes"
```
adiroiban, line 1255: Maybe have explcicit empty configuration ?

```suggestion
        name = ""
        directory = "changes"
```

src/towncrier/test/test_build.py
```diff
@@ -1248,6 +1248,49 @@ def test_default_start_string_markdown(self, runner):
...
+        filename = "NEWS.md"
+        version = "1.2.3"
+        """
+    )
+    def test_markdown_no_name_title(self, runner):
+        """
+        Markdown renders a title even when there's no project name
```
adiroiban: I think that this is a better description for the test.

```suggestion
        When configured with an empty `name` option,
        the default template used for Markdown,
        renders the title of the release note with just
        the version number and release date.
```

src/towncrier/newsfragments/587.bugfix
```diff
@@ -0,0 +1 @@
+When using the default Markdown template, render a title even without a `name`
```
adiroiban: I think that your release notes sound like a commit message and less as a public release announcement.

What is `name` ?

Maybe something like this? Feel free to modify my suggestion.

```suggestion
The default Markdown template now renders a title containing the release version and date, even when the `name` configuration is left empty. 
```

adiroiban (approved):
Many thanks for the changes.

They look good.

I left a few minor comments.

I hope we can have those improvemnets before the merge.

Thanks again!

src/towncrier/test/test_build.py
infinisil: Why that? It doesn't seem to be relevant for this test

Edit: https://github.com/twisted/towncrier/pull/587#discussion_r1567701470 explains it, never mind!

infinisil:
Great suggestions, thanks! Applied them :D

adiroiban:
Thanks for your help. I have enabled auto-merge. I hope this will land soon.

PR #588, opened 2024-04-23
Fix incorrect handling of orphan fragment names consists of only digits

### LINKED ISSUE #584: Issue number incorrectly inferred when hash is all digits
I've been using towncrier with [mostly default configuration](https://github.com/jaraco/skeleton/blob/6ff02e0eefcd90e271cefd326b460ecfa0e3eb9e/towncrier.toml) to generate news entries. I'll sometimes give a filename with an issue number (`{issue}.{scope}.rst`) and other times use `+.{scope}.rst`, and that works well. When I supply an issue number, it's included in the changelog and when I don't, it's omitted.

However, today when using towncrier 23.11, my expectation was violated when the generated hash was all digits:

```
 backports.tarfile main @ towncrier create -c "Initial release based on tarfile from Python 3.12.2." +.removal.rst
Created news fragment at /Users/jaraco/code/jaraco/backports.tarfile/newsfragments/+96333363.removal.rst
 backports.tarfile main @ git add newsfragments; git commit -a -m "Add news fragment."
[main ec51166] Add news fragment.
 1 file changed, 1 insertion(+)
 create mode 100644 newsfragments/+96333363.removal.rst
 backports.tarfile main @ tox -e finalize
finalize: install_deps> python -I -m pip install 'jaraco.develop>=7.23' towncrier
finalize: commands[0]> python -m jaraco.develop.finalize
Loading template...
/Users/jaraco/code/jaraco/backports.tarfile/.tox/finalize/lib/python3.12/site-packages/towncrier/build.py:169: EncodingWarning: 'encoding' argument not specified
  resources.files(config.template[0]).joinpath(config.template[1]).read_text()
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
Removing the following files:
/Users/jaraco/code/jaraco/backports.tarfile/newsfragments/+96333363.removal.rst
Removing news fragments...
Done!
[main 786b22e] Finalize
 2 files changed, 7 insertions(+), 1 deletion(-)
 delete mode 100644 newsfragments/+96333363.removal.rst
  finalize: OK (5.36=setup[4.58]+cmd[0.78] seconds)
  congratulations :) (5.42 seconds)
 backports.tarfile main @ head NEWS.rst
v1.0.0
======

Deprecations and Removals
-------------------------

- Initial release based on tarfile from Python 3.12.2. (#96333363)

```

It seems that if the hex hash happens not to have any digits greater than 9, the hash is inferred to be an issue number. That feels like a bug, since the same behavior would not produce an issue number if there were hex digits greater than 9. Perhaps towncrier should provide a way to distinguish between a hash and a typical issue number.

SmileyChris:
Fixes #584

# Checklist

### DIFF
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -37,7 +37,8 @@ def parse_newfragment_basename(
             # something-cool.feature.ext for projects that don't use ticket
             # numbers in news fragment names.
             ticket = ".".join(parts[0:i]).strip()
-            # If the ticket is an integer, remove any leading zeros.
+            # If the ticket is an integer, remove any leading zeros (to resolve
+            # issue #126).
             if ticket.isdigit():
                 ticket = str(int(ticket))
             counter = 0
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -13,15 +13,6 @@ from typing import Any, DefaultDict, Iterable, Iterator, Mapping, Sequence
 from jinja2 import Template
 
 
-def strip_if_integer_string(s: str) -> str:
-    try:
-        i = int(s)
-    except ValueError:
-        return s
-
-    return str(i)
-
-
 # Returns ticket, category and counter or (None, None, None) if the basename
 # could not be parsed or doesn't contain a valid category.
 def parse_newfragment_basename(
@@ -45,7 +36,10 @@ def parse_newfragment_basename(
             # NOTE: This allows news fragment names like fix-1.2.3.feature or
             # something-cool.feature.ext for projects that don't use ticket
             # numbers in news fragment names.
-            ticket = strip_if_integer_string(".".join(parts[0:i]))
+            ticket = ".".join(parts[0:i]).strip()
+            # If the ticket is an integer, remove any leading zeros.
+            if ticket.isdigit():
+                ticket = str(int(ticket))
             counter = 0
             # Use the following part as the counter if it exists and is a valid
             # digit.
--- a/src/towncrier/newsfragments/588.bugfix
+++ b/src/towncrier/newsfragments/588.bugfix
@@ -1,1 +1,1 @@
-Fix orphans consisting of only digits (e.g. '+12345678.feature') losing their orphan status.
+Orphan news fragments, fragments not associated with an issue, consisting of only digits (e.g. '+12345678.feature') now retain their leading marker character.
--- a/src/towncrier/newsfragments/588.bugfix
+++ b/src/towncrier/newsfragments/588.bugfix
@@ -1,0 +1,1 @@
+Fix orphans consisting of only digits (e.g. '+12345678.feature') losing their orphan status.
--- a/src/towncrier/test/test_builder.py
+++ b/src/towncrier/test/test_builder.py
@@ -125,3 +125,10 @@ class TestParseNewsfragmentBasename(TestCase):
             parse_newfragment_basename("+orphan_12.3.feature", ["feature"]),
             ("+orphan_12.3", "feature", 0),
         )
+
+    def test_orphan_all_digits(self):
+        """Orphaned snippets can consist of only digits."""
+        self.assertEqual(
+            parse_newfragment_basename("+123.feature", ["feature"]),
+            ("+123", "feature", 0),
+        )


SmileyChris:
Looking closer the true source of the error is because `strip_if_integer_string` reads the orphan of `+12345` as an actual integer, stripping off the '+', causing this issue to lose its status as an orphan!

SmileyChris:
There, that's tidier!

src/towncrier/_builder.py
```diff
@@ -45,7 +36,10 @@ def parse_newfragment_basename(
             # NOTE: This allows news fragment names like fix-1.2.3.feature or
             # something-cool.feature.ext for projects that don't use ticket
             # numbers in news fragment names.
-            ticket = strip_if_integer_string(".".join(parts[0:i]))
+            ticket = ".".join(parts[0:i]).strip()
+            # If the ticket is an integer, remove any leading zeros.
+            if ticket.isdigit():
```
adiroiban, line 42: just a drive-by comment, for reference

I see this was done due to this report https://github.com/twisted/towncrier/issues/126

someone is creating files as `0020.bugfix` , but they expect towcrier to render them as `20` 

... not sure if this feature was needed... but we are stuck with it.

src/towncrier/newsfragments/588.bugfix
```diff
@@ -0,0 +1 @@
+Fix orphans consisting of only digits (e.g. '+12345678.feature') losing their orphan status.
```
adiroiban: This message sound more like a git comment, and less like a release note.

We might need a terminology glossary  in the documentation.


I never felt the need for this feature.
I alwasy have to check, what orphan fragments are :)


 ```suggestion
Orphans fragments, fragments without an associated URL, consisting of only digits (e.g. '+12345678.feature') now retain the leading marker character.
```

**Review by adiroiban — APPROVED** [review]
Looks good. Many thanks. Only a minor comment regarding the release notes.

File: src/towncrier/test/test_builder.py [hunk 3]
```diff
@@ -125,3 +125,10 @@ def test_orphan_with_dotted_number(self):
             parse_newfragment_basename("+orphan_12.3.feature", ["feature"]),
             ("+orphan_12.3", "feature", 0),
         )
+
+    def test_orphan_all_digits(self):
+        """Orphaned snippets can consist of only digits."""
+        self.assertEqual(
+            parse_newfragment_basename("+123.feature", ["feature"]),
+            ("+123", "feature", 0),
```
webknjaz, line 133: Could you add another non-orphan test with an identifier looking like a commit sha? In the templates I've started using recently, I try to output PRs/issues, commits and arbitrary identifiers separately: https://github.com/cherrypy/cheroot/blob/3591a1c/docs/changelog-fragments.d/.towncrier-template.rst.j2#L59-L63.

It'd be nice if that remained functional. Hence an explicit test...

src/towncrier/newsfragments/588.bugfix
webknjaz: @adiroiban did you mean to use plural or singular in the first word "Orphans" in your suggested change?

SmileyChris: How about "non-issue fragments"?

src/towncrier/_builder.py
SmileyChris: I threw that in the comment for posterity then.

src/towncrier/test/test_builder.py
SmileyChris: I'd say that should be a separate PR -- that's an interesting functional test but not really tied to the changes being made here.

webknjaz: The reason I brought it up is that I wasn't sure if this PR influences it. So it's kinda related as a possible regression. But I understand that it may be non-atomic and separate otherwise.

adiroiban: Thanks for your feedback. 

This test is named `test_orphan_all_digits` . Based on that, I think that `+123` is ok.

I also think that using a git commit ref is also a valid test... but it should be a separate test.

adiroiban:
It looks good. I have enabled auto-merge.

Thanks for your help.

I am happy to review and approve a PR that add more tests... but that should not be a blocker for this PR.

PR #589, opened 2024-04-24
Add docs for CLI orphan fragments

SmileyChris:
`towncrier create +.feature.rst` wasn't documented.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -79,7 +79,8 @@ If the fragments directory does not exist, it will be created.
 If the filename exists already, ``towncrier create`` will add (and then increment) a number after the fragment type until it finds a filename that does not exist yet.
 In the above example, it will generate ``123.bugfix.1.rst`` if ``123.bugfix.rst`` already exists.
 
-To create a news fragment not tied to a specific issue, start the fragment name with a ``+``. If that is the entire fragment name, a random hash will be added for you::
+To create a news fragment not tied to a specific issue (which towncrier calls an "orphan fragment"), start the fragment name with a ``+``.
+If that is the entire fragment name, a random hash will be added for you::
 
    $ towncrier create +.feature.rst
    $ ls newsfragments/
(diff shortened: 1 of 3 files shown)


docs/cli.rst
```diff
@@ -79,6 +79,12 @@ If the fragments directory does not exist, it will be created.
 If the filename exists already, ``towncrier create`` will add (and then increment) a number after the fragment type until it finds a filename that does not exist yet.
 In the above example, it will generate ``123.bugfix.1.rst`` if ``123.bugfix.rst`` already exists.
 
+To create a news fragment not tied to a specific issue, start the fragment name with a ``+``. If that is the entire fragment name, a random hash will be added for you::
```
adiroiban: Maybe use [semantic newlines](https://sembr.org/) for the documentation

I think that this is a good place to remind what "orphan fragments" are :)

Minor suggestion

```suggestion
To create a news fragment not tied to a specific issue,
what is called an orphan fragment,
start the fragment name with a ``+``.
If that is the entire fragment name,
a random hash will be added for you::
```

adiroiban (approved):
Thanks. This is much better. Much appreciated :)

I was always confused by the orphan thing...as I don't use or need it

I think is best to also mention this behaviour in the create docstring

https://github.com/twisted/towncrier/blob/914b446e64d62f3968a97931e490b1abe5c8c581/src/towncrier/create.py#L46-L68

In this way, the info is found in `towncrier create --help`

The idea is that in the future we can just run some helper utility for click and it will automatically generate the .RST documentation for the CLI.

I think is best to keep as much info as possible into the `--help` part.

For the `--help` docstring , we can keep the info brief,
Just mention that you can use `+` as the ticket number and towncrier will auto-generate

More info and examples, can be found in the RST version.

Thanks again

docs/cli.rst
SmileyChris: Done!

adiroiban:
Thanks Chris. I have merged this. I added you to the list of maintainer, so if the future you should be able to merge a PR once it is approved.

Just make sure that any change is reviewed and approved before a merge.

Thanks again!

PR #572, opened 2023-12-20
#571 Use pre-commit dog food and update release process for pre-commit autoupdate.

adiroiban:
# Description
Fixes: #571

Try to reproduce the reported error.

Besides using `pre-commit` as part of our CI

this updates the release process to make it work with `pre-commit auto-update`

PR #590, opened 2024-04-28
Add --bleeding-edge into pre-commit autoupdate

### LINKED ISSUE #571: pre-commit autoupdate seems to fail
When I launch pre-commit autoupdate I get this

```
$ pre-commit autoupdate
[https://github.com/pre-commit/pre-commit-hooks] already up to date!
[https://github.com/psf/black-pre-commit-mirror] already up to date!
[https://github.com/adamchainz/blacken-docs] already up to date!
[https://github.com/rstcheck/rstcheck] already up to date!
[https://github.com/charliermarsh/ruff-pre-commit] already up to date!
[https://github.com/nbQA-dev/nbQA] already up to date!
[https://github.com/twisted/towncrier]
=====> /var/folders/2z/142033n17rbfy969s6h4hymw0000gn/T/tmphs0r_674/.pre-commit-hooks.yaml is not a file
```

So all other git hooks get updated, but it crashes only when trying to update this project.

I also tried to manually update the config file with what I see from your docs,

```yaml
  - repo: https://github.com/twisted/towncrier
    rev: 22.13.0
    hooks:
      - id: towncrier-check
```

but the result doesn't change.

Also `pre-commit clean` doesn't seem to help.

sadikkuzu:
# Description

Resolves #571 and #572

before:
image

after:
image

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,7 +218,7 @@ jobs:
 
     - name: Update pre-commit
       run: |
-        pre-commit autoupdate
+        pre-commit autoupdate --bleeding-edge
 
     - name: Run pre-commit
       run: |


adiroiban:
Thanks for the PR.


This is a good start, but it looks like a brute force fix for our problem.

It looks like bleeding edge just uses the latest version of any dependency.

Do we know which dependency is causing our issue.

While "bleeding-edge" might fix the current issue... I think that in the future we might end up with other transient error.

Also, I think that it's important to include the pre-commit update as part of our CI, to make sure we will not regress on this. That is, include the changes from #572 to check that this fixes the issue :)

sadikkuzu:
`pre-commit autoupdate` follows these git commands:
https://github.com/pre-commit/pre-commit/blob/main/pre_commit/commands/autoupdate.py#L42-L61 

To demonstrate so, I ran such commands:

```shell
git clone https://github.com/twisted/towncrier.git
cd towncrier
git checkout --track origin/571-pre-commit-hooks
git fetch origin HEAD --tags
git describe FETCH_HEAD --tags --abbrev=0
```

Then I got: `22.8.0`

When I go to [towncrier 22.8.0](https://github.com/twisted/towncrier/tree/22.8.0) version, there is no `.pre-commit-hooks.yaml` there.

image

sadikkuzu:
On the other hand,
[23.10.0](https://github.com/twisted/towncrier/tree/23.11.0) version has such warning message:

image

sadikkuzu:
@adiroiban hi,
Can you approve remaining workflow runs?
image

adiroiban:
> When I go to [towncrier 22.8.0](https://github.com/twisted/towncrier/tree/22.8.0) version, there is no .pre-commit-hooks.yaml there.

Ok 28.8.0 is an older version.

But https://github.com/twisted/towncrier/tree/23.11.0 has that file.

--------

> 23.10.0 version has such warning message:

We remove the branch that is used to do the release... so that we don't have too many branches in the auto-complete list

This is why you have that warning.
It should not matter.

adiroiban:
> Can you approve remaining workflow runs?

I have approved the CI.

I only now saw that this PR is targeting the ohter 571 branch... sorry, I was not expecting this... and I was just doing all kind of experiments on my branch.

sadikkuzu:
It seems ok now:
image

adiroiban:
Many thanks for the help here.

I think I got this.

So the issue is with `git describe` that things that 28.8.0 is our latest version

```
$ git describe trunk --tags
22.8.0-230-g7221d5d
```

We release towncrier from a branch/tag and not from `trunk`

We then squash the branch...and the tag commit is lost.

I think that the fix here, is to do a new towncrier release, in which we make sure we don't use a squash merge.

sadikkuzu:
Moreover, [this pre-commit step](https://github.com/twisted/towncrier/blob/571-pre-commit-hooks/.github/workflows/ci.yml#L215-L218) is not necessary:
image

Remaining next steps alone do right actions:
https://github.com/twisted/towncrier/blob/571-pre-commit-hooks/.github/workflows/ci.yml#L219-L225 👍🏼

adiroiban:
Thanks for your help and feedback.

You help was critical in finding the source of this bug.

I don't think that we should merge this PR and use bleeding edge.

Instead, we should fix the release process to add a note that release PR should not be squashed...and then do a new release without a squash.

sadikkuzu:
> Thanks for your help and feedback.
> 
> You help was critical in finding the source of this bug.
> 
> I don't think that we should merge this PR and use bleeding edge.
> 
> Instead, we should fix the release process to add a note that release PR should not be squashed...and then do a new release without a squash.

My pleasure 🙏🏼 ✨ 

Yes, we may close this PR. New release process will fix the issue. 👍🏼

sadikkuzu:
> Moreover, [this pre-commit step](https://github.com/twisted/towncrier/blob/571-pre-commit-hooks/.github/workflows/ci.yml#L215-L218) is not necessary: ...
> 
> Remaining next steps alone do right actions: https://github.com/twisted/towncrier/blob/571-pre-commit-hooks/.github/workflows/ci.yml#L219-L225 👍🏼

For this
I will propose a new pull-request later on.

adiroiban:
The idea of the CI jobs was to do a full pre-commit cycle, similar to how it is used by developers, not only the auto-update step that is needed for this bug.

sadikkuzu:
> The idea of the CI jobs was to do a full pre-commit cycle, similar to how it is used by developers, not only the auto-update step that is needed for this bug.

Then it should be `pre-commit install`, not `pre-commit` only.

https://github.com/twisted/towncrier/pull/592

SmileyChris:
Bleeding edge isn't the solution here.

adiroiban:
Please add your feedback and suggestion to the #571 PR
THanks

PR #591, opened 2024-04-28
Prefer pathlib for read/write operations

jaraco:
- **Rely on pathlib when reading and writing simple text.**
- **Universal newlines is the default, so remove the comment.**

# Description
While working on #577, I noticed it might be nice to leverage pathlib for reading and writing files. This approach avoids adding nesting and with blocks by utilizing `read_text` and `write_text`.

This PR depends on #577 and should be rebased after merging that one.

# Checklist

### DIFF
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -17,7 +17,7 @@ from typing import Any
 if sys.version_info < (3, 10):
 
     def _newline_write_text(path: Path, content: str, **kwargs: Any) -> None:
-        with path.open("w", **kwargs) as strm:
+        with path.open("w", **kwargs) as strm:  # pragma: no cover
             strm.write(content)
 
 else:
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -65,8 +65,6 @@ def _figure_out_existing_content(
         # Non-existent files have no existing content.
         return "", ""
 
-    # If we didn't use universal newlines here, we wouldn't find *start_string*
-    # which usually contains a `\n`.
     content = Path(news_file).read_text(encoding="utf-8")
 
     t = content.split(start_string, 1)
--- a/src/towncrier/_writer.py
+++ b/src/towncrier/_writer.py
@@ -37,15 +37,16 @@ def append_to_newsfile(
     if top_line and top_line in prev_body:
         raise ValueError("It seems you've already produced newsfiles for this version?")
 
-    # Leave newlines alone. This probably leads to inconsistent newlines,
-    # because we've loaded existing content with universal newlines, but that's
-    # the original behavior.
-    with news_file.open("w", encoding="utf-8", newline="") as f:
-        if header:
-            f.write(header)
+    news_file.write_text(
         # If there is no previous body that means we're writing a brand new news file.
         # We don't want extra whitespace at the end of this new file.
-        f.write(content + prev_body if prev_body else content.rstrip() + "\n")
+        header + (content + prev_body if prev_body else content.rstrip() + "\n"),
+        encoding="utf-8",
+        # Leave newlines alone. This probably leads to inconsistent newlines,
+        # because we've loaded existing content with universal newlines, but that's
+        # the original behavior.
+        newline="",
+    )
 
 
 def _figure_out_existing_content(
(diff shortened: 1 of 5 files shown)


jaraco:
Oh, this change fails on Python 3.9 and earlier due to the `newline` parameter only being added in that edition.

If this proposed transition is acceptable, I'd be tempted to wrap the offending call in a compatibility wrapper (something that would signal its removal after Python 3.9 support is dropped), but I'm not sure it's worth the effort.

jaraco:
The coverage failure doesn't make any sense:

```
Name                       Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------
src/towncrier/_writer.py      25      0     10      1    97%   20->exit
----------------------------------------------------------------------
TOTAL                        677      0    321      1    99%
```

Does `20->exit` mean that the coverage check wants to see an exception occur in [the context](https://github.com/jaraco/towncrier/blob/e5c6df77d40c8a5ae7aee7a5ac4fbb74c25a3bcd/src/towncrier/_writer.py#L20-L21) in order to check both `__exit__(None, None, None)` and `__exit__(type, val, ctx)`? That's absurd and didn't need to be checked before. Or is it a bug in coverage, where it's failing to detect that the context is exited at all (which it must for the function to return)?

src/towncrier/_writer.py
```diff
@@ -8,7 +8,22 @@
...
+import sys
+
 from pathlib import Path
+from typing import Any
+
+
+if sys.version_info < (3, 10):
+
+    def _newline_write_text(path: Path, content: str, **kwargs: Any) -> None:
+        with path.open("w", **kwargs) as strm:  # pragma: no cover
```
adiroiban: Is this an issue of line coverage or branch coverage ?

```suggestion
        with path.open("w", **kwargs) as strm:  # pragma: no branch
```

src/towncrier/_writer.py
```diff
@@ -8,7 +8,22 @@
 
 from __future__ import annotations
 
+import sys
+
 from pathlib import Path
+from typing import Any
+
+
+if sys.version_info < (3, 10):
```
adiroiban, line 17: I think that this needs a comment.
Once we no longer need to support 3.9 or older, this code can be removed.

src/towncrier/newsfragments/591.bugfix.rst
```diff
@@ -0,0 +1 @@
+Leveraged pathlib in most file operations.
```
adiroiban: It this a bugfix or just a .misc internal cleanup ?

jaraco: Good call. I'd forgotten there might be other options than the defaults. Agreed it's misc.

jaraco: Renamed in [416a2c8](https://github.com/twisted/towncrier/pull/591/commits/416a2c8fffa643a6988e3df91b97b61848045cfd).

adiroiban (approved):
Thanks. Happy to use pathlib.

I left a few minor comments.

-----------

I don't line universal newlines.

On my projects, we always use `\n` for the source code.

You don't want a source file created on Windows to conflict with Linux.

---------

So I think towncrier should just auto-detect the newlines used in the "filename" configuration and use that for "create" and other cases.

src/towncrier/_writer.py
jaraco: Fair, though in my mind, the code is pretty explicit about which versions rely on the compatibility code. I was tempted to add a comment just to explain why, but my feeling was that the why could be readily derived by removing the compatibility and testing what fails. Still, it's faster to read a comment.

Added in 8b821dc.

adiroiban:
Thanks. If you are happy with this PR, feel free to merge.

jaraco:
> Thanks. If you are happy with this PR, feel free to merge.

I would merge, but something's wrong with credentials. I'm a member of towncrier-contributors, yet I have no option to merge. I've signed out and back in but yet can't merge. It says I need write access to merge.

image

adiroiban:
Sorry. My bad. I set the wrong permissiosn for the team. I have merged it, but in the future you should be able to merge. Thanks again!

PR #592, opened 2024-04-29
Install pre-commit via pre-commit install command

sadikkuzu:
# Description

# Checklist


adiroiban:
Thanks @sadikkuzu for the fix.

I have applied your fix to #572 

If there are any changes that needs to be made please add a comment or suggestion to that PR.

Thanks again

PR #596, opened 2024-05-06
Remove github edit link in non-trunk versions of docs

SmileyChris:
# Description

In the sphinx configuration, remove github edit link in non-trunk versions of docs.

This only triggers specifically for read the docs, since it's based one of their custom environment variables.

Fixes #505

PR #597, opened 2024-05-06
Fix readthedocs to keep the edit link for `latest` in addition to `trunk`

SmileyChris:
# Description
Followon from #596
Let's make it work for the `latest` version since that's the version more likely to be read...


adiroiban (approved):
THanks. I hope this will work :)

adiroiban:
Looks good. Thanks. I can see the edit button on

* https://towncrier.readthedocs.io/en/latest/
* https://towncrier.readthedocs.io/en/trunk/

There is still an edit link on "stable" and previous tags, but this is because there were no new builds.

* https://towncrier.readthedocs.io/en/stable/

Maybe this is a good reason do make a new release

PR #601, opened 2024-05-22
Allow `towncrier` to traverse back up directories looking for the configuration file

### LINKED ISSUE #433: Add support for calling towncrier from sub-directories not only from the repo root.
I have a repo that looks like this

```
repo_root/
    - sub_project/
         - changes/ 
         - towncrier.toml
    - other_project/
```
where I want to use towncrier only in the sub_project directory (changes is where the fragments are). I've been able to create the first fragment in the default branch but when adding a new fragment `1234.feature` from a new branch and using the `check` command like

`towncrier check --compare-with origin/main ` 

this is what produces

```
Looking at these files:
----
1. /home/user/repo-root/sub_project/sub_project/changes/1234.feature
----
No new newsfragments found on this branch.

```

It has found the new fragment but doesn't recognize it as a new fragment. Also the directory that has checked 
is `sub_project/sub_project`, which it doesn't exist.

SmileyChris:
# Description
Fixes #433 

# Checklist


src/towncrier/_settings/load.py
```diff
@@ -67,7 +67,7 @@ def load_config_from_options(
 ) -> tuple[str, Config]:
     if config_path is None:
         if directory is None:
-            directory = os.getcwd()
+            return traverse_for_config(None)
```
adiroiban: I think that this should be private.

I know that this is from `_settings.load` , but I think that we should make it explicit that we don't want to expose this as a public API.

Also, there are a lot of branches here,
so I think that it would help to add an explicit comment for this branch.

With the missing docstring for the `load_config_from_options` it's a bit hard to understand the purpose of this function.

Can we have something like this.

It lools like a bit of duplication between 
 `traverse_for_config` and these following lines

```python
base_directory = os.path.abspath(directory)
config = load_config(base_directory)
```

maybe have something like this

```python
    if config_path is None:
        return _traverse_for_config(directory)
```

src/towncrier/_settings/load.py
```diff
@@ -85,6 +85,26 @@ def load_config_from_options(
...
+    """
+    start_directory = directory = os.path.abspath(path or os.getcwd())
+    while True:
+        config = load_config(directory)
+        if config is not None:
+            return directory, config
+
+        parent = os.path.dirname(directory)
+        if parent == directory:
+            raise ConfigError(
```
adiroiban, line 109: Do we have a test for this case ?

I don't see any report for missing coverate, but to me it looks like new code, without new tests :)

src/towncrier/test/test_build.py
```diff
@@ -121,11 +121,21 @@ def test_in_different_dir_dir_option(self, runner):
         self.assertEqual(0, result.exit_code)
         self.assertTrue((project_dir / "NEWS.rst").exists())
 
+    @with_project()
+    def test_traverse_up_to_find_config(self, runner):
+        """
+        If the current directory doesn't contain the configuration file, Towncrier
+        should traverse up the directory tree until it finds it.
```
adiroiban: Very minor comment - https://jml.io/test-docstrings/

```suggestion
        If the current directory doesn't contain the configuration file, Towncrier
        will traverse up the directory tree until it finds it.
```
```suggestion
        When the current directory doesn't contain the configuration file, Towncrier
        will traverse up the directory tree until it finds it.
```

adiroiban (changes requested):
Many thanks for the changes.

They look good.

I left a few minor inline comments.

The major comment is about the missing test for the case in which the config file was not found, even when traversing the whole tree.

I think that we should have such a test before merging this.

Thanks again!

src/towncrier/_settings/load.py
SmileyChris: Covered by existing `test_needs_config` test which used:
```python
        self.assertTrue(result.output.startswith("No configuration file found."))
```

src/towncrier/_settings/load.py
SmileyChris: I don't think there's any assumption that non underscored methods become automatically part of a public API is there? If people _want_ to shoot themselves in the foot, let em 🤠

SmileyChris: I've added a docstring and some comments to explain the different branches (I don't like the complexity of the towncrier's loading, but I'm not going to rewrite it just for this addition)

adiroiban (approved):
Thanks for the updates.

All good.

src/towncrier/_settings/load.py
SmileyChris: Actually, I see what you're getting at with the branching logic now. I've simplified it (and added another test).

adiroiban:
Thanks for the cleanup for `load_config_from_options`. Much appreciated.

PR #602, opened 2024-05-23
Update docs around version discovery

SmileyChris:
# Description
The tutorial text should be updated to match the version discovery changes.

# Checklist

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -56,6 +56,7 @@ Detecting Dates & Versions
 1. Manually pass ``--version=<myversionhere>`` when interacting with ``towncrier``.
 2. Set a ``version`` key in your configuration file.
 3. For Python projects with a ``package`` key in the configuration file:
+
    - install the package to use its metadata version information
    - add a ``__version__`` in the top level package that is either a string literal, a tuple, or an `Incremental <https://github.com/twisted/incremental>`_ version
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -52,7 +52,7 @@ Detecting Version
 -----------------
 
 ``towncrier`` needs to know what version your project is when generating news files.
-These are the ways you can provide it in order of precedence (with the first taking precedence over the second, and so on):
+These are the ways you can provide it, in order of precedence (with the first taking precedence over the second, and so on):
 
 1. Manually pass ``--version=<myversionhere>`` when interacting with ``towncrier``.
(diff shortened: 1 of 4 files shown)


docs/tutorial.rst
```diff
@@ -51,13 +51,16 @@ The ``.gitignore`` will remain and keep Git from not tracking the directory.
 Detecting Dates & Versions
```
adiroiban: Maybe it's best to have a separate section for version and date.

Most of the info from this section is about version.

A separate section should make it easier to jump to the information aboute date.

and I think that it only detects the current version, so not plural.

```suggestion
Detecting version
```

docs/tutorial.rst
```diff
@@ -51,13 +51,16 @@ The ``.gitignore`` will remain and keep Git from not tracking the directory.
...
 --------------------------
 
-``towncrier`` needs to know what version your project is, and there are two ways you can give it:
+``towncrier`` needs to know what version your project is. These are the ways you can provide it (and their order of precedence):
```
adiroiban: I think it's best to use [Semantic newlines](https://sembr.org/) for documentation content.

Maybe also update this page
https://towncrier.readthedocs.io/en/latest/contributing.html

I guess that not every operaition needs the version.
I don't expect to requiere the version for a `create` command.

It's not 100% clear what's the order of precendece.

```suggestion
Most ``towncrier`` commands, like `build` or `check`, needs to know what version your project is.
These are the ways you can provide it and their order of precedence.
A lower precedence number will overwrite any other method with a higher number:
```

docs/tutorial.rst
```diff
@@ -51,13 +51,16 @@ The ``.gitignore`` will remain and keep Git from not tracking the directory.
...
 
-- For Python 2/3-compatible projects, a ``__version__`` in the top level package.
-  This can be either a string literal, a tuple, or an `Incremental <https://github.com/twisted/incremental>`_ version.
-- Manually passing ``--version=<myversionhere>`` when interacting with ``towncrier``.
+1. Manually pass ``--version=<myversionhere>`` when interacting with ``towncrier``.
+2. Set a ``version`` key in your configuration file.
```
adiroiban:
```suggestion
2. Set a value for the ``version`` option in your configuration file.
```

adiroiban (approved):
Many thanks. Looks good.

My bad for not requesting this documentation updates as part of #502 

Only minor comments.

PR #603, opened 2024-05-23
Section aware create

SmileyChris:
# Description
`towncrier create` should work better with sections if defined in the configuration file.

This adds a new `towncrier create --section <section name>` option, or if you're using the new interactive create then you'll be asked for the section if there are multiple ones defined in your config:

```console
$ towncrier create --content "Sections!"
Pick a section
 1. Backend
 2. Frontend
Section (1, 2) [1]: 2
Issue number (`+` for none): 603
Fragment type (feature, bugfix, doc, removal, misc): feature
Created news fragment at /home/chris/demo/changes/frontend/603.feature.rst
```

If you use sections and none have an empty path, you must now specify the section when creating a new news fragment (if one _does_ have an empty path, that section will be used by default).

This PR also deduplicates the code that determines the base fragment path which was done in three places!

# Checklist


adiroiban:
Thanks for the PR.
I am not using the `create` command, so there is not much UI/UX feedback I can provide.

As long as you are happy with the behaviour, we can merge it.

----------

I am tring to understand what "sections" is in the context of towncrier.

I am reading something about it here https://towncrier.readthedocs.io/en/latest/configuration.html#sections

Especially the part from `Section Path Behaviour` is confusing.

I think that it needs better example.

For example

> If `directory` is not set, “newsfragments” is added to the end of each path. For example, with the above sections, the paths would be:

Which `directory` , I guess that this is `[tool.towncrier] -> directory`

Are people using this behaviour ?

To me, this all seems very confusing.

docs/cli.rst
```diff
@@ -98,6 +98,11 @@ If that is the entire fragment name, a random hash will be added for you::
    Whether to start ``$EDITOR`` to edit the news fragment right away.
    Default: ``$EDITOR`` will be started unless you also provided content.
 
+.. option:: --section SECTION
+
+   The section to use for the news fragment.
+   Default: the first section with no path.
```
adiroiban: I don't fully understand the sections... but can you have *multiple sections* without a path ?

src/towncrier/_builder.py
```diff
@@ -69,25 +100,21 @@ def parse_newfragment_basename(
 # Also returns a list of the paths that the fragments were taken from.
 def find_fragments(
     base_directory: str,
-    sections: Mapping[str, str],
-    fragment_directory: str | None,
-    frag_type_names: Iterable[str],
-    orphan_prefix: str | None = None,
+    config: Config,
```
adiroiban, line 104: I think this change is on.

I am not sure what we consider public or private API.

I can do something like this, which might suggest that this is a public API.

```
 $ python 
Python 3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0] on linux
>>> import towncrier
>>> from towncrier.check import find_fragments
```

this is one reason, why I think that we should always prefix private code with an underline.

It make it easier to check with code is public and which is not.


My suggestion is to rename this to `_find_fragments` to make it clear that it should not be used, outside of `towncrier`

src/towncrier/_builder.py
```diff
@@ -53,6 +55,35 @@ def parse_newfragment_basename(
         return invalid
 
 
+class FragmentsPath:
```
adiroiban, line 59: Is this expected to be public API?

I can do something like this: `from towncrier.create import FragmentsPath`

src/towncrier/create.py
```diff
@@ -134,19 +182,16 @@ def __main(
...
-        fragments_directory = os.path.abspath(
-            os.path.join(
-                base_directory,
-                config.package_dir,
-                config.package,
-                "newsfragments",
+    if not section:
+        if default_section is None:
+            raise click.UsageError(
+                "Multiple sections defined in configuration file, all with paths."
```
adiroiban: This looks like a big inconvenient 

Can we just fallback to the first section ?

It will act as a kind of default.

Again, I don't use the `create` command.

So maybe this suggestion makes no sense :)

--------

Also, maybe we can have messages without `please` - https://developers.google.com/style/tone#politeness-and-use-of-please

It's just extra text to read ... keep it short and to the point :)

src/towncrier/newsfragments/603.feature.rst
```diff
@@ -0,0 +1,4 @@
+The ``towncrier create`` action now uses sections defined in your config (either interactively, or via the new ``--section`` option).
+
+If you use sections and none have an empty path, you must now specify the section when creating a new news fragment.
```
adiroiban: this looks like it's breacking backward compatibility ?

Can we continue to have the old behaviour for the case when `section` is not specified?

--------

If this change is needed, then this text should not be placed in a `.feature` fragment

Instead, it should be in an `.removal` to be highlighted as part of "breaking changes"

src/towncrier/test/test_create.py
```diff
@@ -415,6 +415,113 @@ def test_without_filename_no_orphan_config(self, runner: CliRunner):
         with open(expected) as f:
             self.assertEqual(f.read(), "Edited content\n")
 
+    @with_isolated_runner
+    def test_sections(self, runner: CliRunner):
```
adiroiban, line 419: Can we have a docstring for this test desribing what is the purpose of this test.

The thing is that we don't have tests for the test code,
so a human reading the code is the only check that we have to make sure the test code is correct.

A docstring should help check and validate that the test steps are correct.

Especially in this case, `test_sections` is very generic.
It's hard to tell what part if the `create sections` functionality is expected to be tested here.

Same comment for the other tests.

adiroiban (changes requested):
Many thanks for the PR. It looks very good.

I left an initial round of feedback.

I have never used the `create` command .

Also, for my privarte project, I am not using the "section" functionality.

I know that "sections" are used in `twisted/twisted` without a path ... and that's all my experience with it.

In `twisted/twisted` there is a single section without a path.

https://github.com/twisted/twisted/blob/c465c460c68bec3cf481c2815d1039da6796eb03/pyproject.toml#L215-L245


The major concert is the backward incompatible change for the `create` CLI

I don't know why you would want multiple sections without a path.... so maybe this is not a big deal, and it's ok to have the current behaviour.

I would still document this as `.removal`.

docs/cli.rst
SmileyChris: It wouldn't make sense and wouldn't work.
It should be considered invalid to have multiple sections with the same path (even the same path of `""`). No need to add complexity to this ticket, but the config warning of that might be worthy of a separate issue.

src/towncrier/_builder.py
SmileyChris: If this is the case, there is a *lot* of methods that we should rename. The public API in my opinion, is anything we have documented. It shouldn't matter if it's underscored or not -- if you're using an undocumented method of a project, I don't see why you should assume it should continue working.

Maybe we need to have an official agreement on this and document the policy?

docs/cli.rst
SmileyChris: Opened #606

src/towncrier/_builder.py
SmileyChris: Opened #605 for discussion.

src/towncrier/newsfragments/603.feature.rst
SmileyChris: My reasoning behind it **not** being backwards incompatible is that if you didn't have any empty-path sections, then `create` wouldn't have been working for you in the past anyway

I'd be fine moving this to removal if it's actually considered backwards incompatible. Or maybe the second paragraph is just confusing and redundant since it was behaviour that didn't work previously anyway.

src/towncrier/_builder.py
adiroiban: I agree that there are a lot of methods... but at least we should try no to add more :)

Feel free to leave this as it is.


I will add more info on #605

docs/cli.rst
adiroiban: Agree that we can have this on #606 

So, maybe we should not add any comment about the default behaviour here.

src/towncrier/newsfragments/603.feature.rst
adiroiban: If it was not working before, I think that we can just remove this info.
As part of #606 we can find document the default section.

Thanks

SmileyChris: Removed, I've gone with the concept of a default section and this isn't necessary to say here.

src/towncrier/create.py
SmileyChris: Now using the first section with an empty path as the default, or if all have paths, then the first defined one.

adiroiban (approved):
Thanks all good. I saw the auto-merge, so I hope this will be merged.

Issue #600, opened 2024-05-21
The `towncrier-check` pre-commit hook doesn't actually run `towncrier check`


SmileyChris:
This seems an unnatural correlation and I'm not sure what the purpose of the current hook (`towncrier --draft`) would really even be!

Also, it would make sense that this always runs, not just when a file is found in `newsfragments/` since the check is supposed to be checking whether or not there are actually changelog updates, correct?

adiroiban:
I am using `towncrier --draft` from my project as part of CI to make sure there is no invalid RST syntax in those files.

adiroiban:
I am not using pre-commit hooks for my projects ... so I don't know how and why they should be used.

I think that we should start fixing this issue by documenting the pre-commit hooks so that we have the information of what we want to achieve with the hook

--------

Things that I check for a branch / PR:

* make sure every branch/PR has a newsfragment file
* make sure the compiled news file is valid RST
* for a release PR make sure there are no newsfragment files 

Things that are needed and not yet implemented 
* [add support for reversal branches #598](https://github.com/twisted/towncrier/issues/598)

SmileyChris:
pre-commit is useful to enforce standards. Since it needs active buy-in from your developers to actually use it, it usually lives as a git action too to actually enforce it :)

From the things you check for, I'd say we should have different hooks that do each of those:
* `check` to actually check for newsfragments (since we have a towncrier command of the same name). 
  - default `stage` should be `pre-push`
  - should use [always_run](https://pre-commit.com/#hooks-always_run)
* `validate` which ensures the validity of the compiled news file (and maybe fragments too) This should be an extension of `towncrier check` tbh.
  - probably `types: ['text']`
* I wonder if the release PR check should also be a new argument to `towncrier check` ensuring there are no loose fragments? This probably isn't that useful as a hook, since you can't tie hooks to a specific branch merge.

adiroiban:
So far the approach is to have a single `towncrier check` that does everything.

`towncrier check` should be smart enough to detect when we have a normal branch, a release branch or a revert branch, and do its magic.

SmileyChris:
Since this discussion was getting a bit esoteric, I thought it would be good to actually just suggest the changes to the current hooks. The current `towncrier-check` hook is confusing since all it does is `build --draft` when we have a check action.

krpatter-intc:
[...]

PR #604, opened 2024-05-24
Pre-commit hook suggestions

SmileyChris:
# Description
As discussed in #600, here are my suggestions on the pre-commit changes.

* renaming `towncrier-check` -> `towncrier-draft`. It is set to trigger if any text files have changed.
* adding `towncrier-checks` (which does `towncrier check`, but keeping the name different so it doesn't clash with the renamed previous hook). It is set to always trigger before `git push`.
* update `towncrier-update` (this does `towncrier build` but I don't think the name needs changing) to only trigger by the `manual` stage by default (as I doubt you'd want to do this every time you commit...)

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,16 +1,27 @@
-- id: towncrier-check
-  name: towncrier-check
+- id: towncrier-draft
+  name: towncrier build --draft
   description: Check towncrier changelog updates
-  entry: towncrier --draft
+  language: python
+  entry: towncrier build --draft
   pass_filenames: false
   types: [text]
-  files: newsfragments/
+
+- id: towncrier-checks
+  name: towncrier check
+  description: Check for missing news fragments
   language: python
-- id: towncrier-update
-  name: towncrier-update
-  description: Update changelog with towncrier
-  entry: towncrier
+  entry: towncrier check
   pass_filenames: false
-  args: ["--yes"]
-  files: newsfragments/
+  stages:
+    - pre-push
+  always_run: true
+
+- id: towncrier-update
+  name: towncrier build
+  description: Update changelog with towncrier
   language: python
+  entry: towncrier build --yes
+  pass_filenames: false
+  always_run: true
+  stages:
+    - manual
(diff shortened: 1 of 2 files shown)


adiroiban:
Thanks for the PR.

I have never used to pre-commit hooks for my projects.

I have custom GitHub actions steps to call `towncrier` as part of a PR branch protection checks and I always do both check and draft.

And draft is piped to GitHub Action summary , for easy review... just like we have for towncrier itself

https://github.com/twisted/towncrier/actions/runs/9216320582#summary-25356369369

I am -1 for having a separate hook for `check and draft` 

Don't worry to much. As I mentioned, I am not using pre-commit hooks... so I don't care too much.

But I think that it would help to get feedback for more people. 
So maybe, just as a review to @twisted/towncrier-contributors 

--------------

My only important comment is about backward compatibility.

I see the [test pass](https://github.com/twisted/towncrier/actions/runs/9216721691/job/25357527992?pr=604#step:7:1). So the change should be ok.

I think that for backward compatibility testing, it's important to keep this check, pinned at the current version

https://github.com/twisted/towncrier/blob/15d1e25d9dc676cd85fe1488341b272f6bb0e836/.pre-commit-config.yaml#L37-L40

---------

I guess that this PR will update .pre-commit-config.yaml to add the news hooks.

-------

I have never used `towncrier-update` .

Does this hook make sense ?

SmileyChris:
Personally I think the only sensible hook is one that does the equivalent of `towncrier check`. It makes sense to have a pre-merge git hook to check for the presence of a news fragment.

My suggestions here were to add that one, but keep the others and just better document them (and have more sensible naming and better defaults). The only other related change that may make sense is keeping `towncrier-check` around and just have it raise a deprecation warning.

If we think there's no use to them, I'm fine "burning the chaff" too.

adiroiban:
It ok to have them. No problem.

I am not a big user of pre-commit , so anything is ok for me.

I am not -1.

We can have this PR in review for one week, and if you are happy with the hooks and nobody is agains the changes, we can merge it.

Thanks

dorschw:
Hey, this can be useful for us, any ETA for a merge?  (I saw the discussion)

adiroiban:
Hi @dorschw this PR is still draft.... and currenlty inactive.

If you have time, you can try to create a PR with the kind of pre-commit hooks that you would like to have in `towncrier`.

Cheers

Issue #2, opened 2015-12-27
package_dir default not allowed


johnwlockwood:
The comment over package_dir in the ini section of the readme says default is current directory, but If I remove package_dir line, I get an error when running the script.

Issue #1, opened 2015-12-27
Initial; get it basically working, get it on PyPI.


hawkowl:

hawkowl:
It's on PyPI now! https://pypi.python.org/pypi/towncrier

PR #608, opened 2024-06-05
Refactor issue_key function to sort issues in a human-friendly way

SmileyChris:
# Description

In the issue key sorting method, there's this comment:
```python
        # Maybe we should sniff strings like "gh-10" -> (10, "gh-10")?
```

Let's do that.

To be more backwards compatible, we'll continue putting just plain numbers last. In reality, it's very unlikely a project is mixing issue types, but if they are, then they'll be grouped:

```python
>>> sorted(["2", "#11", "#3", "gh-10", "gh-4", "omega", "alpha"], key=issue_key)
['alpha', 'omega', '#3', '#11', 'gh-4', 'gh-10', '2']
```

# Checklist


src/towncrier/test/test_builder.py
```diff
@@ -132,3 +134,59 @@ def test_orphan_all_digits(self):
...
+            fragments=fragments,
+            definitions={},
+            underlines=[],
+            wrap=False,
+            versiondata={},
+        )
+
+    def test_ordering(self):
+        """
+        Issues are ordered by their number, not lexicographically.
```
adiroiban: Not sure if this is correct.

I see gh-2 is ordered before gh-1.

I guess the docstring is missing some details.


```suggestion
        When a single fragment is associated with multiple issue numbers, the issue references are sorted by their number, not lexicographically.
```

src/towncrier/test/test_builder.py
```diff
@@ -132,3 +134,59 @@ def test_orphan_all_digits(self):
             parse_newfragment_basename("+123.feature", ["feature"]),
             ("+123", "feature", 0),
         )
+
+
+class TestIssueOrdering(TestCase):
```
adiroiban: Can we have a docstring for this set of tests ?

I am not sure if this is for ordering for news fragments based on their issue number or for issue numbers inside a single fragment

src/towncrier/newsfragments/608.feature.rst
```diff
@@ -0,0 +1,2 @@
+Issues are now correctly sorted by issue number even if they have non-digit characters.
```
adiroiban: What does "correct" means ?

Maybe something like this

```suggestion
For news fragments that have multiple associted issues, issues are now sorted ascending by issue number even if they have non-digit characters.
```

src/towncrier/_builder.py
```diff
@@ -180,18 +181,45 @@ def split_fragments(
...
+    non_digit_part: str
+    number: int
 
 
-def entry_key(entry: tuple[str, Sequence[str]]) -> tuple[str, list[tuple[int, str]]]:
+def issue_key(issue: str) -> IssueParts:
+    """
+    Used to sort issues in a human-friendly way.
+
+    Issues are grouped their non-integer part, then sorted by their integer part.
```
```suggestion
    Issues are grouped by their non-integer part, then sorted by their integer part.
```

adiroiban (approved):
Thanks! Looks good. Only minor comments.

src/towncrier/newsfragments/608.feature.rst
SmileyChris: I've clarified it now. This doesn't only affect fragments with multiple associated issues -- it also affects overall sorting of fragments.

src/towncrier/test/test_builder.py
SmileyChris: Both :D

src/towncrier/test/test_builder.py
SmileyChris: Again, this primarily is a fix to overall ordering, not just when a fragment has multiple numbers. The primary sort key is the non-numeric portion and specifically with non-numbers first for backwards compatibility, hence why `(random, gh-2)` occurs before `(gh-1)`

src/towncrier/newsfragments/608.feature.rst
```diff
@@ -0,0 +1,8 @@
+Issues are now sorted by issue number even if they have non-digit characters.
+For example::
+
+    - some issue (gh-3, gh-10)
+    - another issue (gh-4)
+    - yet another issue (gh-11)
+
+The sorting algorithm groups the issues first by non-text characters and then by number.
```
SmileyChris, line 8: An interesting aside: without some text after the pre-formatted block, it breaks the rendering of the issue number towncrier adds when building because it's considered part of the block.

src/towncrier/test/test_builder.py
```diff
@@ -132,3 +134,73 @@ def test_orphan_all_digits(self):
...
+        )
+
+    def test_ordering(self):
+        """
+        Issues are ordered first by the non-text component, then by their number.
+
+        For backwards compatibility, issues with no number are grouped first and issues
+        which are only a number are grouped last.
+
+        Orhpan issues are always last, sorted by their fragment text.
```
adiroiban:
```suggestion
        Orhpan news fragments are always last, sorted by their text.
```

src/towncrier/test/test_builder.py
```diff
@@ -132,3 +134,73 @@ def test_orphan_all_digits(self):
...
+            fragments=fragments,
+            definitions={},
+            underlines=[],
+            wrap=False,
+            versiondata={},
+        )
+
+    def test_ordering(self):
+        """
+        Issues are ordered first by the non-text component, then by their number.
```
adiroiban, line 175: I guess here "issues" means "news fragments" 


```suggestion
        News fragments are ordered first by the non-text component, then by their number.
```

src/towncrier/test/test_builder.py
```diff
@@ -132,3 +134,73 @@ def test_orphan_all_digits(self):
             parse_newfragment_basename("+123.feature", ["feature"]),
             ("+123", "feature", 0),
         )
+
+
+class TestIssueOrdering(TestCase):
```
adiroiban:
```suggestion
class TestNewsFragmentsOrdering(TestCase):
```

src/towncrier/newsfragments/608.feature.rst
```diff
@@ -0,0 +1,8 @@
+Issues are now sorted by issue number even if they have non-digit characters.
```
adiroiban:
```suggestion
News fragments are now sorted by ticket number even if they have non-digit characters.
```

Maybe we should update the [Philosophy](https://towncrier.readthedocs.io/en/latest/#philosophy) paragram in the documentation to also introduce the "issue number"

In the past we had Trac `tickets`, now we have GitHub `issues` ... and I think some people have Jira `tickets`


I see that in [towncrier create](https://towncrier.readthedocs.io/en/latest/cli.html#towncrier-create) doc we have 

> To create a news fragment not tied to a specific issue

----------

But in [the tutorial](https://towncrier.readthedocs.io/en/latest/tutorial.html#creating-news-fragments) I see 

> misc: A ticket has been closed, but it is not of interest to users.

> When you create a news fragment, the filename consists of the ticket ID (or some other unique identifier) as well as the ‘type’. 

> For orphan news fragments (those that don’t need to be linked to any ticket ID or other identifier), 


---------------

I think that it's important to have consistent terminology.

For the purpose of being generic, I would prefer to use the term `ticket`

src/towncrier/_builder.py
```diff
@@ -180,18 +181,45 @@ def split_fragments(
...
+    is_digit: bool
+    has_digit: bool
+    non_digit_part: str
+    number: int
 
 
-def entry_key(entry: tuple[str, Sequence[str]]) -> tuple[str, list[tuple[int, str]]]:
+def issue_key(issue: str) -> IssueParts:
+    """
+    Used to sort issues in a human-friendly way.
```
```suggestion
def ticket_key(issue: str) -> IssueParts:
    """
    Used to sort the ticket ID inside a news fragment in a human-friendly way.
```

src/towncrier/_builder.py
```diff
@@ -180,18 +181,45 @@ def split_fragments(
...
+
+    Issues are grouped by their non-integer part, then sorted by their integer part.
+
+    For backwards compatible consistency, issues without no number are sorted first and
+    digit only issues are sorted last.
```
```suggestion
    Ticket IDs are grouped by their non-integer part, then sorted by their integer part.

    For backwards compatible consistency, tickets without no number are sorted first and
    digit only issues are sorted last.
```

src/towncrier/newsfragments/608.feature.rst
adiroiban: We might want to create a GitHub issue for this with the input, actual result vs expected result

adiroiban (approved):
Thanks for the update. 

I left a few inline comments.

I think that part of my confusion is the use of the term "issue" in this PR.

I am not sure what an "issue" is: news fragment vs ticket ID

I think it helps to use consistent terminolgy across the code base and the documentation.

I am not a big fan of "news fragment" ... but this is what we have and I think that we should continue to use it.

Thanks again!

SmileyChris:
> I am not sure what an "issue" is: news fragment vs ticket ID

@adiroiban I actually decided to switch to issue rather than ticket. My logic is it's the only term actually used in code and changing it for consistency would be backwards incompatible.

We use the `issue_format` configuration option and the `issues_by_category` context argument when rendering.

Even though it's not your preference, are you ok with using "issue" across the board instead of ticket?

adiroiban:
Thanks for the update

> Even though it's not your preference, are you ok with using "issue" across the board instead of ticket?

Sure. No problem.

Your suggestion is much bettter.

Happy to see some improvements here

Thanks

adiroiban (approved):
Thanks for the update. All good!

src/towncrier/newsfragments/608.feature.rst
SmileyChris: Done.

PR #610, opened 2024-06-07
Refactor rendering of title via `config.title_format`

### LINKED ISSUE #609: The `title_format` configuration option assumes restructuredtext
The correct solution to my issue in #607 is to use the `title_format` configuration option.

The only problem is that this option [assumes](https://github.com/twisted/towncrier/blob/15d1e25d9dc676cd85fe1488341b272f6bb0e836/src/towncrier/build.py#L249) that you want:
```
<formatted_title>
<config.underscores[0] * formatted_title_length>
<rendered content>
```

But for markdown, we only want:
```
# <formatted_title>
<rendered content>
```

The builder however doesn't have a concept of markdown or restructuredtext.

I think the most compatible solution would be to use the markdown method if the template ends with '.md', otherwise use the current method, correct?

SmileyChris:
# Description
Make `title_format` work for markdown as well as restructured text.
Fixes #609.

# Checklist

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -80,7 +80,7 @@ Top level keys
 
     ``""`` by default.
 
-    Formatted titles are formatted with underlines on the following line (in restructured text) unless the template has an ``.md`` suffix, in which case the title will instead be prefixed with ``# `` (markdown title format).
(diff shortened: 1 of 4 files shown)


adiroiban:
Thanks for the PR

For this PR we can continue with the hardcoded support for RST and MD.

------------

I think that in an ideal case, the builder should not care about makrdown or restructured format.

In the future we can also have AsciiDoc, or other formats.

It would be nice for `title_format` to export all the required variable to be able to build a multi-line jinja template.

PR #612, opened 2024-06-13
Orphans in non showcontent categories

SmileyChris:
# Description
Currently, orphans (newsfragments without an issue number) don't display correctly in a non "showcontent" category, such as the "misc" category.

These fragments don't have an issue number, but if they were important enough to have a fragment, I still think their content should then be shown.

# Checklist

* [x] Ensure documentation is still up-to-date.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -212,9 +212,12 @@ These may include the following optional keys:
 ``showcontent``
     A boolean value indicating whether the fragment contents should be included in the news file.
 
-    Note that orphan fragments (those without an issue number) will still show their content.
-
     ``true`` by default.
+
+    .. note::
+
+        Orphan fragments (those without an issue number) always have their content included.
+        If a fragment was created, it means that information is important for end users.
 
 For example, if you want your custom fragment types to be ``["feat", "fix", "chore",]`` and you want all of them to use the default configuration except ``"chore"`` you can do it as follows:
 
@@ -255,9 +258,12 @@ Each table within this array has the following mandatory keys:
 ``showcontent``
     A boolean value indicating whether the fragment contents should be included in the news file.
 
-    Note that orphan fragments (those without an issue number) will still show their content.
-
     ``true`` by default.
+
+    .. note::
+
+        Orphan fragments (those without an issue number) always have their content included.
+        If a fragment was created, it means that information is important for end users.
 
 For example:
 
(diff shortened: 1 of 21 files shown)


docs/configuration.rst
```diff
@@ -212,6 +212,8 @@ These may include the following optional keys:
 ``showcontent``
     A boolean value indicating whether the fragment contents should be included in the news file.
 
+    Note that orphan fragments (those without an issue number) will still show their content.
```
adiroiban: Maybe have it as an RST admonition  ... but below the `true by default` line

```suggestion
.. note::

    Orphan fragments (those without an issue number) always have their content included.
    If a fragment was created, it means that information is important for end users.
```

docs/configuration.rst
```diff
@@ -253,6 +255,8 @@ Each table within this array has the following mandatory keys:
...
```
adiroiban: same as the other comment.
Have it as a dedicated `RST note`

src/towncrier/templates/default.md
```diff
@@ -15,7 +15,6 @@
 {% for category, val in definitions.items() if category in sections[section] %}
 ### {{ definitions[category]['name'] }}
 
-{% if definitions[category]['showcontent'] %}
```
adiroiban, line 18: I think that we should add an `612.removal` fragment, informat that when `showcontent = False` the template will no longer receive the fragment content.

This is a note, just in case someone was missusing this feature with a custom template.

adiroiban (approved):
Thanks. Looks good.

My only major comment is to add a `.removal` release note informing that the Jinja2 template will no longer receive the fragment content, for `showcontent = False` categories.

src/towncrier/templates/default.md
SmileyChris: I think there's some confusion here. The existing behaviour is already that fragment content is not sent to the template (https://github.com/twisted/towncrier/pull/612/files#diff-36db088409efe07783d49792a18cec6ae3778ace338bf1cd08daa8a3fce29e93L166). This change just means that orphan fragment contents **are** sent.

adiroiban: True. My bad. I misinterpreted the changes in the code.

adiroiban (approved):
All good. Thanks.

Sorry for the previous review that was mediocre.

PR #615, opened 2024-06-13
Require jinja2 v3.0

### LINKED ISSUE #569: Constraint PIP Dependencies Version Range for Towncrier Compatibility
## Description

The pip version on my machine is '2.11.3'. When I use `pip install towncrier` to install the program, there are no errors during the process. This is because towncrier depends on any version of pip, but this is problematic.

When I use the `build` command, the following exception is thrown:

```
Warning: In a future version of Poetry, PyPI will be disabled automatically if at least one custom source is configured with another priority than 'explicit'. In order to avoid a breaking change and make your pyproject.toml forward compatible, add PyPI explicitly via 'poetry source add pypi'. By the way, this has the advantage that you can set the priority of PyPI as with any other source.
Loading template...
Finding news fragments...
Rendering news fragments...
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  ...
  ...
jinja2.exceptions.TemplateSyntaxError: expected token 'end of statement block', got '+'
```

## Reproduction
```
pip install jinja2=="2.11.3"
pip install markupsafe=="2.0.1"
towncrier build --draft --yes --version=1.0.0
```

## Solution

Upgrade the version of jinja2 to `3.*`
```
pip install jinja2 --upgrade
```

Should the version range of pip be constrained?

https://github.com/twisted/towncrier/blob/57f203651a080c2a91a3cf9b0e47eb6fca42128b/pyproject.toml#L35

SmileyChris:
Fixes #569

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     "importlib-resources>=5; python_version<'3.10'",
     "importlib-metadata>=4.6; python_version<'3.10'",
     "incremental",
-    "jinja2",
+    "jinja2>=3",
     "tomli; python_version<'3.11'",
 ]
 
--- a/src/towncrier/newsfragments/569.misc.rst
+++ b/src/towncrier/newsfragments/569.misc.rst
@@ -1,0 +1,1 @@
+Constrain jinja2 to v3+ -- our templates require this and it was released over 3 years ago (May 2021)


adiroiban (approved):
Thanks.

PR #618, opened 2024-06-20
Ignore types on check

### LINKED ISSUE #617: allow specifying fragment types to consider as "valid" for `towncrier check`
Hello, it would be helpful to us to be able to specify within `towncrier check` which fragment types it should look for or not look for.

**Background on why this would be useful:** We have automatic processes that create a newsfragment of a special type (`dependency_bump`) whenever a version of one of dependency libraries is updated. Even though these are real changes that we'd like to be noted in the changelog, we'd like `towncrier check` to assert that there are real human-readable newsfragments that explain the actual change, if any.

I propose an additional, optional key to configured sections (`[[tool.towncrier.type]]`): `ignore_on_check`. if this key is configured to `True`, `towncrier check` will ignore all newsfragments of this type

bentheiii:
# Description

closes  #617 
Adds a new optional key to newsfragment categories called `check` (boolean, defaults to true). If set to false, towncrier check will not consider newsfragments of this category as enough to pass

# Checklist


src/towncrier/test/test_check.py
```diff
@@ -166,6 +166,71 @@ def test_fragment_missing(self):
                 result.output.endswith("No new newsfragments found on this branch.\n")
             )
 
+    def test_fragment_exists_but_not_in_check(self):
```
adiroiban, line 169: Can you please add a paragraph describing the scope and the end user functionality tested here ?

Same comment as for `def test_fragment_exists_but_in_check(self):`

I know that as part of this PR, the tests make sense.

but if someone reads these tests one year into the future, and without being familiar with the check feature, it might not be that obvious

No need to be verbose. Just the important part

See https://jml.io/test-docstrings/


--------


We don't have tests for the testing code.

So the test docstring is the only input that can be used to manualy validate / check that the test code is valid.

docs/configuration.rst
```diff
@@ -281,3 +295,9 @@ For example:
    directory = "chore"
    name = "Other Tasks"
    showcontent = false
+
+   [[tool.towncrier.type]]
+   directory = "auto"
+   name = "Automatic Changes"
```
adiroiban: I prefer not to use `auto` here.

Some people might thing that `auto` has special meaning for `towncrier` and that `towncrier` does something automatically.

```suggestion
   [[tool.towncrier.type]]
   directory = "deps"
   name = "Dependency Changes"
```

src/towncrier/_builder.py
```diff
@@ -102,6 +102,7 @@ def __call__(self, section_directory: str = "") -> str:
 def find_fragments(
     base_directory: str,
     config: Config,
+    only_check_categories: bool = False,
```
adiroiban: This flag is hard do understand.

would it be possible to refactor the `find_fragments` to also return the category for each file ?

In this way, we don't need to implement any filtering here.

We can implement the filtering in `check.py`

-------

No need to implement the following sugestion in this PR.

With such an API, in check.py we can have 2 error message:

* no fragmtents at all
* only ignored fragments

src/towncrier/test/test_check.py
```diff
@@ -166,6 +166,71 @@ def test_fragment_missing(self):
...
+            create_project(
+                "pyproject.toml",
+                main_branch="master",
+                extra_config="[[tool.towncrier.type]]\n"
+                'directory = "feature"\n'
+                'name = "Features"\n'
+                "showcontent = true\n"
+                "[[tool.towncrier.type]]\n"
+                'directory = "auto"\n'
+                'name = "Automatic"\n'
```
adiroiban: very minor comment... just to make it clear that this is the scope/targe of this test.

```suggestion
                'directory = "sut"\n'
                'name = "System Under Test"\n'
```

adiroiban (changes requested):
Many thanks for the PR.

It looks good.

I left a few minor comments

Major comments:

* The `only_check_categories` is confusing... especially without any extra documentation... but I hope we can have it removed.
* The tests needs a docstring/description

Thanks again

docs/configuration.rst
```diff
@@ -221,6 +221,11 @@ These may include the following optional keys:
         Orphan fragments (those without an issue number) always have their content included.
         If a fragment was created, it means that information is important for end users.
 
+``check``
```
adiroiban, line 224: Sorry for not raising this in the first place. My bad.

Should this be more explicit / specific ?

Mabye in the future we would want to add some other things related to `check` behaviour. ... I am not sure what we might add... so this is just a minor comment.

```suggestion
``ignore_check``
```

docs/configuration.rst
```diff
@@ -221,6 +221,11 @@ These may include the following optional keys:
...
+    A boolean value indicating whether the fragment should be considered by the ``towncrier check`` command.
```
adiroiban, line 225: This is not very clear... but I don't know how to better explain ths... writing good documentation is hard :)

Maybe

```suggestion
    A boolean value indicating whether the fragment should be considered by the ``towncrier check`` command.
    When set to `false`, ``towncrier check`` will fail, unless the branch contains other newsfragment types.
```

adiroiban (approved):
Looks good. Thanks.

Only minor comments.

I will let this PR unmerged for a few days to see if @SmileyChris has any extra feedback about this PR.

No news is good news, is we can check on Monday and see if this can be merged.

Thanks again

docs/configuration.rst
bentheiii: I think that if this were ever needed, it would be easier to make the `check` a `dict|bool` rather than a separete knob from the `ignore_check`
```toml
[[tool.towncrier.type]]
...
check=false
[[tool.towncrier.type]]
...
check=true
[[tool.towncrier.type]]
...
check={"some_new_flag": "some_value"}
```

adiroiban:
```
check={"some_new_flag": "some_value"}
```

this looks bad to me... if we use this, why bother with .INI or .TOML and not just go with JSON  ?

If you dislike `ignore_check` then we can keep the current option.

bentheiii:
Hi @adiroiban, can we merge this?

PR #622, opened 2024-07-10
Check if any news fragments have invalid filenames

### LINKED ISSUE #426: Multiple fragment types for same fragment (e.g. `.bug` and `.bugfix`)
We have multiple developers and sometimes end up with changelog files like  `123.bugfix.md` and `123.bug.md`.

**I would love to have multiple fragment types grouped together.**

Currently we use some CI task to verify no unknown fragments get committed, but I would love to see this as a feature.

Configuring multiple fragment types using the same `name` attribute just doubles the entry in the parsed changelog file.

### LINKED ISSUE #619: Check for invalid filenames in `changes.d/`
Although `towncrier create` enforces valid fragment file names, creating a fragment file manually can sometimes lead to mistakes in the file name. It would be useful to include in our CI a check that all files in `changes.d/` have valid fragment file names (excluding `changelog-template.jinja` of course).

Note we don't use `towncrier check` as we don't create fragments for small or non-user-facing changes.

MetRonnie:
# Description

Closes #619
Fixes #426

``towncrier check`` will now fail if any news fragments have invalid filenames.

Added a new configuration option called ``build_ignore_filenames`` that allows you to specify a list of filenames that should be ignored. If this is set, ``towncrier build`` will also fail if any filenames are invalid, except for those in the list.

Added a `--strict` option to `towncrier build` to enable this behaviour for when `build_ignore_filenames` has not been configured.

# Checklist

* [x] If you add new **configuration options** (or change the meaning of existing ones), make sure `docs/configuration.rst` reflects those changes. (Not applicable)

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -64,11 +64,6 @@ also remove the fragments directory if now empty.
 
    Don't delete news fragments after the build and don't ask for confirmation whether to delete or keep the fragments.
 
-.. option:: --strict
-
-   Fail if there are any news fragments that have invalid filenames.
-   This is automatically turned on if ``build_ignore_filenames`` has been set in the configuration.
-
 
 ``towncrier create``
 --------------------
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -64,6 +64,11 @@ also remove the fragments directory if now empty.
 
    Don't delete news fragments after the build and don't ask for confirmation whether to delete or keep the fragments.
 
+.. option:: --strict
+
+   Fail if there are any news fragments that have invalid filenames.
+   This is automatically turned on if ``build_ignore_filenames`` has been set in the configuration.
+
 
 ``towncrier create``
 --------------------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -130,6 +130,13 @@ Top level keys
 
     ``true`` by default.
 
+``build_ignore_filenames``
+    A list of filenames in the news fragments directory to ignore when building the news file.
+
+    If this is configured, it turns on the ``--strict`` build mode which will fail if there are any news fragment files not in this list that have invalid filenames. Note that if a template is used, it should also be included here.
+
+    ``None`` by default.
+
 Extra top level keys for Python projects
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -130,10 +130,12 @@ Top level keys
 
     ``true`` by default.
 
-``build_ignore_filenames``
-    A list of filenames in the news fragments directory to ignore when building the news file.
+``ignore``
+    A list of filenames in the news fragments directory to ignore.
 
-    If this is configured, it turns on the ``--strict`` build mode which will fail if there are any news fragment files not in this list that have invalid filenames. Note that if a template is used, it should also be included here.
+    ``towncrier check`` will fail if there are any news fragment files that have invalid filenames, except for those in the list. ``towncrier build`` will likewise fail, but only if this list has been configured (set to an empty list if there are no files to ignore).
+
+    Note that if a custom template is stored in the news fragment directory, you should add it to this list.
 
     ``None`` by default.
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -131,11 +131,11 @@ Top level keys
     ``true`` by default.
 
 ``ignore``
-    A list of filenames in the news fragments directory to ignore.
+    A case-insensitive list of filenames in the news fragments directory to ignore.
 
     ``towncrier check`` will fail if there are any news fragment files that have invalid filenames, except for those in the list. ``towncrier build`` will likewise fail, but only if this list has been configured (set to an empty list if there are no files to ignore).
 
-    Note that if a custom template is stored in the news fragment directory, you should add it to this list.
+    Some filenames such as ``.gitignore`` and ``README`` are automatically ignored. However, if a custom template is stored in the news fragment directory, you should add it to this list.
 
     ``None`` by default.
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -47,11 +47,6 @@ Create this folder::
 
 The ``.gitignore`` will remain and keep Git from not tracking the directory.
 
-If needed, you can also specify a list of filenames for ``towncrier`` to ignore in the news fragments directory::
-
-   [tool.towncrier]
-   build_ignore_filenames = ["README.rst"]
-
 
 Detecting Version
 -----------------
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -47,6 +47,11 @@ Create this folder::
 
 The ``.gitignore`` will remain and keep Git from not tracking the directory.
 
+If needed, you can also specify a list of filenames for ``towncrier`` to ignore in the news fragments directory::
+
+   [tool.towncrier]
+   build_ignore_filenames = ["README.rst"]
+
 
 Detecting Version
 -----------------
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -45,8 +45,6 @@ Create this folder::
    # This makes sure that Git will never delete the empty folder
    $ echo '!.gitignore' > src/myproject/newsfragments/.gitignore
 
-The ``.gitignore`` will remain and keep Git from not tracking the directory.
-
 
 Detecting Version
 -----------------
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -114,9 +114,9 @@ def find_fragments(
 
     If strict, raise ClickException if any fragments have an invalid name.
     """
-    ignored_files = {".gitignore"}
+    ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
     if config.ignore:
-        ignored_files.update(config.ignore)
+        ignored_files.update(filename.lower() for filename in config.ignore)
 
     get_section_path = FragmentsPath(base_directory, config)
 
@@ -137,7 +137,7 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
-            if basename in ignored_files:
+            if basename.lower() in ignored_files:
                 continue
 
             issue, category, counter = parse_newfragment_basename(
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -107,21 +107,16 @@ class FragmentsPath:
 def find_fragments(
     base_directory: str,
     config: Config,
-    strict: bool | None,
+    strict: bool,
 ) -> tuple[Mapping[str, Mapping[tuple[str, str, int], str]], list[tuple[str, str]]]:
     """
     Sections are a dictonary of section names to paths.
 
-    In strict mode, raise ClickException if any fragments have an invalid name.
+    If strict, raise ClickException if any fragments have an invalid name.
     """
-    if strict is None:
-        # If strict mode is not set, turn it on only if build_ignore_filenames is set
-        # (this maintains backward compatibility).
-        strict = config.build_ignore_filenames is not None
-
     ignored_files = {".gitignore"}
-    if config.build_ignore_filenames:
-        ignored_files.update(config.build_ignore_filenames)
+    if config.ignore:
+        ignored_files.update(config.ignore)
 
     get_section_path = FragmentsPath(base_directory, config)
 
@@ -142,7 +137,6 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
-            # Skip files that are deliberately ignored
             if basename in ignored_files:
                 continue
 
@@ -154,7 +148,7 @@ def find_fragments(
                     raise ClickException(
                         f"Invalid news fragment name: {basename}\n"
                         "If this filename is deliberate, add it to "
-                        "'build_ignore_filenames' in your configuration."
+                        "'ignore' in your configuration."
                     )
                 continue
             assert issue is not None
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -12,6 +12,7 @@ from collections import defaultdict
 from pathlib import Path
 from typing import Any, DefaultDict, Iterable, Iterator, Mapping, NamedTuple, Sequence
 
+from click import ClickException
 from jinja2 import Template
 
 from towncrier._settings.load import Config
@@ -106,10 +107,22 @@ class FragmentsPath:
 def find_fragments(
     base_directory: str,
     config: Config,
+    strict: bool | None,
 ) -> tuple[Mapping[str, Mapping[tuple[str, str, int], str]], list[tuple[str, str]]]:
     """
     Sections are a dictonary of section names to paths.
+
+    In strict mode, raise ClickException if any fragments have an invalid name.
     """
+    if strict is None:
+        # If strict mode is not set, turn it on only if build_ignore_filenames is set
+        # (this maintains backward compatibility).
+        strict = config.build_ignore_filenames is not None
+
+    ignored_files = {".gitignore"}
+    if config.build_ignore_filenames:
+        ignored_files.update(config.build_ignore_filenames)
+
     get_section_path = FragmentsPath(base_directory, config)
 
     content = {}
@@ -129,10 +142,20 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
+            # Skip files that are deliberately ignored
+            if basename in ignored_files:
+                continue
+
             issue, category, counter = parse_newfragment_basename(
                 basename, config.types
             )
             if category is None:
+                if strict and issue is None:
+                    raise ClickException(
+                        f"Invalid news fragment name: {basename}\n"
+                        "If this filename is deliberate, add it to "
+                        "'build_ignore_filenames' in your configuration."
+                    )
                 continue
             assert issue is not None
             assert counter is not None
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -106,6 +106,17 @@ def _validate_answer(ctx: Context, param: Option, value: bool) -> bool:
     help="Do not ask for confirmations. But keep news fragments.",
     callback=_validate_answer,
 )
+@click.option(
+    "--strict",
+    "strict",
+    default=None,
+    flag_value=True,
+    help=(
+        "Fail if there are any news fragments that have invalid filenames (this is "
+        "automatically turned on if 'build_ignore_filenames' has been set in the "
+        "configuration)."
+    ),
+)
 def _main(
     draft: bool,
     directory: str | None,
@@ -115,6 +126,7 @@ def _main(
     project_date: str | None,
     answer_yes: bool,
     answer_keep: bool,
+    strict: bool | None,
 ) -> None:
     """
     Build a combined news file from news fragment.
@@ -129,6 +141,7 @@ def _main(
             project_date,
             answer_yes,
             answer_keep,
+            strict,
         )
     except ConfigError as e:
         print(e, file=sys.stderr)
@@ -144,6 +157,7 @@ def __main(
     project_date: str | None,
     answer_yes: bool,
     answer_keep: bool,
+    strict: bool | None,
 ) -> None:
     """
     The main entry point.
@@ -178,7 +192,7 @@ def __main(
 
     click.echo("Finding news fragments...", err=to_err)
 
-    fragment_contents, fragment_files = find_fragments(base_directory, config)
+    fragment_contents, fragment_files = find_fragments(base_directory, config, strict)
     fragment_filenames = [filename for (filename, _category) in fragment_files]
 
     click.echo("Rendering news fragments...", err=to_err)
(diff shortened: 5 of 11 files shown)


src/towncrier/_builder.py
```diff
@@ -17,15 +17,18 @@
...
 def parse_newfragment_basename(
-    basename: str, frag_type_names: Iterable[str]
+    basename: str, frag_type_names: Iterable[str], strict: bool = False
 ) -> tuple[str, str, int] | tuple[None, None, None]:
     invalid = (None, None, None)
     parts = basename.split(".")
 
     if len(parts) == 1:
+        if strict:
+            raise ValueError(f"Invalid news fragment name: {basename}")
```
MetRonnie: I have raised a `ValueError` here but not sure if there is a better exception type to raise. Currently this results in traceback in the command line

adiroiban: There is `click.ClickException` for the CLI, as we are using `click`

src/towncrier/_builder.py
```diff
@@ -17,15 +17,18 @@
 from towncrier._settings.load import Config
 
 
-# Returns issue, category and counter or (None, None, None) if the basename
-# could not be parsed or doesn't contain a valid category.
+# Returns issue, category and counter. But if the basename could not be parsed
+# or doesn't contain a valid category, returns (None, None, None) or raises
+# a ValueError if in strict mode.
 def parse_newfragment_basename(
```
adiroiban: I think that this function should be kept as it is.

We can have a separate helpet method that raises an exception if the invalid placehoder is returned.

src/towncrier/_builder.py
```diff
@@ -129,8 +140,10 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
+            if basename == template_filename:
```
adiroiban: this needs a comment.
Why are we skipping it ?

src/towncrier/_builder.py
```diff
@@ -129,8 +140,10 @@ def find_fragments(
...
+                continue
             issue, category, counter = parse_newfragment_basename(
-                basename, config.types
+                basename, config.types, strict
             )
```
adiroiban: Maybe we can have something like this 
```suggestion
                basename, config.types
            )
        if strict and not (issue is None and category is None):
            raise ClickException(f'Invalid news fragment found: {basename}.')
```

src/towncrier/build.py
```diff
@@ -106,6 +106,16 @@ def _validate_answer(ctx: Context, param: Option, value: bool) -> bool:
     help="Do not ask for confirmations. But keep news fragments.",
     callback=_validate_answer,
 )
+@click.option(
```
adiroiban: I prefer to have this CLI option inside `towncrier check`

I think that for `towncrier build` this should be the default.

That is, `towncrier build` should always fail when unknown fragments are found.

--------

and maybe we don't need a `towncrier check` fragment.

Just make this the defaul and add an option to explicly ignore selected files inside the `newsfragments` directory.

adiroiban (changes requested):
Thanks for working on this.

This needs documentation and tests to be merged.

--------

While reviewing this, my question is.

Why do we ever want to allow `towncrier build` or `towncrier check` to succeed if there are in invalid fragments.

I understand that you might want to have other files inside the `newsfragment` folder (for example the template and a README) but other than that, I think `towncrier` should fail

Cheers

src/towncrier/build.py
MetRonnie: Sounds good

adiroiban: Note that my suggestion about the default behaviour for `towncrier build` might not be backward compatible .

I was just talking about the ideal case.

------

So we might need to add a configuration like this `build_ignore_filenames = ["README.rst", "template.j2"]`

and then fail by default, only if `build_ignore_filenames` is not empty.

and `build_ignore_filenames` should be empty by default

This should give us backward compatibility

-------

and then we can have `towncrier build --strict` that will fail if there are extra files, even when `build_ignore_filenames` is empty.

MetRonnie: So `towncrier build` should have a `--strict` flag, but should it also have an opposite flag so that even if `build_ignore_filenames` is set, a user can ignore any invalid filenames?

Also I'm a bit confused about what you would like for `towncrier check`, is it the same flags as for `towncrier build`?

adiroiban: I think that `towncrier check` don't need any flag. It should just fail by default for invalid fragments.

The backward compatibility should be for `towncrier build`.

I think that for `towncrier check` people will expect that newer version of `towncrier` will introduce more checks ... so you might end up with a check that fails.

MetRonnie: Hmm, I'm not sure this works, because if you haven't set `config.build_ignore_filenames` then `towncrier check` will fail on the template file.

We could do `strict=(config.build_ignore_filenames is not None)`, but then might as well add the `--strict` option to `towncrier check`...

adiroiban: > Hmm, I'm not sure this works, because if you haven't set config.build_ignore_filenames then towncrier check will fail on the template file.

The failure should be ok.
If `towncrier check` fails, you will need to udpate `config.build_ignore_filenames`
The error message might include a hint that the file should be added to `towncrier` ignore liss

-----

The template file might be stored outside of the `newsfragments` directory,
in which case, all is ok.

But most probably we will alwasy have a ".keep" or ".gitignore" file that should be ignored by default

For example this is what we have for `towncrier` and similar for `twisted`

https://github.com/twisted/towncrier/blob/trunk/src/towncrier/newsfragments/.gitignore

MetRonnie:
I think all feedback has been addressed now. Let me know if there is anything else.

Maybe a warning if `towncrier build` encounters invalid filenames but `build_ignore_filenames` has not been configured? Do you envisage deprecating the `--strict` option and making the fail behaviour the default in a future version of Towncrier?

docs/cli.rst
```diff
@@ -64,6 +64,11 @@ also remove the fragments directory if now empty.
 
    Don't delete news fragments after the build and don't ask for confirmation whether to delete or keep the fragments.
 
+.. option:: --strict
```
adiroiban: Do we need this argument ?

I think that having `build_ignore_filenames` is enough.

docs/configuration.rst
```diff
@@ -130,6 +130,13 @@ Top level keys
 
     ``true`` by default.
 
+``build_ignore_filenames``
+    A list of filenames in the news fragments directory to ignore when building the news file.
+
+    If this is configured, it turns on the ``--strict`` build mode which will fail if there are any news fragment files not in this list that have invalid filenames. Note that if a template is used, it should also be included here.
```
adiroiban: I think that we can remove the `--strict` CLI argument

docs/tutorial.rst
```diff
@@ -47,6 +47,11 @@ Create this folder::
 
 The ``.gitignore`` will remain and keep Git from not tracking the directory.
```
adiroiban: Just a drive by comment. What does this mean ? :)

Maybe we can improve this paragraph as part of this PR.

docs/tutorial.rst
```diff
@@ -47,6 +47,11 @@ Create this folder::
...
 
+If needed, you can also specify a list of filenames for ``towncrier`` to ignore in the news fragments directory::
```
adiroiban: I think that `.gitignore` should alwasy be ignored ... so there is nothing more to do for this tutorial.

This is why I think it needs to be moved outside of the tutorial.

-------

I think that this needs a better introduction, besides moving it outside of the tutorial.


If needed ... by whom? For doing doing what?

src/towncrier/_builder.py
```diff
@@ -106,10 +107,22 @@ def __call__(self, section_directory: str = "") -> str:
 def find_fragments(
     base_directory: str,
     config: Config,
+    strict: bool | None,
```
adiroiban: I hope that we can remove this argument and just use the configuration value to control the behaviour of this method.

src/towncrier/_builder.py
```diff
@@ -129,10 +142,20 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
+            # Skip files that are deliberately ignored
+            if basename in ignored_files:
```
adiroiban: I think that is best to have the comment inside the `if` block.

But with the newer variable names, I don't think that the comment is needed anymore.

my previous request for a comment was for a different code.



```suggestion
            if basename in ignored_files:
                # The file is configured to be ignored.
```

src/towncrier/build.py
adiroiban: Maybe we can remove this CLI argument and just use the configuration to control the behaviour of `build`

src/towncrier/check.py
```diff
@@ -101,12 +101,14 @@ def __main(
         click.echo(f"{n}. {change}")
     click.echo("----")
 
+    # This will fail if any fragment files have an invalid name:
```
adiroiban, line 104: I don't think that this comment is needed.

The docstring for `find_fragments` should be enough of a documentation.

src/towncrier/_builder.py
```diff
@@ -106,10 +107,22 @@ def __call__(self, section_directory: str = "") -> str:
 def find_fragments(
     base_directory: str,
     config: Config,
+    strict: bool | None,
 ) -> tuple[Mapping[str, Mapping[tuple[str, str, int], str]], list[tuple[str, str]]]:
     """
     Sections are a dictonary of section names to paths.
+
+    In strict mode, raise ClickException if any fragments have an invalid name.
```
adiroiban: I know that this questions might be easy to answer in the context of this PR, but I think that the documentation should be improved... so that it will make sense when read outside of this PR.


Is this exception raised only in `strict mode` ?
What is `strict mode` ?

--------

What about raising the exception when `build_ignore_filenames` is set ?

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1530,60 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+            f.write("Peanut butter")
+        # Automatically ignored:
+        with open("foo/newsfragments/.gitignore", "w") as f:
+            f.write("!.gitignore")
+
+        result = runner.invoke(_main, opts)
+        # Should succeed
+        self.assertEqual(0, result.exit_code, result.output)
+
+        # Invalid filename:
```
adiroiban: I think that this should be a different test case with a dedicated test docstring.

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1530,60 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        """
+    )
+    def test_invalid_fragment_names(self, runner):
+        """
+        When build_ignore_filenames is set, files with those names are ignored.
+        """
+        opts = ["--draft", "--date", "01-01-2001", "--version", "1.0.0"]
+        # Valid filename:
+        with open("foo/newsfragments/123.feature", "w") as f:
+            f.write("Adds levitation")
```
adiroiban: Maybe just document the purpose of the file inside it's content.
Tehn there is no need for a separate comment.
```suggestion
            f.write("Adds levitation in valid filename.")
```

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1530,60 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        # Files that should be ignored:
+        with open("foo/newsfragments/template.jinja", "w") as f:
+            f.write("Jinja template")
```
adiroiban:
```suggestion
        with open("foo/newsfragments/template.jinja", "w") as f:
            f.write("Jinja template file is ignored.")
```

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1530,60 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        with open("foo/newsfragments/CAPYBARAS.md", "w") as f:
+            f.write("Peanut butter")
+        # Automatically ignored:
+        with open("foo/newsfragments/.gitignore", "w") as f:
+            f.write("!.gitignore")
+
+        result = runner.invoke(_main, opts)
+        # Should succeed
```
adiroiban: Does this comment adds any value ?

https://jml.io/test-docstrings/

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1530,60 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        self.assertEqual(0, result.exit_code, result.output)
+
+        # Invalid filename:
+        with open("foo/newsfragments/feature.124", "w") as f:
+            f.write("Extends levitation")
+
+        result = runner.invoke(_main, opts)
+        # Should now fail
```
adiroiban: I don't think that this comment helps.

It's better to have a separate test, with a dedicated docstring and in that docstring add all in information needed to understand the scope and purpose of the test.

src/towncrier/test/test_build.py
```diff
@@ -1530,3 +1530,60 @@ def test_orphans_in_non_showcontent_markdown(self, runner):
...
+        """
+        When using --strict, any invalid filenames will cause an error even if
+        build_ignore_filenames is NOT set.
+        """
+        opts = ["--draft", "--date", "01-01-2001", "--version", "1.0.0"]
+        # Invalid filename:
+        with open("foo/newsfragments/feature.124", "w") as f:
+            f.write("Extends levitation")
+
+        result = runner.invoke(_main, opts)
```
adiroiban: I think that this needs a separate test.

The docstring talks about `using --string` but I can see that `--strict` is not used here.

src/towncrier/test/test_check.py
```diff
@@ -470,3 +470,23 @@ def test_in_different_dir_with_nondefault_newsfragments_directory(self, runner):
         self.assertTrue(
             result.output.endswith("No new newsfragments found on this branch.\n")
         )
+
+    @with_isolated_runner
+    def test_invalid_fragment_name(self, runner):
```
adiroiban: I think that this needs a docstring.

docs/configuration.rst
```diff
@@ -130,6 +130,13 @@ Top level keys
 
     ``true`` by default.
 
+``build_ignore_filenames``
```
adiroiban: Just a minor suggestion.

Since this is used by both `build` and `check` maybe keep the config name short.

```suggestion
``ignore_filenames``
```

adiroiban (changes requested):
Many thanks for the changes.

They look good.

As a general rule, I think that is best to start any new functionality by writing the documentation.

In this way, we can agree on how the feature should behave, before writing more code.

------

I think that we can have this implemented without the `--strict` option.

In terms for testing I am expecting 3 main test scenarios:

* `build_ignore_filenames` not set
* `build_ignore_filenames` is set to empty list
* `build_ignore_filenames` is set with some filenames

Thanks again for working on this!

docs/configuration.rst
MetRonnie: I'm changing it to just `ignore`, but if `ignore_filenames` or something else is preferable, let me know

docs/tutorial.rst
MetRonnie: Nothing to do with me, but this line seems a bit redundant looking at the snippet above

docs/configuration.rst
adiroiban: We can go witn `ignore` . Thanks. It was just a minor feedback.

adiroiban:
I forgot to leave this feedback

> Maybe a warning if towncrier build encounters invalid filenames but build_ignore_filenames has not been configured?

That would be nice. We can also have this in a separate PR.
This should not be a blocker.

-------------

> Do you envisage deprecating the --strict option and making the fail behaviour the default in a future version of Towncrier?

I hope that we don't have to add the `--strict` option... so that we don't have to deprecate it.

It would be nice to have a warinig raised when `build_ignore_filenames` is not configured and make a release that raised the warning.

In a followup release, we can then change the default value of  `build_ignore_filenames` so thta it will alwasy do a strict check by default... while still allowing users to configure it with ` `build_ignore_filenames = None`

adiroiban:
I have updated the PR description to note that it will also fix #426  ... the #619 might be a duplicate

src/towncrier/check.py
MetRonnie: I added this comment to make sure anyone thinks twice before possibly moving this back down below to just before the result is used (because the next part could lead this this check being skipped)

src/towncrier/_builder.py
MetRonnie: I don't think so, because we have different behaviour for ``towncrier check`` vs ``towncrier build``. Unless you want to align their behaviours?

src/towncrier/_builder.py
```diff
@@ -106,10 +107,17 @@ def __call__(self, section_directory: str = "") -> str:
...
     base_directory: str,
     config: Config,
+    strict: bool,
 ) -> tuple[Mapping[str, Mapping[tuple[str, str, int], str]], list[tuple[str, str]]]:
     """
     Sections are a dictonary of section names to paths.
+
+    If strict, raise ClickException if any fragments have an invalid name.
     """
+    ignored_files = {".gitignore"}
```
adiroiban: maybe also add `.keep` .. and maybe `README, README.rst, README.md` ... and mabye the check can be case insensitive.

I expect that in the future we might receive suggestion to add more default values... but that should be fine.

src/towncrier/_builder.py
```diff
@@ -129,10 +137,19 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
+            if basename in ignored_files:
```
adiroiban: Does it make sense to do a case insenstive match  ?

```suggestion
            if basename.lower() in ignored_files:
```

and making sure values in `ignored_files` are kept lower case.

src/towncrier/check.py
adiroiban: We can keep it. No problem.

The critical part is to make sure that we have automated tests that will fail is anyone would move this code :)

docs/tutorial.rst
adiroiban: Maybe we can just remove it.

I as reviewer, I try to read the documentation in the context of the whole page... and this paragraph doens't make sense.

So maybe we should just remove it as part of this PR. Thanks

src/towncrier/newsfragments/622.feature.rst
```diff
@@ -0,0 +1,3 @@
+``towncrier check`` will now fail if any news fragments have invalid filenames.
```
adiroiban, line 1: Not sure how this will ended up in the rendered version.

We should have 2 fragments here.
One for `towncrier check` 

and another one for the `ignore` configuration option.

-----

But maybe the `towncrier check` part can be left out.

If `build` fails, it should be expect for `check` to also fail.

adiroiban (commented):
Many thanks for the changes.

They look good.

I have to still review the automated tests... but the other part looks good.

src/towncrier/check.py
MetRonnie: I think this will fail because of the control case fragment, but I haven't tried moving the line back down and re-running

https://github.com/twisted/towncrier/blob/3c93f134eda738ca7e27eef541b08448bec80d88/src/towncrier/test/test_check.py#L493-L503

src/towncrier/newsfragments/622.feature.rst
MetRonnie: I feel like the addition of the config option and the new fail behaviour are intertwined, so I think it makes sense to have them in the same fragment. At least I think the change in behaviour of ``towncrier check`` should be clearly signposted.

I could rewrite this to be clearer, how about:
> ``towncrier check`` will now fail if any news fragments have invalid filenames. You can specify filenames to ignore in the news fragments directory with the new ``ignore`` configuration option. ``towncrier build`` will not fail on invalid filenames unless ``ignore`` is set (can be an empty list).

src/towncrier/check.py
adiroiban: We need to check this, as the test are the critical part of a PR.

MetRonnie: Done (test updated to ensure)

adiroiban: Thanks. I didn't had time to check the test coverage... but if say that you have checked it, we can have this merged :)

no problem

src/towncrier/newsfragments/622.feature.rst
adiroiban: As you think it's best

we can leave it as it is. no problem.

adiroiban (approved):
Looks good. Many thanks for the changes.

src/towncrier/_builder.py
```diff
@@ -129,10 +137,19 @@ def find_fragments(
         file_content = {}
 
         for basename in files:
+            if basename.lower() in ignored_files:
```
adiroiban, line 140: It looks like with this change, the coverage if failing for line 30 :(

adiroiban (approved):

src/towncrier/_builder.py
MetRonnie: Ah right, I think my latest commit should sort that

adiroiban:
Thanks for the update. I have enabled auto-merge.

I hope this will be merged soon.

I plan to do a new release soon... I hope #627  will also be merge soon and then we can have a release.

PR #625, opened 2024-07-21
Remove the explicit setting of a pygments theme

pradyunsg:
This allows the underlying theme being used to provide the pygments theme instead which looks better (subjectively).

# Description

# Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,9 +74,6 @@ release = towncrier_version.public()
 # This patterns also effect to html_static_path and html_extra_path
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
-# The name of the Pygments (syntax highlighting) style to use.
-pygments_style = "sphinx"
-
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 


pradyunsg:
> * Create a file in `src/towncrier/newsfragments/`.

I don't think this is a news worthy change, but let me know if you want one.

docs/conf.py
```diff
@@ -75,7 +75,7 @@
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = "sphinx"
+# pygments_style = "sphinx"
```
adiroiban: Thanks for the update.
Why not just remove these 2 lines?

Is there any value in leaving them commented?

adiroiban:
Thanks for the update.

I triggerd the build so can check[ the new version on Read the docs.](https://towncrier--625.org.readthedocs.build/en/625/markdown.html)

> I don't think this is a news worthy change, but let me know if you want one.

If it's not news worthy, then we need a `newsfragment/625.misc` file.

This confirm that the release notes information was considred and it was decided not to add any explicit text.

As opposed of not having a newsfragment file and thinking that the author of the PR has forgot about the release notes :)

docs/conf.py
pradyunsg: That's the format of the conf.py file today -- with a bunch of unused commented out configuration variables. While I would strongly encourage doing something like https://github.com/pytest-dev/pytest/commit/05979bc4a97491c8b75f8baf95ec16f5b5516353, I didn't have much interaction with the maintainers prior to this and did not want to show up with a giant PR that changes everything. 😅

pradyunsg: Actually, I take that back. There's only a tiny bit of code at the top of the file -- Lemme remove this.

adiroiban:
This is subjective ... but I think that we can merge it.

Thanks :)

Issue #616, opened 2024-06-14
Release time?


SmileyChris:
We've got a big backlog of fixes in trunk now -- time for a release you think @adiroiban ?

Maybe #604 should be resolved first, but nothing else is outstanding. (if we do the precommit hooks updates, we should also update the version referenced in docs)

adiroiban:
Hi. I think #604 can be merged. There was no other feedback from other devs.

Do you have time to check the release process and try to do the release your self - https://towncrier.readthedocs.io/en/latest/release.html

In this way, we can check that the release notes make sense and anyone can do a release.

Otherwise, I can also trigger the release, no problem.

thanks

adiroiban:
I hope that we can have #622 merged by the end of the next week and then we can have a released.

@SmileyChris regarding #604 It looks line nobody else is interested in the `pre-commit` integration. I think that you can go ahead and make sure all checks pass on that PR and then we can have it merged.

 @bentheiii is there any particular feature that you want to have in the next release?

PR #627, opened 2024-07-28
♻️ Remove incremental

### LINKED ISSUE #308: Remove dependency on incremental in favor of general compatible solution

sigma67:
# Description

Fixes #308

This removes the dependency on `incremental` on `towncrier` itself, while keeping support for detecting `incremental` version in projects for which towncrier is used to build the release notes.

# Checklist


RELEASE.rst
```diff
@@ -19,9 +19,9 @@ The same branch is used for the release candidated and the final release.
 In the end, the release branch is merged into the main branch.
 
 Update the version to the release candidate with the first being ``rc1`` (as opposed to 0).
-In ``src/towncrier/_version.py`` the version is set using ``incremental`` such as::
+In ``pyproject.toml`` the version is set like::
```
adiroiban: I was hoping that we can set the version in single place, like `src/towncrier/_version.py` and the build system can automatically extract the version from there.

docs/conf.py
```diff
@@ -29,13 +29,13 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
+import importlib.metadata as importlib_metadata
```
adiroiban: Why not just use `from towncrier import __verison__` or something similar ?

src/towncrier/_project.py
```diff
@@ -63,56 +62,50 @@ def get_version(package_dir: str, package: str) -> str:
...
     # When no version if found, fall back to looking for the package in `package_dir`.
     module = _get_package(package_dir, package)
-
     version = getattr(module, "__version__", None)
-
     if not version:
-        raise Exception(
-            f"No __version__ or metadata version info for the '{package}' package."
-        )
+        try:
```
adiroiban: I this change needed ?

I was thinking that this was sorted out in https://github.com/twisted/towncrier/pull/502

src/towncrier/_shell.py
```diff
@@ -9,17 +9,24 @@
 
 from __future__ import annotations
 
+from importlib.metadata import PackageNotFoundError, version
+
 import click
 
-from ._version import __version__
```
adiroiban: Why not keep the import ?

It's just that now `__version__` will be a simple text.

adiroiban (commented):
Many thanks for picking up on this.

I just did a quick review before approving the CI run.

I have not yet fully reviewed the automated tests.

docs/conf.py
sigma67: The changes in `doc` were recycled but now they are no longer relevant. Reverted

RELEASE.rst
sigma67: The changes in `doc` were recycled but now they are no longer relevant. Reverted

src/towncrier/_project.py
sigma67: Right, I think it's just a merge artifact. Reverted

src/towncrier/_shell.py
sigma67: done

src/towncrier/test/test_packaging.py
```diff
@@ -1,22 +1,26 @@
...
-from incremental import Version
 from twisted.trial.unittest import TestCase
 
-from towncrier._version import _hatchling_version
+import towncrier
 
 
 class TestPackaging(TestCase):
-    def test_version_warning(self):
+    def test_version_attr(self):
```
adiroiban, line 10: I am not sure why this change is needed.

The previous test look good enough to me. :)

I am misssing something ?

src/towncrier/test/test_project.py
```diff
@@ -45,32 +44,65 @@ def test_tuple(self):
...
+    """
+                )
+            version = get_version(temp, "mytestprojinc")
+            self.assertEqual(version, "1.3.12rc1")
+
+    def test_not_incremental(self):
         """
-        An incremental Version __version__  is picked up.
+        An incremental-like Version __version__  is picked up, base takes unexpected
+        arguments.
```
adiroiban: Maybe something like this ?


```suggestion
        An excpettion is raised when the version could not be detected.
        For this test we use an incremental-like object,
        that has the `base` method but that method
        does not matches the return type for `incremental`.
```

adiroiban (changes requested):
Thanks for the changes.
They look good.

I left a few minor comments.

In order to merge this, the `mypy` checks must pass.

Also the coverage must pass

```
src/towncrier/_project.py      52      1     18      1    97%   90, 104->exit
```

src/towncrier/_project.py
```diff
@@ -82,37 +78,30 @@ def get_version(package_dir: str, package: str) -> str:
...
+        " If you can't provide that, use the --version argument and specify one."
     )
 
 
 def get_project_name(package_dir: str, package: str) -> str:
     module = _get_package(package_dir, package)
-
     version = getattr(module, "__version__", None)
+    # Incremental has support for package names, try duck-typing it.
+    with contextlib.suppress(AttributeError):
```
adiroiban, line 104: I think that we are missing a brach test here

src/towncrier/_project.py
```diff
@@ -82,37 +78,30 @@ def get_version(package_dir: str, package: str) -> str:
...
     if isinstance(version, tuple):
         return ".".join(map(str, version)).strip()
 
+    # Try duck-typing as an Incremental version.
+    if hasattr(version, "base"):
+        try:
+            version = str(version.base()).strip()
+            # Incremental uses `X.Y.rcN`.
+            # Standardize on importlib (and PEP440) use of `X.YrcN`:
+            return version.replace(".rc", "rc")  # type: ignore
```
adiroiban, line 90: I don't know why coverage is reporting this line as not covered.... we need to check

src/towncrier/test/test_project.py
sigma67:
```suggestion
        An exception is raised when the version could not be detected.
        For this test we use an incremental-like object,
        that has the `base` method, but that method
        does not match the return type for `incremental`.
```
fixed some typos

src/towncrier/_project.py
sigma67: added

src/towncrier/test/test_packaging.py
sigma67: The previous test was checking for the presence of an Incremental Version object (which obviously no longer exists in towncrier)

I've replaced it with a test to assert that a `DeprecationWarning` is present

src/towncrier/_project.py
sigma67: There was an indentation problem with a test, fixed now

sigma67 (commented):
thanks, mypy and coverage should be passing now

src/towncrier/_project.py
adiroiban: I still see this error reported 

> line 104 didn't return from function 'get_project_name' because the return on line 105 wasn't executed

image

sigma67: The miss indicator was a bit confusing for me, should be gone now hopefully

sigma67:
@adiroiban just to know a timeline, is there any ETA on the next release that could include this change?

This is just a question, no pressure

adiroiban:
We can do a release.
There is already a ticket for that here https://github.com/twisted/towncrier/issues/616

the important part is to have someone who cares for a new release and has some time to test the release

adiroiban:
Thanks for fixing all the checks.

As part of this PR we need to update the RELEASE.rst ... if we just merge, the information for RELEASE.rst will no longer be valid

I now see that RELEASE.rst changes were removed from this PR

https://github.com/twisted/towncrier/pull/491/files#diff-0a92214d83ea13ad864f4b1d6c0b5703956cf4bd927c116e75f68d4048e0744f

--------

My hope was that with this PR we can have `pyproject.toml` with

```
[metadata]
version = attr: towncrier.__version__
```

adiroiban:
I made a few small changes to the PR description... the important one is linking to the issue, so that when we merge this PR, the issue is automatically closed.

sigma67:
Ok, I'm a bit confused admittedly because in the prior MR it was stated explicitly to keep the version file.

Here: https://github.com/twisted/towncrier/pull/491#discussion_r1221613227

If we still want to move to pyproject.toml I'd prefer to do that in a separate MR.

What do you want changed in RELEASE.rst ? Should I update the version as well or is that done by the release manager?

adiroiban:
Right now the content of the RELEASE.rst is

https://github.com/twisted/towncrier/blob/4b58be5da5b1fabf38f47d5b777eb9c328046cda/RELEASE.rst?plain=1#L21-L24


----------

I think it should change to

```rst
 Update the version to the release candidate with the first being ``rc1`` (as opposed to 0). 
 In ``src/towncrier/_version.py`` the version is set using PEP440 a compliant string, such as:: 
  
     __version__ = "24.7.0rc1" 
```

---------

This revert is not quite OK https://github.com/twisted/towncrier/pull/627/commits/2cf4e76af7f16d8125118cf4fd47f0747741573f

Instead of updating the pyproject.toml we update the `_version.py` file.

We no longer want to use `incremental` . This is why `towncrier` own development documentation should no longer instruct developers to use `incremental`


---------

But I understand your concern.

The question is where to store the version:

* `pyproject.toml`
* `src/towncrier/_version.py`
* `src/towncier/__init__.py`

-----------

Since the version was already stored in `src/towncrier/_version.py` , my suggestion for this PR is to continue to store it there.

In a separate PR we can consider moving it inside `pyproject.toml` or somewhere else.

sigma67:
Ok my bad, I guess it only needed some changes instead of a revert. RELEASE.rst is updated.

adiroiban (approved):
Many thanks for the update.

---

I did a few change to release.rst

There was a section underline mismatch ... so I start doing the edit and then I realized that 19.9.0 is a better example as it shows that we don't add leasing 0 in the version ... but we use ISO date format with leading zero.

adiroiban:
I have enabled auto-merge on this.

and once this is merge I plan to do a new release.

Thanks again.

I am very happy to see that we no longer use incremental.

PR #629, opened 2024-07-31
Remove unused `wheel` build dependency

mgorny:
# Description
Remove the `wheel` dependency as it is not used anywhere.  Hatchling does not use `wheel` (and setuptools either bundled it or pulled it in automatically).

# Checklist


hynek (approved):

mgorny:
Thanks!

PR #630, opened 2024-07-31
Stop using deprecated packages in CI

hynek:
pep517 has been deprecated for a long time and we don't need it anymore, courtesy of `importlib.metadata`.

I appreciate that it's a bit icky since we cannot test it before actually releasing, and I really don't want to slow down a future release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,9 @@ defaults:
   run:
     shell: bash
 
+env:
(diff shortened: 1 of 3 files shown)


hynek:
I tried a fake version and looks good!
CleanShot 2024-07-31 at 08 59 58@2x


cf https://github.com/twisted/towncrier/actions/runs/10175573095/job/28143297512

adiroiban (approved):
Thanks for the cleanup. Looks good.

PR #633, opened 2024-07-31
automatically ignore template files

### LINKED ISSUE #632: Regression/breakage on 24.7.0 - check for invalid files
Hi,
thanks for great project!
When we upgraded to towncrier 24.7.0, we started getting this error in CI:
```
Looking at these files:
----
1. /home/runner/work/operator/operator/Cargo.lock
2. /home/runner/work/operator/operator/changelog.d/+url-safe-fingerprint.internal.md
3. /home/runner/work/operator/operator/changelog.d/544.changed.md
4. /home/runner/work/operator/operator/THIRDPARTY.yml
5. /home/runner/work/operator/operator/Cargo.toml
6. /home/runner/work/operator/operator/CHANGELOG.md
----
Error: Invalid news fragment name: changelog_template.jinja
If this filename is deliberate, add it to 'ignore' in your configuration.
```
We have `changelog_template.jinja` set as `template` in the `towncrier.toml` config file. I realize we can add it to `ignore`, but I checked other repositories and realized we've probably followed a pattern seen in other projects, where the template is inside the changes directory - as can be seen here:

- https://github.com/pytest-dev/pytest/tree/main/changelog
- https://github.com/python-attrs/attrs/tree/main/changelog.d

I assume it's a common pattern, and will break for them. Obviously each repo can fix it by adding `ignore` but I think it'd be better to handle it on towncrier side, potentially by either deprecating the feature (showing warning saying it will fail in next versions) or releasing in major version, another option is to make towncrier ignore the file used in template automatically.

Would be happy to send a PR implementing one of the options.

Related:
https://github.com/twisted/towncrier/pull/622
https://github.com/twisted/towncrier/issues/619

aviramha:
Automatically ignore template files

Closes #632

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -5,60 +5,6 @@ Release notes
 
 .. towncrier release notes start
 
-Towncrier 23.11.1.dev0 (2024-07-31)
-===================================
-
-Features
---------
-
-- ``towncrier build`` now handles removing news fragments which are not part of the git repository. For example, uncommitted or unstaged files. (`#357 <https://github.com/twisted/towncrier/issues/357>`_)
-- Inferring the version of a Python package now tries to use the metadata of the installed package before importing the package explicitly (which only looks for ``[package].__version__``). (`#432 <https://github.com/twisted/towncrier/issues/432>`_)
-- If no filename is given when doing ``towncrier`` create, interactively ask for the issue number and fragment type (and then launch an interactive editor for the fragment content).
-
-  Now by default, when creating a fragment it will be appended with the ``filename`` option's extension (unless an extension is explicitly provided). For example, ``towncrier create 123.feature`` will create ``news/123.feature.rst``. This can be changed in configuration file by setting `add_extension = false`.
-
-  A new line is now added by default to the end of the fragment contents. This can be reverted in the configuration file by setting `add_newline = false`. (`#482 <https://github.com/twisted/towncrier/issues/482>`_)
-- The temporary file ``towncrier create`` creates now uses the correct ``.rst`` or ``.md`` extension, which may help your editor with with syntax highlighting. (`#594 <https://github.com/twisted/towncrier/issues/594>`_)
-- Running ``towncrier`` will now traverse back up directories looking for the configuration file. (`#601 <https://github.com/twisted/towncrier/issues/601>`_)
-- The ``towncrier create`` action now uses sections defined in your config (either interactively, or via the new ``--section`` option). (`#603 <https://github.com/twisted/towncrier/issues/603>`_)
-- News fragments are now sorted by issue number even if they have non-digit characters.
-  For example::
-
-      - some issue (gh-3, gh-10)
-      - another issue (gh-4)
-      - yet another issue (gh-11)
-
-  The sorting algorithm groups the issues first by non-text characters and then by number. (`#608 <https://github.com/twisted/towncrier/issues/608>`_)
-- The ``title_format`` configuration option now uses a markdown format for markdown templates. (`#610 <https://github.com/twisted/towncrier/issues/610>`_)
-- newsfragment categories can now be marked with ``check = false``, causing them to be ignored in ``towncrier check`` (`#617 <https://github.com/twisted/towncrier/issues/617>`_)
-- ``towncrier check`` will now fail if any news fragments have invalid filenames.
-
-  Added a new configuration option called ``ignore`` that allows you to specify a list of filenames that should be ignored. If this is set, ``towncrier build`` will also fail if any filenames are invalid, except for those in the list. (`#622 <https://github.com/twisted/towncrier/issues/622>`_)
-
-
-Bugfixes
---------
-
-- Add explicit encoding to read_text. (`#561 <https://github.com/twisted/towncrier/issues/561>`_)
-- The default Markdown template now renders a title containing the release version and date, even when the `name` configuration is left empty. (`#587 <https://github.com/twisted/towncrier/issues/587>`_)
-- Orphan news fragments, fragments not associated with an issue, consisting of only digits (e.g. '+12345678.feature') now retain their leading marker character. (`#588 <https://github.com/twisted/towncrier/issues/588>`_)
-- Orphan news fragments, fragments not associated with an issue, will now still show in categories that are marked to not show content, since they do not have an issue number to show. (`#612 <https://github.com/twisted/towncrier/issues/612>`_)
-
-
-Improved Documentation
-----------------------
-
-- Clarify version discovery behavior. (`#432 <https://github.com/twisted/towncrier/issues/432>`_, `#602 <https://github.com/twisted/towncrier/issues/602>`_)
-- The tutorial now introduces the `filename` option in the appropriate paragraph and mentions its default value. (`#586 <https://github.com/twisted/towncrier/issues/586>`_)
-- Add docs to explain how ``towncrier create +.feature.rst`` (orphan fragments) works. (`#589 <https://github.com/twisted/towncrier/issues/589>`_)
-
-
-Misc
-----
-
-- `#491 <https://github.com/twisted/towncrier/issues/491>`_, `#561 <https://github.com/twisted/towncrier/issues/561>`_, `#562 <https://github.com/twisted/towncrier/issues/562>`_, `#568 <https://github.com/twisted/towncrier/issues/568>`_, `#569 <https://github.com/twisted/towncrier/issues/569>`_, `#571 <https://github.com/twisted/towncrier/issues/571>`_, `#574 <https://github.com/twisted/towncrier/issues/574>`_, `#575 <https://github.com/twisted/towncrier/issues/575>`_, `#582 <https://github.com/twisted/towncrier/issues/582>`_, `#591 <https://github.com/twisted/towncrier/issues/591>`_, `#596 <https://github.com/twisted/towncrier/issues/596>`_, `#597 <https://github.com/twisted/towncrier/issues/597>`_, `#625 <https://github.com/twisted/towncrier/issues/625>`_, `#629 <https://github.com/twisted/towncrier/issues/629>`_, `#630 <https://github.com/twisted/towncrier/issues/630>`_
-
-
 towncrier 23.11.0 (2023-11-08)
 ==============================
 
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -5,6 +5,60 @@ Release notes
 
 .. towncrier release notes start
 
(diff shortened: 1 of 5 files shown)


aviramha:
Not sure if I should create a changelog that is `fixed` or `added`

src/towncrier/test/test_build.py
```diff
@@ -1631,6 +1631,29 @@ def test_invalid_fragment_name(self, runner):
...
+    @with_project(
+        config="""
+        [tool.towncrier]
+        package = "foo"
+        template = ["template.jinja"]
+        """
+    )
+    def test_ignored_template_string(self, runner):
+        """
+        Files used in `template` are automatically ignored.
```
adiroiban, line 1643: Can there be more template files ?

I am reading the docs from [here](https://towncrier.readthedocs.io/en/latest/configuration.html) 

```
template

Path to the template for generating the news file.
If the path looks like <some.package>:<filename.ext>, it is interpreted as a template bundled with an installed Python package.
"towncrier:default.rst" by default unless filename ends with .md, in which case the default is "towncrier:default.md".
```

src/towncrier/test/test_build.py
```diff
@@ -1631,6 +1631,29 @@ def test_invalid_fragment_name(self, runner):
         self.assertEqual(1, result.exit_code, result.output)
         self.assertIn("Invalid news fragment name: feature.124", result.output)
 
+    @with_project(
+        config="""
+        [tool.towncrier]
+        package = "foo"
+        template = ["template.jinja"]
```
adiroiban: I don't understand why this is not the full path to the template

```suggestion
        template = "foo/newsfragments/template.jinja"
```

src/towncrier/_builder.py
```diff
@@ -115,6 +115,8 @@ def find_fragments(
     If strict, raise ClickException if any fragments have an invalid name.
     """
     ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
+    if isinstance(config.template, str):
```
adiroiban, line 118: Why we need to do this check ?

Can't we just alwasy add it.

I understand that sometimes this can be `towncrier:default.rst` ... but I think that it's ok to just add that value...

I don't expact that people will not want to ignore a file with than name.

... just to keep it simple  and don't worry too much about edge cases that will never happen in real life :)

adiroiban (changes requested):
Thanks for the quick update.
This is a good start.

We can consider it a bugfix.

The tests needs to pass in order to merge this.

Thanks again

src/towncrier/test/test_build.py
aviramha: I think this is an irrelevant case, because to my understanding it'd pull the said file from a Python package, meaning it's not in your current directory/where changes lie.

src/towncrier/_builder.py
aviramha: According to the types, config.template can be `Tuple[str,str]` so that might cause troubles down the line.

NEWS.rst
```diff
@@ -5,6 +5,60 @@ Release notes
 
 .. towncrier release notes start
 
+Towncrier 23.11.1.dev0 (2024-07-31)
```
adiroiban: I think that you have accidentally applied this change ... which was supposed to be only for testing.

src/towncrier/test/test_build.py
adiroiban: Why the test was passing with `template = ["template.jinja"]`

Is the test valid ?

Was the test failing before applying the patch in `_builder.py` ?

adiroiban (changes requested):

src/towncrier/_builder.py
adiroiban: As long as the test pass, we should not worry about that and try to write the most simple code possible.

No need to write code for imaginary cases for which we don't have a real world use case.

If this check is indeede needed, I think that it needs a comment to explain why we can end up with a tuple

src/towncrier/test/test_build.py
aviramha: Yes, it was failing.

src/towncrier/_builder.py
aviramha: I don't know why we can end with a tuple, as I am first time diving into the source code but according to the dataclass of Config:
```python
@dataclasses.dataclass
class Config:
    template: str | tuple[str, str]
```
should I just disregard it?

src/towncrier/test/test_build.py
adiroiban: Perfect. I was a bit confuse of why we have the initial version of the test.

I double check and it looks good.

src/towncrier/_builder.py
adiroiban: I have applied this patch and I can see that all the tests still pass via `trial src/twisted`

```diff
diff --git a/src/towncrier/_builder.py b/src/towncrier/_builder.py
index dabd64a..78d8bbe 100644
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -115,8 +115,7 @@ def find_fragments(
     If strict, raise ClickException if any fragments have an invalid name.
     """
     ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
-    if isinstance(config.template, str):
-        ignored_files.add(config.template)
+    ignored_files.add(config.template)
     if config.ignore:
         ignored_files.update(filename.lower() for filename in config.ignore)
```

adiroiban: but it looks like we are missing an automated test.

I can see that settings is set to a tuple when a package is provided

https://github.com/twisted/towncrier/blob/68a7a3b7e2f04a32a2eea670a5ad5c38913ff470/src/towncrier/_settings/load.py#L187-L203

aviramha: I have applied the change.
I got to say I highly disagree but you're the boss here :)

src/towncrier/_builder.py
```diff
@@ -115,6 +115,7 @@ def find_fragments(
     If strict, raise ClickException if any fragments have an invalid name.
     """
     ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
+    ignored_files.add(config.template)
```
```suggestion
    if isinstance(str, config.template)
        # Template can be a tuple of (package_name, resource_name).
        #
        # See https://github.com/twisted/towncrier/issues/634
        ignored_files.add(config.template)
```

src/towncrier/newsfragments/632.bugfix.rst
```diff
@@ -0,0 +1 @@
+Automatically ignore template files
```
adiroiban: Can we have more than one template ?

```suggestion
When the template file is stored in the same directory with the news fragments, it is automatically ignored when checking for valid fragment file names.
```

adiroiban:
The documentation also needs to be updated.

> Some filenames such as .gitignore and README are automatically ignored. However, **if a custom template is stored in the news fragment directory, you should add it to this list**.

Maybe something like this

```diff
-Some filenames such as .gitignore and README are automatically ignored. However, if a custom template is stored in the news fragment directory, you should add it to this list.
+Some filenames such as .gitignore, README.rst. README.md, and the template file, are automatically ignored.
```

src/towncrier/newsfragments/632.bugfix.rst
aviramha: You're right, we can't afaik

adiroiban (approved):
It looks good. Many thanks for the update.

src/towncrier/_builder.py
```diff
@@ -115,6 +115,11 @@ def find_fragments(
     If strict, raise ClickException if any fragments have an invalid name.
     """
     ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
+    if isinstance(str, config.template)
```
adiroiban: my bad
```suggestion
    if isinstance(str, config.template):
```

src/towncrier/_builder.py
```diff
@@ -115,6 +115,11 @@ def find_fragments(
     If strict, raise ClickException if any fragments have an invalid name.
     """
     ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
+    if isinstance(str, config.template):
```
```suggestion
    if isinstance(config.template, str):
```

adiroiban:
Sorry for the mess. I will try to get this this merged and then look at doing a new release.

src/towncrier/_builder.py
```diff
@@ -115,6 +115,11 @@ def find_fragments(
     If strict, raise ClickException if any fragments have an invalid name.
     """
     ignored_files = {".gitignore", ".keep", "readme", "readme.md", "readme.rst"}
+    if isinstance(config.template, str):
+        # Template can be a tuple of (package_name, resource_name).
+        #
+        # See https://github.com/twisted/towncrier/issues/634
+        ignored_files.add(config.template)
```
adiroiban, line 122: This is not ok.

We should extract the filename from the filepath.

freakboy3742:
I'm not sure this has been fixed. I'm seeing CI failures with the same error using 24.7.1:

```
Run tox -e towncrier-check
towncrier-check: install_deps> python -I -m pip install towncrier==24.7.1
towncrier-check: freeze> python -m pip freeze --all
towncrier-check: click==8.1.7,Jinja2==3.1.4,MarkupSafe==2.1.5,pip==24.1,towncrier==24.7.1
towncrier-check: commands[0]> python -m towncrier.check --compare-with origin/main
Looking at these files:
----
Error: Invalid news fragment name: template.rst
If this filename is deliberate, add it to 'ignore' in your configuration.
1. /home/runner/work/briefcase/briefcase/automation/pyproject.toml
2. /home/runner/work/briefcase/briefcase/changes/1935.misc.rst
3. /home/runner/work/briefcase/briefcase/tox.ini
4. /home/runner/work/briefcase/briefcase/pyproject.toml
----
towncrier-check: exit 1 (0.10 seconds) /home/runner/work/briefcase/briefcase> python -m towncrier.check --compare-with origin/main pid=2026
  towncrier-check: FAIL code 1 (2.43=setup[2.33]+cmd[0.10] seconds)
  evaluation failed :( (3.08 seconds)
```

* https://github.com/beeware/gbulb/actions/runs/10239027441/job/28324134923
* https://github.com/beeware/rubicon-objc/actions/runs/10238982532/job/28324041023
* https://github.com/beeware/toga-chart/actions/runs/10238980336/job/28324037142
* https://github.com/beeware/travertino/actions/runs/10238976056/job/28325827449
* https://github.com/beeware/toga/actions/runs/10238971800/job/28324020117
* https://github.com/beeware/briefcase/actions/runs/10238950647/job/28323974328

PR #639, opened 2024-07-31
Fix for ignoring the template based only on filename.

### LINKED ISSUE #638: Ignore template only base on filename not an full path
This is not ok.

We should extract the filename from the filepath.

_Originally posted by @adiroiban in https://github.com/twisted/towncrier/pull/633#discussion_r1698695121_

adiroiban:
# Description

Fixes #638

This is take-2 for the fix from #633

### DIFF
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -1604,9 +1604,7 @@ class TestCli(TestCase):
         with open("foo/newsfragments/.gitignore", "w") as f:
             f.write("gitignore is automatically ignored")
 
-        result = runner.invoke(
-            _main, ["--draft"]
-        )
+        result = runner.invoke(_main, ["--draft"])
         self.assertEqual(0, result.exit_code, result.output)
 
     @with_project(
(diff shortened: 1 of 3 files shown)


src/towncrier/test/test_build.py
```diff
@@ -1604,9 +1604,7 @@ def test_ignored_files(self, runner):
         with open("foo/newsfragments/.gitignore", "w") as f:
             f.write("gitignore is automatically ignored")
 
-        result = runner.invoke(
-            _main, ["--draft", "--date", "01-01-2001", "--version", "1.0.0"]
-        )
+        result = runner.invoke(_main, ["--draft"])
```
adiroiban, line 1607: the version and date are not needed in these tests

src/towncrier/test/test_build.py
```diff
@@ -1625,34 +1623,45 @@ def test_invalid_fragment_name(self, runner):
...
         self.assertEqual(1, result.exit_code, result.output)
         self.assertIn("Invalid news fragment name: feature.124", result.output)
 
     @with_project(
         config="""
         [tool.towncrier]
         package = "foo"
-        template = "foo/newsfragments/template.jinja"
+        template = "foo/newsfragments/template.j2"
+        ignore = ["placeholder-to-trigger-strict-checks.txt"]
```
adiroiban, line 1635: @aviramha  the extra file was ignored by default for `towncrier build` when "ignore" was not defined.

I am not sure how this tests was failing on your side.

Without adding the `ignore` config, the tests was passing , even with the code from 24.7.0

adiroiban:
This is ready for review.

Thanks

aviramha (approved):
Looks good!

PR #640, opened 2024-08-01
move version into pyproject.toml

sigma67:
# Description

Fixes #

# Checklist

* [x ] Make sure changes are covered by existing or new tests.
* [ x] For at least one Python version, make sure test pass on your local environment.
* [x ] Create a file in `src/towncrier/newsfragments/`. Briefly describe your
* [x ] Make sure all GitHub Actions checks are green (they are automatically checking all of the above).
* [x ] Ensure `docs/tutorial.rst` is still up-to-date.
* [- ] If you add new **CLI arguments** (or change the meaning of existing ones), make sure `docs/cli.rst` reflects those changes.
* [x ] If you add new **configuration options** (or change the meaning of existing ones), make sure `docs/configuration.rst` reflects those changes.


sigma67:
@adiroiban as requested this is the separate PR to move the version into ``pyproject.toml``.

I suggest we also remove the deprecated ``__version__`` attribute as it has been deprecated for three releases now.

src/towncrier/newsfragments/640.misc
```diff
@@ -0,0 +1,3 @@
+Moved towncrier version definition from src/towncrier/_version.py to pyproject.toml
+
+Removed deprecated __version__ attribute from towncrier/__init__.py
```
adiroiban: This should be in a separate 640.removal file.

```suggestion
towncrier.__version__ was removed, after being deprecated in 23.6.0.
Use importlib.metadata to get the towncrier's version.
```

src/towncrier/_project.py
```diff
@@ -67,6 +69,10 @@ def get_version(package_dir: str, package: str) -> str:
     if version := _get_metadata_version(package):
         return version
 
+    with contextlib.suppress(PackageNotFoundError):
```
adiroiban: Do we need this code?

What does it to, on top of the existing `_get_metadata_version` ?

Maybe move this to `_get_metadata_version`

If this needs to stay here, then it needs a comment to explain why `_get_metadata_version` is not good enough here and why we get the version again via `importlib.metadata.

adiroiban (changes requested):
Thanks. It looks very good.

Just a comment regarding `_get_metadata_version` and the removal information should have a separate fragment.

Other than that, all good.

Thanks again

src/towncrier/newsfragments/640.misc
sigma67: Not sure why you put twisted, shouldn't it be `towncrier.__version__` ?

src/towncrier/_project.py
sigma67: The test `test_version_from_metadata` will fail without this code.

The existing function `_get_metadata_version_` uses `importlib.metadata.packages_distributions`, which seems to not contain towncrier at test runtime. Not sure why, but I also don't know why this method of getting the version is used instead of straight up passing `package` to `importlib.metadata.version`.

I can move the check into the function, maybe inside the `if not distribution_names` block. Not sure. 

What do you think?

src/towncrier/newsfragments/640.misc
adiroiban: Is `twisted` as most of my work is on the `twisted/twisted` repo.

I have fixed it. Thanks

src/towncrier/_project.py
adiroiban: I think that `_get_metadata_version` should do everything.

So the code should be moved there.

Maybe also remove `importlib.metadata.packages_distributions` and replace it with the current method.

Do we need both methods? Just asking.

I am not very familiar with the `importlib` API  ... but as long as the tests pass, we should be ok.

sigma67: Ok, I thought there might be some reason why it was done this way.

I replaced it as you suggested.

src/towncrier/newsfragments/640.misc
```diff
@@ -0,0 +1,4 @@
+Moved towncrier version definition from src/towncrier/_version.py to pyproject.toml
+
+towncrier.__version__ was removed, after being deprecated in 23.6.0.
```
adiroiban: these 2 lines needs to be in a separate newsfragment file of type `.removal`

sigma67: done

adiroiban (approved):
Thanks for the cleanup.

-------

We should not worry about keeping "zombie" code.

If we think that some part of the code is not needed, we should just delete it.

As long as the tests still pass, it should be ok.

PR #644, opened 2024-08-02
Add support for wildcards in ignore

### LINKED ISSUE #642: Allow wildcards in "ignore" configuration option
Since version 24.7.0 (and PR #622) there is an "ignore" configuration option that allows users to specify files that are known to have "wrong" filenames but still belong in the fragments directory.

It would be very helpful if this "ignore" option supported wildcards in filenames. For example:

```
[tool.towncrier]
    ignore = [".gitempty", "useful_file_num*.ext"]
```

Thanks!

eigenbrot:
# Description

Fixes #642 

The `ignore` config option now supports wildcard matching. Any wildcards understood by [fnmatch](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch) are supported.

# Checklist

### DIFF
--- a/src/towncrier/test/test_check.py
+++ b/src/towncrier/test/test_check.py
@@ -476,7 +476,11 @@ class TestChecker(TestCase):
         """
         When `ignore` is set in config, files with those names are ignored.
         """
-        create_project("pyproject.toml", extra_config='ignore = ["template.jinja"]')
+        create_project(
+            "pyproject.toml",
+            extra_config='ignore = ["template.jinja", "star_wildcard*", "question_wildcard_?", '
+            '"seq_wildcard_[ab]"]',
+        )
 
         write(
             "foo/newsfragments/124.feature",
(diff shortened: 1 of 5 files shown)


src/towncrier/test/test_check.py
```diff
@@ -476,14 +476,28 @@ def test_ignored_files(self, runner):
...
+        write("foo/newsfragments/star_wildcard_foo", "Manually ignored with * wildcard")
+        write("foo/newsfragments/STAR_WILDCARD_bar", "Manually ignored with * wildcard")
+        write(
+            "foo/newsfragments/question_wildcard_1", "Manually ignored with ? wildcard"
+        )
+        write(
+            "foo/newsfragments/QUESTION_WILDCARD_1", "Manually ignored with ? wildcard"
+        )
+        write("foo/newsfragments/seq_wildcard_a", "Manually ignored with [] wildcard")
+        write("foo/newsfragments/SEQ_WILDCARD_b", "Manually ignored with [] wildcard")
```
adiroiban: I think that we can keep it simple and have a single test example... otherwise, this test becomes too long and hard to follow.

If you want to check all possible values supported by fmatch, is best to write a separate tests... but I think that a single update is good enough.

```suggestion
```

src/towncrier/test/test_check.py
```diff
@@ -476,14 +476,28 @@ def test_ignored_files(self, runner):
         """
         When `ignore` is set in config, files with those names are ignored.
         """
-        create_project("pyproject.toml", extra_config='ignore = ["template.jinja"]')
+        create_project(
+            "pyproject.toml",
+            extra_config='ignore = ["template.jinja", "star_wildcard*", "question_wildcard_?", '
+            '"seq_wildcard_[ab]"]',
```
adiroiban: Just to keep it simple
```suggestion
            extra_config='ignore = ["template.jinja", "star_wildcard*"]',
```

src/towncrier/test/test_check.py
```diff
@@ -476,14 +476,28 @@ def test_ignored_files(self, runner):
         """
         When `ignore` is set in config, files with those names are ignored.
```
adiroiban, line 477: Maybe just update the scope of this test

```suggestion
        When `ignore` is set in config, files with those names are ignored.
        Configuration supports globbing expressions.
```

adiroiban (approved):
It looks good. Thanks. Only minor comments.

Pleas check my feedback.

If you want to keep the tests as they are now, that's fine , and we can merge it.

Thanks again

eigenbrot:
Thanks for the comments, @adiroiban, I've updated the PR as you suggested.

PR #645, opened 2024-08-02
add utility function to calculate underline size for category names

### LINKED ISSUE #626: 🐛 Bug - Missing underline character for emojis
Whenever running the `towncrier` command for my release notes, the changelog.rst file that's generated for it is missing an underline for every emoji character that's used.

pyproject.toml setting looks as such:

```TOML
[tool.towncrier]
directory = "CHANGES"
filename = "CHANGELOG.rst"
package = "mypackage"
title_format = "mypackage v{version} ({project_date})"
underlines = ["-", "~", "^"]

    [[tool.towncrier.type]]
    directory = "bugfix"
    name = "🐛 Bugfixes"
    showcontent = true
```

changelog.rst looks as such:

```rst

🐛 Bugfixes
~~~~~~~~~~
```

jacobgulan:
# Description

Fixes #626 

Added a helper function to calculate the underline size for the jinja templates. Specifically addresses not enough underline characters being produced for category names that include wide characters like emojis.

# Checklist

* [X] For at least one Python version, make sure test pass on your local environment.
* [X] Create a file in `src/towncrier/newsfragments/`. Briefly describe your

### DIFF
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -204,20 +204,45 @@ def find_fragments(
     return content, fragment_files
 
 
-def get_underline_length(text: str) -> int:
+def get_dict_length(obj: dict) -> int:
     """
-    Given `text` determine the underline length needed for the reStructuredText output.
+    Gets the sum of the underline lengths for all keys and values in a dictionary.
+    """
+    return sum(
+        get_underline_length(key) + get_underline_length(value)
+        for key, value in obj.items()
+    )
+
+
+def get_list_length(obj: list) -> int:
+    """
+    Gets the sum of the underline lengths for all items in a list.
+    """
+    return sum(get_underline_length(item) for item in obj)
+
+
+def get_string_length(text: str) -> int:
+    """
+    Determines the amount of characters needed to underline a string.
+    """
+    return sum(
+        2 if unicodedata.east_asian_width(char) in ("W", "F") else 1 for char in text
+    )
+
+
+def get_underline_length(obj: str | dict | list) -> int:
+    """
+    Given `obj` determine the underline length needed for the reStructuredText output.
 
     Particularly helps determine if an extra underline is needed for wide characters like emojis.
     """
-    underline_length: int = 0
-    for char in text:
-        if unicodedata.east_asian_width(char) in ("W", "F"):
-            underline_length += 2
-        else:
-            underline_length += 1
-
-    return underline_length
+    if isinstance(obj, dict):
+        return get_dict_length(obj)
+    elif isinstance(obj, list):
+        return get_list_length(obj)
+    elif isinstance(obj, str):
+        return get_string_length(obj)
+    raise ValueError("Object must be a string, list, or dictionary.")
 
 
 def indent(text: str, prefix: str) -> str:
@@ -262,10 +287,6 @@ def split_fragments(
                 # the content. If there isn't an issue, still add the content so that
                 # it's recorded.
                 content = ""
-
-            definitions[category]["underline_length"] = get_underline_length(
-                definitions[category]["name"]
-            )
 
             texts = section.setdefault(category, {})
 
@@ -432,9 +453,7 @@ def render_fragments(
         top_underline=top_underline,
         get_indent=get_indent,  # simplify indentation in the jinja template.
         issues_by_category=issues_by_category,
-        versiondata_name_underline_length=get_underline_length(
-            versiondata.get("name", "")
-        ),
+        get_underline_length=get_underline_length,  # helps determine length for non-ascii chars
     )
 
     for line in res.split("\n"):
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -20,7 +20,7 @@ from jinja2 import Template
 from towncrier._settings.load import Config
 
 
-UnderlineLengthType = float | str | int | dict[str, Any] | list[Any]
+UnderlineLengthType = float | str | int | dict[Any, Any] | list[Any]
 
 
 # Returns issue, category and counter or (None, None, None) if the basename
@@ -207,7 +207,7 @@ def find_fragments(
     return content, fragment_files
 
 
-def get_dict_length(obj: dict[str, Any]) -> int:
+def get_dict_length(obj: dict[UnderlineLengthType, UnderlineLengthType]) -> int:
     """
     Gets the sum of the underline lengths for all keys and values in a dictionary.
     """
@@ -217,7 +217,7 @@ def get_dict_length(obj: dict[str, Any]) -> int:
     )
 
 
-def get_list_length(obj: list[Any]) -> int:
+def get_list_length(obj: list[UnderlineLengthType]) -> int:
     """
     Gets the sum of the underline lengths for all items in a list.
     """
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -263,7 +263,6 @@ def split_fragments(
                 # it's recorded.
                 content = ""
 
-            # Calculate the underline size for the name of the category.
             definitions[category]["underline_size"] = get_underline_size(
                 definitions[category]["name"]
             )
@@ -433,6 +432,7 @@ def render_fragments(
         top_underline=top_underline,
         get_indent=get_indent,  # simplify indentation in the jinja template.
         issues_by_category=issues_by_category,
+        versiondata_name_underline_size=get_underline_size(versiondata.get("name", "")),
     )
 
     for line in res.split("\n"):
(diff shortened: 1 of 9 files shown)


src/towncrier/_builder.py
```diff
@@ -224,6 +241,11 @@ def split_fragments(
                 # it's recorded.
                 content = ""
 
+            # Calculate the underline size for the name of the category.
```
adiroiban: I don't think that this comment is needed.

The code is pretty clear.

```suggestion
```

src/towncrier/newsfragments/626.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed section names not having enough underline characters for emojis.
```
adiroiban: Maybe having this read like a release notes 

the current version looks like a commit message.

This information is for end users... and the end result is that they can now use emoji in category names :)

```suggestion
For reStructuredText format, you can now use emojis in category names.
In previous versions, the generated underline was not matching the title width. 
```

src/towncrier/test/test_builder.py
```diff
@@ -208,3 +208,13 @@ def test_ordering(self):
             - Added Fish
 """
         )
+
+
+class TestUtilityFunctions(TestCase):
```
adiroiban: Very minor comment, but I think that we can have these tests in the same test case as the other tests.

------

These are unit tests, but we also need an integration/functional/end to end test to check that the default templates work.

src/towncrier/_builder.py
```diff
@@ -181,6 +182,22 @@ def find_fragments(
...
 
+def get_underline_size(text: str) -> int:
+    """
+    Given `text` determine the underline size needed for the reStructuredText output.
+
+    Particularly helps determine if an extra underline is needed for wide characters like emojis.
+    """
+    underline_size: int = 0
+    for char in text:
+        if unicodedata.east_asian_width(char) in ("W", "F"):
```
adiroiban, line 193: We also need a integration test.

If for example we have something like `if False:`
the existing tests will still pass with success. 

Please consider creating a test similar to this one, but in which a custom category is used, one that has unicode in the name.

In this way, we can make sure that this functionality will continue to work in future versions.

https://github.com/twisted/towncrier/blob/fc3a2556408bcdb0237e54c4ff0773db364c554a/src/towncrier/test/test_build.py#L1042-L1090

adiroiban (changes requested):
Change looks good.

To merge this we need mypy fixed and at least one integration/end to end test.

Many thanks for the PR

src/towncrier/_builder.py
adiroiban: We also need a integration test.

If for example we have something like `if False:`
the existing tests will still pass with success. 

Please consider creating a test similar to this one, but in which a custom category is used, one that has unicode in the name.

In this way, we can make sure that this functionality will continue to work in future versions.

https://github.com/twisted/towncrier/blob/fc3a2556408bcdb0237e54c4ff0773db364c554a/src/towncrier/test/test_build.py#L1042-L1090

adiroiban:
Note that the underline is also computed here ... line 237 


https://github.com/twisted/towncrier/blob/afa6cc0c5340df2fa2c10462f844a794e003cfdc/src/towncrier/build.py#L230-L239

src/towncrier/_builder.py
jacobgulan: Added

src/towncrier/test/test_builder.py
jacobgulan: Done

src/towncrier/newsfragments/626.bugfix.rst
jacobgulan: Done

src/towncrier/_builder.py
jacobgulan: Removed

jacobgulan:
Wanted to discuss the mypy change before touching it. It seems to be complaining that I'm trying to assign "underline_size" to a Mapping type. Mapping types are read-only. Therefore, mypy fails. I can update the type to a dict and mypy will likely be happy again, though I'm not sure that'd be the right approach.

src/towncrier/test/test_build.py
```diff
@@ -1398,6 +1398,58 @@ def test_default_start_string_markdown(self, runner):
...
+
+          [[tool.towncrier.type]]
+          directory = "bugfix"
+          name = "🐛 Bugfixes"
+          showcontent = true
+        """
+    )
+    def test_underline_size_unicode(self, runner):
+        """
+        TODO: Add better description
```
adiroiban: Maybe something like this.
```suggestion
        The news file can be generaged for project names,
        and categories that contains emoji in their names.
        This is a test for generating the RST section
        underline to meet the docutils requirements. 
```

jacobgulan: Whoops, totally forgot to come back around for that. Thoughts on my mypy comment above?

src/towncrier/_builder.py
```diff
@@ -246,6 +263,10 @@ def split_fragments(
                 # it's recorded.
                 content = ""
 
+            definitions[category]["underline_size"] = get_underline_size(
```
adiroiban: The big problem with this code, is that the underline size should not be computed inside the `split_fragments` function

also, maybe call it `underline_lenght` ? but this is a very minor comment

adiroiban (changes requested):
Thanks for the update. It's a good start.

Please consider using `render_fragments` to create the data for the jinja template.

Inside `render_fragments` we can create a copy of `definitions` and update that copy and then pass it to jinja template rendering

so we would have something like `definitions=updated_definitions` in the call below

https://github.com/twisted/towncrier/blob/4aa8174dc7e843e7b71fc3092216dc2915065135/src/towncrier/_builder.py#L405-L409


this should clean the code and should also make mypy happy

it's best not to randomly update the configuration object or to update the configuration object after it was loaded from the config file

jacobgulan:
Appreciate the feedback. I switched directions on the implementation after your last comment and seeing a better way that should prove to be more reliable and maintainable.

Instead of creating a copy of definitions and then calculating the underline size for that modified dictionary, I'm instead just passing the jinja template the `get_underline_length()` function itself now and calculating the length with it inside the .rst files.

Added some helper functions to recursively handle other types like dicts & lists which proved necessary for some of the other integration tests, but the behavior is working as expected now without modifying the original definitions mapping.

adiroiban:
Thanks for the update.

The data exposed to the jinja template is part of our public API.
We need to double check that it make sense, as deprecating this type of code is very hard.

--------

For a template design , I like the first way of exposing the underline lenght better.

As a general rule for template design, I try to avoid calling methods/functions from the template... but I can't tel if this is good or bad.
I feel that doing function calls in the templates, makes the template harder to read.

jacobgulan:
I personally think that reading a function is easier to follow - if the name is intuitive - than indexing off variables. I would say it also makes sense to consolidate calculating length to one function rather than mixing get_underline_length() & the jinja pipe operation.

That said, I can also see the argument to maintain what's already present and just add another field to it. That second approach would look something like this:
```
  updated_definitions = {
      category: {**value, "underline_length": get_underline_length(value["name"])}
      for category, value in definitions.items()
  }
```
Couldn't use copy.deepcopy() as mypy screams about passing the wrong argument.

Let me know what you think

adiroiban:
We can go with a separate function... and later we can look at making it available as a filter.

Right now, we use the jinja `Template()` API instead of the `Environment()` API and I don't know how to add an environment to an existing template... so that we can add a custom filter to the template.

I know that you can add custom filters to environments, but not to templates.

------

The thing is that jinja2 is a beast of its own... with inheritence, conditional statements, variable assignations ...  and I have seen a lot jinja2 templates that are very hard to understand and update.

This is why I prefer to go with the pre-computed view-model approach, so that in the template, I only have loops and simple variables insertion.

src/towncrier/templates/hr-between-versions.rst
```diff
@@ -1,22 +1,22 @@
 {% if render_title %}
 {% if versiondata.name %}
 {{ versiondata.name }} {{ versiondata.version }} ({{ versiondata.date }})
-{{ top_underline * ((versiondata.name + versiondata.version + versiondata.date)|length + 4)}}
+{{ top_underline * (get_underline_length(versiondata.name + versiondata.version + versiondata.date) + 4)}}
```
adiroiban, line 4: One idea is a function like this

```diff
-{{ versiondata.name }} {{ versiondata.version }} ({{ versiondata.date }})
-{{ top_underline * (get_underline_length(versiondata.name + versiondata.version + versiondata.date) + 4)}}
+{{ top_section(versiondata.name, versiondata.version) }}
```

this function will join the arguments with a single space, and will generate 2 lines.

This looks to me less error prone

just an idea

PR #650, opened 2024-08-05
Enforce fragment name structure

### LINKED ISSUE #649: Enforce fragment name structure
We're using Jira, and want all fragments to be named according to the issue numbering format (e.g. `ABCD-12345.feature.md`).
Is there any way to configure towncrier to enforce the fragment file name somehow? (e.g. supplying a regex)

If there isn't one, I may be able to contribute it.

dorschw:
# Description

Fixes #649

Add a config for enforcing the fragment filename stems, using regex.

# Checklist

### DIFF
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -55,7 +55,7 @@ class Config:
     create_eof_newline: bool = True
     create_add_extension: bool = True
     ignore: list[str] | None = None
-    fragment_filename_stem_pattern: str | None = None
+    issue_pattern: str = ""
 
 
 class ConfigError(ClickException):
--- a/src/towncrier/_settings/load.py
+++ b/src/towncrier/_settings/load.py
@@ -55,6 +55,7 @@ class Config:
     create_eof_newline: bool = True
     create_add_extension: bool = True
     ignore: list[str] | None = None
+    fragment_filename_stem_pattern: str | None = None
 
 
 class ConfigError(ClickException):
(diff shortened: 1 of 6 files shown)


src/towncrier/_settings/load.py
```diff
@@ -55,6 +55,7 @@ class Config:
     create_eof_newline: bool = True
     create_add_extension: bool = True
     ignore: list[str] | None = None
+    fragment_filename_stem_pattern: str | None = None
```
dorschw: we could find it a shorter name

adiroiban: My suggetion is `issue_pattern` 

Inside the code and the documentation, we should refer to this as the "issue ID" , or just "issue" ... it used to be called "ticket" but it was recently renamed to "issue"

adiroiban: Just asking ... can we have this as just `str` ... with default value as empty string?

It should simplify the typing for this variable.

adiroiban:
Thanks. Changes looks good.

We need all the tests to pass in order to merge this.

dorschw:
Thanks @adiroiban, any idea why it fails with Linux PyPy/CPython 3.8? 
Using linux and Python 3.8, all tests pass locally using `trial`.

adiroiban:
I don't know why it fails.  Do you have `3.8.18` on your local system ?

dorschw:
got it (it was `removesuffix`, probably some env mess on my machine made it run on non-`3.8`), please trigger CI again and it should be good to go.

adiroiban (approved):
Looks good. Thanks

PR #651, opened 2024-08-05
Fix regression introduced by #610

### LINKED ISSUE #648: Markdown lines should not have '# ' prepended
if using Markdown, each title has a '# ' prepended (build.py). However, this makes it a top-line header, and the example creates titles like:

```md
# ## <rest of title>
```

as opposed of:

```md
## <rest of title>
```

Certainly a few ways to fix this, such as a config like "default_md_prefix" which would be "# " and could be overridden to "" to ensure backwards compatibility. Happy to send a PR.

znichollscr:
# Description

Fixes #648 

# Checklist

~* [ ] If you add new **CLI arguments** (or change the meaning of existing ones), make sure `docs/cli.rst` reflects those changes.~

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -80,7 +80,10 @@ Top level keys
 
     ``""`` by default.
 
-    Formatted titles are appended a line of ``=`` on the following line (reStructuredText title format) unless the template has an ``.md`` suffix, in which case the title will instead be prefixed with ``#`` (markdown title format).
+    When using reStructuredText, formatted titles are underlined using the ``underlines`` configuration.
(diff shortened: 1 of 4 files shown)


adiroiban (approved):
Thanks. Changes looks good.

I am looking at the Markdown tutorial documentation at https://towncrier.readthedocs.io/en/latest/markdown.html

and I can see that it has this config

```
title_format = "## [{version}](https://github.com/twisted/my-project/tree/{version}) - {project_date}"
```

so it looks like #610 is a bug.

-----

Thanks again.

speedenator:
fixed while I slept. Love it! Thanks @znichollscr and @adiroiban !

PR #664, opened 2024-08-23
Fix section depth in markdown template

### LINKED ISSUE #663: Markdown template skips h2 level when no section exists
MyST doesn’t like that and throws a warning. I’d like to stay warning-free without spuriously ignoring warnings.

I filed #664, but a better solution for this is (and would have been for #648) to add a “initial section depth” config option for markdown instead of allowing users to customize the title section depth while not adapting the other sections.

We can kinda hack this in with the current solution as well:

```py
if m := re.fullmatch(r"([^\n]*\n)*(?P<hashes>#+)\s.*", title_format, re.MULTILINE):
    initial_depth = len(m["hashes"])
else:
    initial_depth = 1
sec_depth = initial_depth + 1
cat_depths = defaultdict(lambda: sec_depth + 1)
cat_depths[""] = sec_depth
depth = dict(title=initial_depth, section=sec_depth, category=cat_depths)
```

and the template could read for e.g. the section header:

```jinja
{{ "#" * depth["section"] }} {{section}}
```

and for the category header:

```jinja
{{ "#" * depth["category"][section] }} {{ definitions[category]['name'] }}
```

<details>
<summary>alternative implementation</summary>

or less complex code

```py
[…]
cat_depth = sec_depth + 1
depth = dict(title=initial_depth, section=sec_depth, category=cat_depth)
```

and more complex template

```jinja
{{ "#" * depth["category" if section else "section"] }} {{ definitions[category]['name'] }}
```

</details>

flying-sheep:
# Description

Fixes #663

# Checklist

### DIFF
--- a/src/towncrier/newsfragments/663.bugfix.rst
+++ b/src/towncrier/newsfragments/663.bugfix.rst
@@ -1,0 +1,1 @@
+Markdown template: Render category headers as h2 when no sections are present.
--- a/src/towncrier/newsfragments/663.bugfix.rst
+++ b/src/towncrier/newsfragments/663.bugfix.rst
@@ -1,1 +1,0 @@
-Markdown template: Render category headers as h2 when no sections are present.
--- a/src/towncrier/newsfragments/663.removal
+++ b/src/towncrier/newsfragments/663.removal
@@ -1,0 +1,3 @@
+When no sections are present,
+the default Markdown template now renders the category headers as H2.
+In previous versions it was rendered as H3.
--- a/src/towncrier/templates/default.md
+++ b/src/towncrier/templates/default.md
@@ -13,7 +13,7 @@
 
 {% if sections[section] %}
 {% for category, val in definitions.items() if category in sections[section] %}
-### {{ definitions[category]['name'] }}
+##{% if section %}#{% endif %} {{ definitions[category]['name'] }}
 
 {% for text, values in sections[section][category].items() %}
 - {{ text }}
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -1085,7 +1085,7 @@ class TestCli(TestCase):
 
             [20-01-2001] CUSTOM RELEASE for FooBarBaz version 7.8.9
 
-            ### Features
+            ## Features
 
             - Adds levitation (#123)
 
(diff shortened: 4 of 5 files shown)


src/towncrier/newsfragments/663.bugfix.rst
```diff
@@ -0,0 +1 @@
+Markdown template: Render category headers as h2 when no sections are present.
```
adiroiban: Just a small comment... and mauybe this should be a `.removal` type of fragment.

It should explain the existing users should manually update existing NEWS file to match the new H2 format.


```suggestion
When no sections are present, the default Markdown template now renders the category headers as H2. In previous versions it was rendered as H3.
```

adiroiban (commented):
Thanks for the PR.

Changes looks good... but this might be considered a backward incompatible change.

Users that already have NEWS files, will have them generated with H3 for previous releases.

I think that this might be ok, but we should communicate this change to end users.

--------

I was expected to also see some updates to the test docstrings, to explain that the H2 render is important.

---------

I hope that someone who cares about MD support in towncrier will have some time and do a review of this PR

flying-sheep:
Hi @adiroiban, I commented a bit on the bigger problem here: https://github.com/twisted/towncrier/issues/663

maybe you’d like to chime in there before I invest more time into this fix

adiroiban (approved):
This looks good.

I think that `trunk` needs to be updated to fix the read the docs build.

flying-sheep:
Huh, never saw that one before!

```
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    pip from https://files.pythonhosted.org/packages/c9/bc/b7db44f5f39f9d0494071bddae6880eb645970366d0a200022a1a93d57f5/pip-25.0.1-py3-none-any.whl:
        Expected sha256 c46efd13b6aa8279f33f2864459c8ce587ea6a1a59ee20de055868d8f7688f7f
             Got        2f15a3708f631f768bfdc86dbcbfc1b5a31ce7097c5858b011f36a909bb3d4de
```

adiroiban:
Sorry for the delay... I was quite busy lately.

The code from here is ok... it might be just a Read The Docs issue... I triggered a re-run

Many thanks for your help!

PR #679, opened 2025-01-11
Use a pkginfo that supports 2.4 metadata

twm:
# Description

Require [pkginfo>=1.12.0](https://pypi.org/project/pkginfo/1.12.0/), which supports metadata 2.4. This allows Twine to work with packages generated by the latest hatchling. Refer to #678.

Also bump build-and-inspect-python-package to [v2.11.0](https://github.com/hynek/build-and-inspect-python-package/releases/tag/v2.11.0) for the same reason.

# Checklist


glyph (approved):
Yay?

hynek:
You have to keep in mind that while the wheel is correct, Twine 6.0.1 will not upload licensing metadata (c.f. attrs's issue tracker & 329509342875 PRs around this). Since Towncrier has a license trove classifier, that might be mostly OK, but the metadata as it stands will not be uploaded until Twine ships a release with https://github.com/pypa/twine/pull/1180

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling<1.27.0"]
+requires = ["hatchling"]
```
adiroiban, line 2: Maybe add this to the metadata.

If someone will try to run `python -m build` with an older pkginfo, I expect they will get an error.

```suggestion
requires = ["hatchling", "pkginfo>=1.12.0"]
```

adiroiban (approved):
It look good. Thanks.

I guess that we need to trigger a new release to actually test this.

pyproject.toml
twm: It actually builds fine without this &mdash; it is `twine` choking because of an old `pkginfo`.

twm:
> You have to keep in mind that while the wheel is correct, Twine 6.0.1 will not upload licensing metadata (c.f. attrs's issue tracker & 329509342875 PRs around this). Since Towncrier has a license trove classifier, that might be mostly OK, but the metadata as it stands will not be uploaded until Twine ships a release with [pypa/twine#1180](https://github.com/pypa/twine/pull/1180)

That's interesting, thanks for the explanation. I don't think that doing a release without this metadata is a blocker, though, since no prior release has had it either.

PR #682, opened 2025-01-22
Use build-and-inspect-python-package v2.12.0

twm:
# Description

For our [glorious metadata future](https://github.com/python-attrs/attrs/issues/1386#issuecomment-2606360508)!

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - uses: hynek/build-and-inspect-python-package@14c7e53f5d033cfa99f7af916fa59a6f7f356394  # v2.11.0
+    - uses: hynek/build-and-inspect-python-package@b5076c307dc91924a82ad150cdd1533b444d3310  # v2.12.0
 
     - name: Set up ${{ matrix.python.name }}
       uses: actions/setup-python@v4


glyph (approved):
I will … have to take your word for it :).

Maybe @hynek should push some tags, so we don't pin to hashes, though?

hynek:
there's plenty tags. some people like to to be extra-careful and pin actions to hashes.

twm:
:joy: I was just following the pattern, but FTR https://github.com/hynek/build-and-inspect-python-package/commit/b5076c307dc91924a82ad150cdd1533b444d3310

adiroiban:
> there's plenty tags. some people like to to be extra-careful and pin actions to hashes.

The pin hash might be my fault and my paranoia. 

With all kind of supply chain actions, I thought that is a bit better to depend on a fixed code.

With the next PR touching this code, we can start using auto-updating tags.

There are so many ways in which one can implement a supply chain attach, that maybe nobody will consider `hynek/build-and-inspect-python-package`

glyph:
> There are so many ways in which one can implement a supply chain attach, that maybe nobody will consider `hynek/build-and-inspect-python-package`

The issue is less that this particular package isn't a target, as that anyone who can take over Hynek's repo and push arbitrary tags to it, can probably also execute a git hash collision; so it's not really a security measure, it's just less readable :)

adiroiban:
>  can probably also execute a git hash collision;

Yes. Pinning is not bullet proof. My reasoning was that a hash collision will be more noticable, in comparison to updating a tag.

But we can start using tags. No problem.

PR #696, opened 2025-03-31
Fix Markdown headers

### LINKED ISSUE #695: Markdown renderer should infer header level from title format
If you e.g. set `title_format = "#### {name} {version}"`, section headers will have the same header level as the title.

This shouldn’t be the case. Instead sections and section-less categories should be one header level below the title, and categories within sections one more.

Since specifying arbitrarily deep header sections in the `title_format` option, this should of course work by default and not require a template.

Example output:

```markdown
<!-- level(header) is defined by `title_format` ⇒ here level 4 -->
#### Foo 1.2.3

<!-- section-less categories are at level(header) + 1 ⇒ here level 5 -->
##### Features
- Added abc

<!-- sections are at level(header) + 1 too ⇒ here level 5 -->
##### Web

<!-- categories within sections are at level(header) + 2 ⇒ here level 6 -->
###### Bugfixes
- Fixed xyz
```

flying-sheep:
# Description

Fixes #695

# Checklist

### DIFF
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -436,7 +436,9 @@ def render_fragments(
         top_underline=top_underline,
         get_indent=get_indent,  # simplify indentation in the jinja template.
         issues_by_category=issues_by_category,
-        **({} if md_header_level is None else dict(header_prefix="#" * md_header_level)),
+        **(
+            {} if md_header_level is None else dict(header_prefix="#" * md_header_level)
+        ),
     )
 
     for line in res.split("\n"):
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -362,7 +362,7 @@ def render_fragments(
     top_underline: str = "=",
     all_bullets: bool = False,
     render_title: bool = True,
-    md_header_level: int | None = 1,
+    md_header_level: int = 1,
 ) -> str:
     """
     Render the fragments into a news file.
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -362,6 +362,7 @@ def render_fragments(
     top_underline: str = "=",
     all_bullets: bool = False,
     render_title: bool = True,
+    md_header_level: int | None = 1,
 ) -> str:
     """
     Render the fragments into a news file.
@@ -435,6 +436,7 @@ def render_fragments(
         top_underline=top_underline,
         get_indent=get_indent,  # simplify indentation in the jinja template.
         issues_by_category=issues_by_category,
+        **({} if md_header_level is None else dict(header_prefix="#" * md_header_level)),
     )
 
     for line in res.split("\n"):
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -436,9 +436,7 @@ def render_fragments(
         top_underline=top_underline,
         get_indent=get_indent,  # simplify indentation in the jinja template.
         issues_by_category=issues_by_category,
-        **(
-            {} if md_header_level is None else dict(header_prefix="#" * md_header_level)
-        ),
+        header_prefix="#" * md_header_level,
     )
 
     for line in res.split("\n"):
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -215,7 +215,7 @@ def __main(
     render_title = config.title_format == ""
 
     # Add format-specific context to the template
-    md_header_level: int | None = None
+    md_header_level = 1
     if is_markdown:
         if config.title_format:
             m = re.search(r"^#+(?=\s)", config.title_format, re.MULTILINE)
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -8,6 +8,7 @@ Build a combined news file from news fragments.
 from __future__ import annotations
 
 import os
+import re
 import sys
 
 from datetime import date
@@ -213,6 +214,16 @@ def __main(
     # in the template.
     render_title = config.title_format == ""
 
+    # Add format-specific context to the template
+    md_header_level: int | None = None
+    if is_markdown:
+        if config.title_format:
+            m = re.search(r"^#+(?=\s)", config.title_format, re.MULTILINE)
+            lvl = len(m[0]) if m else 0
+        else:
+            lvl = 1 if render_title else 0  # TODO: derive from template or make configurable
+        md_header_level = lvl
+
     rendered = render_fragments(
         # The 0th underline is used for the top line
         template,
@@ -225,6 +236,7 @@ def __main(
         top_underline=config.underlines[0],
         all_bullets=config.all_bullets,
         render_title=render_title,
+        md_header_level=md_header_level,
     )
 
     if config.title_format:
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -221,7 +221,9 @@ def __main(
             m = re.search(r"^#+(?=\s)", config.title_format, re.MULTILINE)
             lvl = len(m[0]) if m else 0
         else:
-            lvl = 1 if render_title else 0  # TODO: derive from template or make configurable
+            lvl = (
+                1 if render_title else 0
+            )  # TODO: derive from template or make configurable
         md_header_level = lvl
 
     rendered = render_fragments(
--- a/src/towncrier/build.py
+++ b/src/towncrier/build.py
@@ -220,10 +220,8 @@ def __main(
         if config.title_format:
             m = re.search(r"^#+(?=\s)", config.title_format, re.MULTILINE)
             lvl = len(m[0]) if m else 0
-        else:
-            lvl = (
-                1 if render_title else 0
-            )  # TODO: derive from template or make configurable
+        else:  # TODO: derive from template or make configurable?
+            lvl = 1 if render_title else 0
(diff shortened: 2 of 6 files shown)


src/towncrier/build.py
```diff
@@ -213,6 +214,16 @@ def __main(
     # in the template.
     render_title = config.title_format == ""
 
+    # Add format-specific context to the template
+    md_header_level: int | None = None
+    if is_markdown:
+        if config.title_format:
+            m = re.search(r"^#+(?=\s)", config.title_format, re.MULTILINE)
+            lvl = len(m[0]) if m else 0
+        else:  # TODO: derive from template or make configurable?
```
flying-sheep, line 223: How do you think should we fix this TODO?

src/towncrier/test/test_build.py
```diff
@@ -1086,7 +1086,7 @@ def test_title_format_custom_markdown(self, runner):
 
             [20-01-2001] CUSTOM RELEASE for FooBarBaz version 7.8.9
 
-            ## Features
+            # Features
```
adiroiban, line 1089: This looks like a backward incompatible change...

I don't know what to say...

I think that one issue here is the fact that I don't use Markdown... and I don't really care about markdown support in towncier.

So we end up with various inconsistent changes... as I am ok with whatever is added in the lasest PR

Can we implement this without breaking the existing functionality ?

I am just asking.
I don't know how this can be done.

-----------

My hope is that we can leave the header prefix entirely to the template rendering.

---------

So if you need a different heading level you can create a custom template.
 
---------

Can you please provide a bit of context to this requirement ?

Why the current functionality doesn't work for you ?

src/towncrier/build.py
adiroiban: For RST this is not possible.

This is why we have the `underlines` configuration.

--------

My hope is we can remove the heading markup from the Python code and let this up to the Jinja2 template rendering.

Users can then implement whatever level they want.

src/towncrier/test/test_build.py
flying-sheep: Markdown files need to start with header level 1 (`# Some header`) and increase the header level by increments of 1, no more and no less.

> This looks like a backward incompatible change...

I guess you could call “the behavior is no longer broken” that, but I prefer to call it a bugfix.

> So if you need a different heading level you can create a custom template.

Not as long as `title_format` for markdown allows specifying a deeper heading level. As long as that’s the case, everything should just be compatible with it.

I’m happy to change the approach of the implementation (if you can clearly tell me what should be changed and why), but the tests make sense *after* my changes, not before, so the behavior is currently correct and should not be changed. (as said, as long as custom markup in `title_format` is allowed.)

> Can you please provide a bit of context to this requirement ?

I expanded the explanation in #695 a bit.

src/towncrier/build.py
flying-sheep: > My hope is we can remove the heading markup from the Python code and let this up to the Jinja2 template rendering.

that would be ideal, but would need some breaking changes. E.g. we currently do

```toml
title_format = "(v{version})=\n### {version} {{small}}`{project_date}`"
```

where the `(v{version})=` is a label for intersphinx. This wouldn’t be possible without the freedom we currently have (unless an alternative would be implemented)

adiroiban: Answering the initial question.

We don't have to use the `header_prefix` value for RST template.

We can leave it empty.

-----

I was thinking that we can leave `title_format` empty and make sure that we expose `version` , `small` and `project_data` to the Jinja2 template.

flying-sheep: That doesn’t sound like what this PR does, but maybe I’m misunderstanding.

I think if you want to go in a completely different direction, maybe you should make an alternative PR instead. That would be faster than explaining to me what you want and changing this PR to look like that.

flying-sheep: > We don't have to use the header_prefix value for RST template.
> We can leave it empty.

Check again, this is already the case. The `header_prefix` variable is only passed to Markdown templates, not rST. Or did you mean something else?

adiroiban:
Thanks for the PR and sorry for the noise.
I will leave this to be reviewed by someone familiar with Markdown.
Thanks again

src/towncrier/test/test_build.py
adiroiban: Sorry for the trouble. My bad.
Many thanks for the info and the fix
I re-checked the test and you are right.

src/towncrier/build.py
adiroiban: True...My bad. I am sorry for all the confusing here.

------

I think that we should have a constant value here, that should make the rendered MD file valid wnen using the default template

If users are using a custom template, they will have to also set a custom `title_format`

This can be documenented ?

Make sense?

-------

Also, just asking. Is there an automated test for this use case?

I see that the current tests pass

src/towncrier/_builder.py
```diff
@@ -435,6 +436,9 @@ def get_indent(text: str) -> str:
         top_underline=top_underline,
         get_indent=get_indent,  # simplify indentation in the jinja template.
         issues_by_category=issues_by_category,
+        **(
```
adiroiban: Just asking ?

Would it be possible to always have `header_prefix="#" * md_header_level`

If there is no extra level, we will have just empty string.

I hope that this will make it easier to write custom templates as you don't have to check for missing  `header_prefix` in Jinja template data

src/towncrier/_builder.py
```diff
@@ -362,6 +362,7 @@ def render_fragments(
     top_underline: str = "=",
     all_bullets: bool = False,
     render_title: bool = True,
+    md_header_level: int | None = 1,
```
adiroiban: I hope that we can make this only `int` without having to handle the `None` case

src/towncrier/test/test_build.py
SmileyChris: @flying-sheep i'm not sure I like using header level 1 -- it makes the assumption is that there isn't an existing heading level 1 in content above towncrier.

```markdown
# Change Log
This is the changelog of my project. You can find the
issue tracker at http://blah.

<!-- towncrier release notes start -->
```

src/towncrier/build.py
```diff
@@ -213,6 +214,16 @@ def __main(
     # in the template.
     render_title = config.title_format == ""
 
+    # Add format-specific context to the template
+    md_header_level: int | None = None
```
adiroiban: Maybe start with a default value that will work with the default template

```suggestion

    # This is the default value that should match the desing
    # of the current default template for Markdown.
    md_header_level = 1
```

src/towncrier/test/test_build.py
SmileyChris: > Markdown files need to start with header level 1 (# Some header) and increase the header level by increments of 1, no more and no less.

AFAIK, ​Markdown does not enforce a strict requirement that files must start with a level 1 header (# Header) or that header levels must increase sequentially by increments of one.  Can you provide a link to the spec for this?

flying-sheep: > i'm not sure I like using header level 1 -- it makes the assumption is that there isn't an existing heading level 1 in content above towncrier.

Good thing that that’s exactly what my PR fixes!

Again:
- Before my PR, sections always have header level 3, even if you set `title_format = "### ..."`
- After my PR, section headers are correct, no matter what header level `title_format` has.

If you have a concrete, actionable idea how you would like to handle the initial header level when `title_format = ""` or `title_format = None`, I’m up for it, see the `TODO: derive from template or make configurable?`

I just didn‘t want to bikeshed over how that config setting is supposed to be named, so I left it as a TODO. TBH, I think it should be handled in a separate PR.

> AFAIK, ​Markdown does not enforce a strict requirement that files must start with a level 1 header (# Header) or that header levels must increase sequentially by increments of one. Can you provide a link to the spec for this?

Red herring, ignore it. See above for the real problem this fixes.

src/towncrier/_builder.py
flying-sheep: True, jinja doesn’t care if the template doesn’t use some variables in the context.

src/towncrier/build.py
flying-sheep: > I see that the current tests pass

yeah, I repurposed a rather complicated test that required a custom template to just use the now-functional default template

src/towncrier/test/test_build.py
SmileyChris: Thanks, dived in to the middle of a discussion and made a bad assumption. This seems good to me!

SmileyChris (approved):

adiroiban:
> yeah, I repurposed a rather complicated test that required a custom template to just use the now-functional default template

A dedicated test would be ideal... but as long as we have test coverage, we should be good.

I don't use `towncrier` with markdown, so I don't have any strong arguments.

As long as Chris is happy and has approved the PR, I think that we can merge.

----

I have triggered a review from Copilot...just to doublecheck that all is ok.

The results was :)

> Consider adding an explicit test case that verifies custom markdown header levels are correctly applied when a non-default title format is used.

-------


Many thanks for the PR and sorry for the delay

I have merged this.

If you want to improve the tests, feel free to create a new PR.

I will try to get a new release soon...

PR #697, opened 2025-03-31
Custom allbullet setting

SmileyChris:
It provides a new `all_bullets` setting per category type that can override the global default.
This is useful if you want to have a text-only section (for example, an text introduction to the release)

### Implementation Details

* Updated fragment types loading system to support new features
* Modified templates to handle new `all_bullets` configuration
* Added tests for new functionality
* Updated documentation to explain new features


adiroiban:
Thanks for the PR.

This is something that I was also looking to have in `towncrier`.

I was thinking to have someting like `123.highlight.rst` 

The `higlight` is a special type that is rendered first and without a heading.

`description` make me think like this is the only / single description for a release.

With `highlight` you can have multiple highlights for the release.
The highlights can still be associated with an issue.

This can be currently implemented with a custom template... but maybe we can have a good default

SmileyChris:
Sure, `highlight` might be a better name.

This PR actually makes the existing default template respect the exsiting all_bullets setting as well, so that's a plus :)
Another small feature it adds is the ability to extend the built-in types with a custom one without having to write them all out (as shown in the docs example).

SmileyChris:
@adiroiban so at the moment, this just provides individuals to add this highlight section.

Should we just add it as a default since it's kind-of a useful thing and having a standard for it is better than people having to do it themselves?

adiroiban:
> Should we just add it as a default since it's kind-of a useful thing and having a standard for it is better than people having to do it themselves?

I think that it makes sense to have this available as default and with associated documentation.

adiroiban:
@SmileyChris do you think that you will have time to work on this in the next few days?

I am planning to do a new release... and just wanted to see if I should wait for this PR.

No hurry.

Cheers

PR #699, opened 2025-04-14
Implement Mercurial support

### LINKED ISSUE #394: Support for other VCS (e.g. Mercurial)
I've forked `towncrier` and modified it to handle hg repositories, seemingly well enough. Is there interest in supporting repository types besides git? If so, I'll work on cleaning up my changes and opening a PR.

cdevienne:
# Description

Fixes #394

Add support for mercurial repositories.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -164,6 +164,8 @@ skip_covered = true
 exclude_lines = [
     "pragma: no cover",
     "if TYPE_CHECKING:",
+    # Empty functions of a Protocol definition (used in _vcs.py) can never be
+    # executed. Ignoring them.
     ": \\.\\.\\.$",
 ]
 omit = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -164,6 +164,7 @@ skip_covered = true
 exclude_lines = [
     "pragma: no cover",
     "if TYPE_CHECKING:",
+    ": \\.\\.\\.$",
 ]
 omit = [
     "src/towncrier/__main__.py",
--- a/src/towncrier/_hg.py
+++ b/src/towncrier/_hg.py
@@ -1,0 +1,42 @@
+# Copyright (c) Amber Brown, 2015
+# See LICENSE for details.
+
+from __future__ import annotations
+
+import os
+
+from subprocess import STDOUT, CalledProcessError, call, check_output
+
+
+def remove_files(fragment_filenames: list[str]) -> None:
+    if not fragment_filenames:
+        return
+
+    # Filter out files that are unknown to git
+    try:
+        hg_fragments = check_output(
+            ["hg", "files"] + fragment_filenames, encoding="utf-8"
+        ).split("\n")
+    except CalledProcessError:
+        # we may not be in a git repository
+        hg_fragments = []
+
+    hg_fragments = [os.path.abspath(f) for f in hg_fragments if os.path.isfile(f)]
+    call(["hg", "rm", "--force"] + hg_fragments)
+    unknown_fragments = set(fragment_filenames) - set(hg_fragments)
+    for unknown_fragment in unknown_fragments:
+        os.remove(unknown_fragment)
+
+
+def stage_newsfile(directory: str, filename: str) -> None:
+    call(["hg", "add", os.path.join(directory, filename)])
+
+
+def get_remote_branches(base_directory: str) -> list[str]:
+    return []
+
+
+def list_changed_files_compared_to_branch(
+    base_directory: str, compare_with: str, include_staged: bool
+) -> list[str]:
+    return []
--- a/src/towncrier/_hg.py
+++ b/src/towncrier/_hg.py
@@ -6,6 +6,41 @@ from __future__ import annotations
 import os
 
 from subprocess import STDOUT, CalledProcessError, call, check_output
+from typing import Container
+
+
+def get_default_compare_branch(branches: Container[str]) -> str | None:
+    if "default" in branches:
+        return "default"
+    return None
+
+
+def _topic_enabled(directory: str):
+    for e in (
+        check_output(
+            ["hg", "config"],
+            cwd=directory,
+            encoding="utf-8",
+        )
+        .strip()
+        .splitlines()
+    ):
+        if (
+            e.startswith("extensions.topic")
+            and e.split("=")[0].strip() == "extensions.topic"
+        ):
+            return True
+    return False
+
+
+_has_topics_cache = {}
+
+
+def has_topics(directory: str) -> bool:
+    global _has_topics_cache
+    if directory not in _has_topics_cache:
+        _has_topics_cache[directory] = _topic_enabled(directory)
+    return _has_topics_cache[directory]
 
 
 def remove_files(fragment_filenames: list[str]) -> None:
@@ -33,10 +68,32 @@ def stage_newsfile(directory: str, filename: str) -> None:
 
 
 def get_remote_branches(base_directory: str) -> list[str]:
-    return []
+    branches = check_output(
+        ["hg", "branch"],
+        cwd=base_directory,
+        encoding="utf-8",
+        stderr=STDOUT,
+    ).splitlines()
+
+    if has_topics(base_directory):
+        branches += check_output(
+            ["hg", "topic", "--template", "{topic}"],
+            cwd=base_directory,
+            encoding="utf-8",
+            stderr=STDOUT,
+        ).splitlines()
+
+    return branches
 
 
 def list_changed_files_compared_to_branch(
     base_directory: str, compare_with: str, include_staged: bool
 ) -> list[str]:
-    return []
+    output = check_output(
+        ["hg", "diff", "--stat", "-r", compare_with],
+        cwd=base_directory,
+        encoding="utf-8",
+        stderr=STDOUT,
+    ).splitlines()
+
+    return [l.split("|")[0].strip() for l in output if "|" in l]
--- a/src/towncrier/_hg.py
+++ b/src/towncrier/_hg.py
@@ -1,4 +1,4 @@
-# Copyright (c) Amber Brown, 2015
+# Copyright (c) towncrier contributors, 2025
 # See LICENSE for details.
 
 from __future__ import annotations
--- a/src/towncrier/_hg.py
+++ b/src/towncrier/_hg.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import os
 
-from subprocess import STDOUT, CalledProcessError, call, check_output
+from subprocess import STDOUT, call, check_output
 from typing import Container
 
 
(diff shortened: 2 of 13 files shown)


adiroiban:
Thanks, Christophe for the PR.

I would love to see support for HG in `towncrier`

In order to merge this PR, it's critical to have all the tests succesfully executed and all new code to be covered by tests.

We still support Python 3.8 and Python 3.9

We can remove support for 3.8 ... but in a separate PR.

And if supporting Python 3.9 is too much effort, we can also consider removing support for Python 3.9

mypy checkes are part of our standard tests

```
src/towncrier/_hg.py:18: error: Function is missing a return type annotation  [no-untyped-def]
src/towncrier/_hg.py:42: error: Returning Any from function declared to return "bool"  [no-any-return]
src/towncrier/_vcs.py:6: error: Function is missing a return type annotation  [no-untyped-def]
src/towncrier/_vcs.py:26: error: Returning Any from function declared to return "str | None"  [no-any-return]
src/towncrier/_vcs.py:30: error: Returning Any from function declared to return "None"  [no-any-return]
src/towncrier/_vcs.py:34: error: Returning Any from function declared to return "None"  [no-any-return]
src/towncrier/_vcs.py:38: error: Returning Any from function declared to return "list[str]"  [no-any-return]
src/towncrier/_vcs.py:44: error: Returning Any from function declared to return "list[str]"  [no-any-return]
Found 8 errors in 2 files (checked [19](https://github.com/twisted/towncrier/actions/runs/14451901365/job/40526351238?pr=699#step:7:20) source files)
```

cdevienne:
yep, I am working on getting the tests to pass. The type checking part is a little more problematic for me, but I will find a way.

adiroiban:
Thanks for following up with the tests

Just a reminder... if supporting Python 3.8 is too hard, we can remove support for it.

> The type checking part is a little more problematic for me, but I will find a way.

Don't worry too much. Just ask for help and we can see how we can fix the mypy part.

Thanks again!

cdevienne:
I think the tests and typing are good now, but the tests on Windows fail for an unknown reason.
My guess is that I broke something in the _git module, but I have no idea what.

Since I have no easy access to a windows computer, I would be glad if someone could have a look. Thanks!

src/towncrier/_vcs.py
```diff
@@ -0,0 +1,63 @@
...
+
+        hg: VCSMod = _hg
+
+        return hg
+    elif base_directory == "/":
+        from . import _novcs
+
+        return _novcs
+    else:
+        return _get_mod(os.path.dirname(base_directory))
```
adiroiban: I think that it helps to add a comment for why we are doing this.

... but, do we want to have this functionality ?

```suggestion
        # No VCS was found in current directory.
        # We will try our luck in the parent directory.
        return _get_mod(os.path.dirname(base_directory))
```
it looks like there is a recursion error here

I guess that on WIndows, you never have the Unix root `/` base directory...so we have an endless loop here 

```
  File "D:\a\towncrier\towncrier\.nox\tests-3-9\lib\site-packages\towncrier\_vcs.py", line 35, in _get_mod
    return _get_mod(os.path.dirname(base_directory))
  File "D:\a\towncrier\towncrier\.nox\tests-3-9\lib\site-packages\towncrier\_vcs.py", line 35, in _get_mod
    return _get_mod(os.path.dirname(base_directory))
  File "D:\a\towncrier\towncrier\.nox\tests-3-9\lib\site-packages\towncrier\_vcs.py", line 35, in _get_mod
    return _get_mod(os.path.dirname(base_directory))
  File "D:\a\towncrier\towncrier\.nox\tests-3-9\lib\site-packages\towncrier\_vcs.py", line 35, in _get_mod
    return _get_mod(os.path.dirname(base_directory))
  File "D:\a\towncrier\towncrier\.nox\tests-3-9\lib\site-packages\towncrier\_vcs.py", line 35, in _get_mod
    return _get_mod(os.path.dirname(base_directory))
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\ntpath.py", line 223, in dirname
    return split(p)[0]
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\ntpath.py", line 186, in split
    seps = _get_bothseps(p)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\ntpath.py", line 35, in _get_bothseps
    if isinstance(path, bytes):
builtins.RecursionError: maximum recursion depth exceeded while calling a Python object
```

src/towncrier/newsfragments/394.feature.rst
```diff
@@ -0,0 +1 @@
+Implement Mercurial support
```
adiroiban: Just a small suggestion to have this read as a release note text and not as a PR description.

```suggestion
Support was added for Mercurial SCM.
```

pyproject.toml
```diff
@@ -164,6 +164,7 @@ skip_covered = true
 exclude_lines = [
     "pragma: no cover",
     "if TYPE_CHECKING:",
+    ": \\.\\.\\.$",
```
adiroiban, line 169: just asking. why are we ignoring  these lines?
maybe add a coment to explain why.

src/towncrier/_hg.py
```diff
@@ -0,0 +1,63 @@
+# Copyright (c) Amber Brown, 2015
```
adiroiban:
```suggestion
# Copyright (c) towncrier contributors, 2025
```

src/towncrier/_vcs.py
adiroiban: just some help... this is the output from windows...

I guess that if dirname is the same as current path, we can expect that we are at the root


```
Python 3.12.7 (tags/v3.12.7:0b05ead, Oct  1 2024, 03:06:41) [MSC v.1941 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.dirname('c:\\some-path')
'c:\\'
>>> os.path.dirname('c:\\')
'c:\\'
>>>
```

src/towncrier/test/test_hg.py
```diff
@@ -0,0 +1,79 @@
...
+        assert _hg.get_default_compare_branch(["main", "default"]) == "default"
+        assert _hg.get_default_compare_branch(["main", "a_topic"]) is None
+
+    def test_empty_remove(self):
+        """
+        If remove_files gets an empty list, it returns gracefully.
+        """
+        _hg.remove_files([])
+
+    def test_hg(self):
```
adiroiban: can we have a docstring for these tests describing what is the scope/goal of each test?

What is the different between `test_hg` and `test_get_default_compare_branch` ?

We don't have any other tools to validated the code inside the tests.
So a good description can help a reviewer validated whether the code inside the test function is valid or not

adiroiban:
Thanks for the PR... I only did a quick review.
The code looks good.

I hope you can fix the Windows issue and then we can get a report for the code coverage.

Having good test coverage for the Mergurial code is the only way to guarantee that we will not break hg support in future releases

Thanks again! Great job here!

adiroiban:
You can get access to GitHub VMs using [tmate](https://github.com/mxschmitt/action-tmate)
This should work even on Windows

cdevienne:
I made a good guess and voilà, the tests pass on all platforms!

I will follow your review recommendations shortly, and it should be good.

src/towncrier/_vcs.py
cdevienne: Just found out your comment... would have save me some time if I did earlier, but I came to the same conclusion and fixed it.
I added the comment you suggested.
About the need for the recursion, without it the tests fail because towncrier is supposed to be able to work when invoked in a sub-directory (at least it is what I came to understand).

src/towncrier/_hg.py
cdevienne: done

pyproject.toml
cdevienne: This is to ignore the empty method definitions of the Protocol class in _vcs.py, as they can never be executed.

cdevienne: I added a comment

src/towncrier/test/test_hg.py
cdevienne: Added comments

cdevienne:
I added comments to match your review recommendations (thanks!).

Note that the mercurial tests only test the _hg module itself. No full stack scenario run on a mercurial repository. So the tests are good as long as no change is done to the _vcs API.

One thing I will do is adding a test for _vcs that agnostically tests all the possible vcs, which will ensure all the provided vcs are up-to-date.

Then we should be good for merging.

src/towncrier/test/test_novcs.py
```diff
@@ -0,0 +1,22 @@
+from twisted.trial.unittest import TestCase
+
+from towncrier import _novcs
+
+
+class TestNOVCS(TestCase):
+    def test_get_default_compare_branch(self):
```
adiroiban, line 7:
```suggestion
    def test_get_default_compare_branch(self):
        """
        Witout a version control system, there is no default branch.
        """
```

src/towncrier/test/test_novcs.py
```diff
@@ -0,0 +1,22 @@
...
+        self.assertEqual(None, _novcs.get_default_compare_branch([]))
+
+    def test_empty_remove(self):
+        """
+        If remove_files gets an empty list, it returns gracefully.
+        """
+        _novcs.remove_files([])
+
+    def test_get_remote_branches(self):
```
adiroiban, line 19:
```suggestion
    def test_get_remote_branches(self):
       """
       There are no remote branches, when we don't have a VCS.
       """
```

src/towncrier/test/test_novcs.py
```diff
@@ -0,0 +1,22 @@
...
+        self.assertEqual([], _novcs.get_remote_branches("."))
+
+    def test_list_changed_files_compared_to_branch(self):
```
adiroiban, line 25:
```suggestion
    def test_list_changed_files_compared_to_branch(self):
       """
       No changed files are detected without a VCS.
       """
```

src/towncrier/_vcs.py
```diff
@@ -0,0 +1,72 @@
...
+        if parent == base_directory:
+            # We reached the fs root, abandoning
+            from . import _novcs
+
+            return _novcs
+
+        return _get_mod(parent)
+
+
+def get_default_compare_branch(
```
adiroiban, line 47: I like the way git and Mercurial was implemeted and I think that for this PR the current changes are ok.
No special need to extra work here.

---------

I think that for a future PR, it would be nice to have HG and GIT as a class that implement the `VCS` protocol, so that we don't have to create this type of code

adiroiban:
Just asking


Do we want to support using towncrier without a VCS ?
Does is make sense to have towncrier without a VCS?
Is this a real use case?

------

It would be nice to have at least a few end to end tests for `hg`.

The `towncrier` project doesn't have a big team of developers so most probably when changes are made in a PR, people will not manually run `hg` tests to check for regressions.

As an initial support I think that the test coverage in this PR is good enough ... is not perfect... but `towncrier` is a dev tool, so we should be fine.

If defects are found in `hg` people can follow up with PR to improve the support for `hg`

-------

I added a review from Copilot, just as a double check...and to test how good AI can be.

thanks again for the updates

cdevienne:
In real life the 'without a VCS' case is more likely to be a 'with unsupported VCS', or testing towncrier in a non-versioned directory before adopting it. In the later case I find it more friendly not to throw errors at a potential adopter's face.

I added the test I mentioned. On my side it's all good. Thanks for the time spent on my contribution!

cdevienne:
Oh and of course don't hesitate to ping me if some mercurial related issues come up.

src/towncrier/_vcs.py
cdevienne: I agree, I just wanted to limit the refactoring for a first contribution.

adiroiban:
Thanks for the update.
Let's get this merged and hopefully we will also have a new release soon.

People can then use towncrier with `hg` and if needed we can receive improvements on the current code

I think that the current code is a solid foundation.

PR #700, opened 2025-04-17
Update python versions

hynek:
We should drop EOL 3.8 either way.

I wonder if we really need importlib_resources on 3.9, but that's for another PR.


adiroiban (approved):
Looks good. Thanks for the cleanup.

As `importlib_resources` was only used for python 3.8, I would say that removing the `importlib_resources` dependency is also in the scope of this PR.

But happy to have the removal in a separate PR.

Thanks again!

hynek:
The confusing part is that the import guards are referring to 3.10. But let's try to remove it, sadly the new Mercurial tests fail on my computer, so I'll have to abuse to CI for this.

adiroiban:
I see tests failing... let's keep `importlib_resources` for another PR

-----


> sadly the new Mercurial tests fail on my computer, 

That's my bad.

@cdevienne  do you have time to submit a PR for which the `hg` tests are skipped if `hg` is not installed locally ? Thanks

We should be ok to skip the tests... on GitHub Actions, they should alwasy run... and if they are somehow accidentally skipped, we will get this reported via the code coverage

src/towncrier/test/helpers.py
```diff
@@ -1,22 +1,16 @@
...
 from functools import wraps
+from importlib import resources
 from pathlib import Path
 from subprocess import call
 from typing import Any, Callable
 
 from click.testing import CliRunner
 
 
-if sys.version_info < (3, 9):
```
adiroiban, line 14: I was looking only into this import ... but i see that we also have imports for `if sys.version_info < (3, 10):`

So importlib_resources can be removed only after we remove support for Python 3.9

hynek:
Yeah it looks like some behavior in the spec handling within import lib has changed, and while I've wasted a bunch of time to debug it, I don't think it's worth working around.

---

As for hg: I have Mercurial installed so that's not the issue. And even if not, we could add it to Nox to install it?

cdevienne:
> As for hg: I have Mercurial installed so that's not the issue. And even if not, we could add it to Nox to install it?

Could you give me details on what is failing?

cdevienne:
> @cdevienne do you have time to submit a PR for which the `hg` tests are skipped if `hg` is not installed locally ? Thanks

Sure: https://github.com/twisted/towncrier/pull/703

PR #701, opened 2025-04-17
Fix hg tests to work on unconfigured systems

hynek:
Otherwise they fail and debugging revealed that a user was missing (I don't have an `.hgrc` anymore).

### DIFF
--- a/src/towncrier/test/test_hg.py
+++ b/src/towncrier/test/test_hg.py
@@ -40,7 +40,7 @@ def commit(message):
     There must be uncommitted changes
     """
     check_call(["hg", "addremove", "."])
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
Thanks! All good.

hynek:
I've also added what I've alluded to in the other issue: https://github.com/twisted/towncrier/pull/701/commits/1a16acd01e126f0dad2c0be4019e04648d315f27 or does that break some coverage stuff?

adiroiban:
I think that GitHub Action VMs already have hg installed ... 
I prefer not to force `hg` install on each developers computer ... and just skip the tests... see #702

hynek:
OK, gone 🪄

PR #713, opened 2025-06-03
Fix version trove classifiers for #700

hynek:
looks like the idiot responsible for #700 forgot to adjust trove classifiers, too. we'll never know who that was.


twm (approved):
:see_no_evil:

Issue #714, opened 2025-06-04
Overwrite files


tirolerstefan:
Currently, a postfix "{retry}" is appended to the fragment when the file is already available, e.g. 
`ABC-1234.feat.md` is available, so a file `ABC-1234.feat.1.md` is used.

I would like to overwrite existing files.

Our use case: if there is already a snippet available for this issue / type, then we want to edit it. 
Afterwards, we invoke `towncrier create` with the updated text and simple would like to overwrite it.

adiroiban:
Thanks for the report.
As long as backward compatiblity is preserved, we can add extra options.

tirolerstefan:
Hi  @adiroiban , thanks a lot for the feedback. I tested it and added auto tests to the PR.
But it's bloody complicated to get the pre-commit checks done (black / isort).


```
$ pre-commit run --all-files 
black....................................................................Passed
pyupgrade................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
debug statements (python)................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
towncrier-check..........................................................Passed
```

```
$ pre-commit autoupdate
[https://github.com/psf/black-pre-commit-mirror] updating 25.12.0 -> 26.5.1
[https://github.com/asottile/pyupgrade] already up to date!
[https://github.com/PyCQA/isort] updating 7.0.0 -> 9.0.0a3
[https://github.com/PyCQA/flake8] already up to date!
[https://github.com/pre-commit/pre-commit-hooks] already up to date!
[https://github.com/twisted/towncrier] already up to date!
```

```
$ pre-commit run --all-files 
black....................................................................Failed
- hook id: black
- files were modified by this hook

Warning: Python 3.12 cannot parse code formatted for Python 3.15. To fix this: run Black with Python 3.15, set --target-version to py312, or use --fast to skip the safety check. Black's safety check verifies equivalence by parsing the AST, which fails when the running Python is older than the target version.
reformatted src/towncrier/__main__.py
reformatted bin/towncrier
reformatted src/towncrier/_settings/__init__.py
reformatted admin/check_tag_version_match.py
reformatted docs/conf.py
reformatted src/towncrier/_writer.py
reformatted noxfile.py
reformatted src/towncrier/test/test_hg.py
reformatted src/towncrier/click_default_group.py
reformatted src/towncrier/test/test_builder.py
[...]

PR #715, opened 2025-06-04
create: new option 'overwrite' 

tirolerstefan:
# Description

added new option 'overwrite' which prevents creating additional snippets for already existing fragments

Issue #714

### DIFF
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -38,6 +38,11 @@ DEFAULT_CONTENT = "Add your info here"
     help=config_option_help,
 )
 @click.option(
+    "--overwrite",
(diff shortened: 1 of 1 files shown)


adiroiban (changes requested):
Thanks for the PR.

I think that this is a good start.

In order to merge the PR we need automate tests for the new functionality.

Also, all GitHub Action jobs should be successful. 

And I guess that it would also be nice to update the documentation to explain the purpose of this functionality

We have some docs here

https://towncrier--715.org.readthedocs.build/en/715/cli.html#towncrier-create

tirolerstefan:
closed as replaced by PR #735

PR #729, opened 2025-11-04
Update build-and-inspect-python-package

hynek:
GHA eagerly switched '3.x' to Python 3.14 and there's some incompatibilities.


glyph (approved):
LGTM

PR #734, opened 2026-05-22
Enable coverage reporting for test code

### LINKED ISSUE #711: Enable coverage report for the test code
Right now, we only report the coverage for the "production" code

This should be an important indicator in the case in which we have some tests that are always skipped.

2253781369:
## Summary

Remove test files from the coverage omit list so that test code is included in coverage tracking. This helps detect tests that are always skipped — they would show 0% coverage.

## Changes

- **pyproject.toml**: Remove `src/towncrier/test/*` from coverage omit list
- **noxfile.py**: Update coverage_report session to show separate production and test coverage reports
- **.github/workflows/ci.yml**: CI now reports full coverage (including tests) to the step summary, while still enforcing 100% coverage for production code only
- **src/towncrier/newsfragments/711.misc**: Required changelog fragment

Fixes #711

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,16 +288,16 @@ jobs:
         merge-multiple: true
         path: .
 
-    - name: Combine coverage & fail if it's <100%.
+    - name: Combine coverage & check thresholds.
       run: |
         python -Im coverage combine
         python -Im coverage html --skip-covered --skip-empty
 
-        # Report and write to summary.
+        # Report full coverage (including test code) to summary.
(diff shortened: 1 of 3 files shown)


adiroiban:
Is this AI generated without any review from a person?

If this was previously reviewed by a person, please double-check the diff.
it looks like this introduces newline changes.

I am closing this for now since I suspect that this is AI spam, but we can reopen it.

glyph:
I've blocked this user. In the future I think we need to be aggressive about such bans because otherwise we leave a vector for abuse open; they can request reconsideration outside of Github if they feel it's unfair.

PR #735, opened 2026-05-27
#714 feature(create): added option 'index' which allows direct access to indexed snippets

### LINKED ISSUE #714: Overwrite files

tirolerstefan:
Fixes https://github.com/twisted/towncrier/issues/714

Towncried currently adds an index if a snippet already exists, e.g. `issue-1234.feat.md` is available, so `issue-1234.feat.1.md` is created.

We have a graphical tool for changelog snippet creation and wanted to directly access the indexed snippets - so an overwrite is possible.

With this change, a new option "--index" is added.

e.g.
```
towncrier create --dir .../changelogs/preliminary --config .../config/changelogs/towncrier.toml --section BLA --content "test3<br/>
    another snippet" --index 8 ISSUE-1000.feat.md
Created news fragment at .../changelogs/preliminary/BLA/ISSUE-1000.feat.8.md
```

If index == 0, the filename without index is accessed.
If index > 0, the index is added to the filename.

If option "index" is not passed, the tool behaves as before - the next index is chosen.


adiroiban:
Hi Stefan.

Thanks for the PR.

It might be the case that our pre-commit is not yet ready for Python 3.15

I will take a look

tirolerstefan:
Thanks a lot for checking this. 
Note: Currently I'm using Python 3.12.3 on Ubuntu 24.04.

adiroiban:
Can you push your changes using `git --no-verify` ... I will check the pre-commit errors.
They might not be related to this PR

tirolerstefan:
My changes are pushed and up2date. All I wanted to commit are the fixes for
the hooks. Thanks!

src/towncrier/newsfragments/714.feature.rst
```diff
@@ -0,0 +1 @@
+A new option "--index" is added.
```
adiroiban: Maybe add a few more options so that end users will know where it is added and what it is used for.

```suggestion
The `towncrier create` command line tool now has the `--index` option that is used when creating multiple fragments for the same issue number.
```

src/towncrier/click_default_group.py
```diff
@@ -52,6 +52,7 @@ def bar():
    bar
 
 """
+
```
adiroiban, line 55: is this change needed?

src/towncrier/create.py
```diff
@@ -54,6 +54,14 @@
     type=str,
     help="The section to create the fragment for.",
 )
+@click.option(
+    "--index",
+    type=click.IntRange(min=0),
+    metavar="x",
```
adiroiban: Mabye N to signal this this is an umber.
```suggestion
    metavar="N",
```

src/towncrier/create.py
```diff
@@ -54,6 +54,14 @@
...
+    default=None,
+    help="Optional numeric index of the fragment "
+    "(e.g. x=1 -> 'issue.feat.1.md' is accessed and overwritten)",
```
adiroiban: Maybe the help command can be brief and do not include this info.

We have the separate  documentation https://github.com/twisted/towncrier/blob/trunk/docs/cli.rst where we have all the space to provide use case examples

src/towncrier/create.py
```diff
@@ -54,6 +54,14 @@
     type=str,
     help="The section to create the fragment for.",
 )
+@click.option(
+    "--index",
```
adiroiban: Index is kind of generic...
Maybe it's best to be more explicit to prevent future conflicts and confusions/
```suggestion
    "--issue-index",
```


I don't know if "index" is the best name for this feature... maybe we can call it `--sub-issue` or someting like that

adiroiban (changes requested):
Thanks for the PR and sorry for the late review.

I left a few inline comments.

To merge, this PR needs updated autoamted tests.

Thanks again

tirolerstefan:
hi @adiroiban - thanks, i have added my comments above and pushed a new commit.
waiting for the final naming of the option, then i will adapt this, too.

src/towncrier/click_default_group.py
tirolerstefan: this was one of the "auto fixes" of pre-commit.
no, it's not needed in my eyes, but i have to commit it using "--no-verify".
if you don't mind, i will leave it as it is to satisfy the pre-commit tools.

src/towncrier/create.py
tirolerstefan: hmmm, it's like an index - 0 is the first fragment, 1 the next, etc. but i can live with any text.
"sub-issue" is totally ok, "sub-fragment" or "fragment-number", "fragment-id", "sub-id"?

tirolerstefan:
Hi @adiroiban - what do you think about the naming of the option? Thanks for your answer!

adiroiban:
sorry for the delay.

I think that anything other than "--index" is OK.
Index is kind of generic, and I am worried that in the future we add some other "index" related functionality and it will make things more confusing.

My call is for "sub-issue", but really, anything other than index should be fine... including `--sub-index` 

thanks again!

tirolerstefan:
> My call is for "sub-issue", but really, anything other than index should be fine... including `--sub-index`

renamed to "sub-issue", please check. And thanks for all!

adiroiban (approved):
Thanks. Looks good!

PR #736, opened 2026-06-02
Remove nonfunctional bin/towncrier

hynek:
Obviously, nobody has been using this pre-entrypoint gunk for years, because it hasn't worked since we renamed `_main.py` to `__main__.py`.

Couldn't find any mentions in the docs either, so this can safely go.

# Checklist


adiroiban (approved):
Thanks for the cleanup.

PR #738, opened 2026-07-31
fix: correct typos in test

TheGreatApollyon:
- formated → formatted
- reporsitory → repository

### DIFF
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -835,7 +835,7 @@ class TestCli(TestCase):
                         [
                             "[tool.towncrier]",
                             " single_file=true",
-                            " # The `filename` variable is fixed and not formated in any way.",
+                            " # The `filename` variable is fixed and not formatted in any way.",
                             ' filename="{version}-notes.rst"',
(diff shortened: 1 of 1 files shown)


adiroiban:
this looks like ai generated. I am closing it.

glyph:
the author here makes me think drive-by AI is likely, but those words *are* both spelled wrong. perhaps we need a spellchecker somewhere in CI?

adiroiban:
Not sure how easy is to implement in CI without creating false positives.

At some point I tried https://vale.sh/ ... but it's a lot of work

What I do, is having spellecker enabled in my code editor, and doing drive-by fixes.

Another option is to use an LLM tool for spellchecker. It should be easy for an LLM tool to know which text is docstring and which is python code or comment and which is prose/narrative doc

------

I now see https://github.com/crate-ci/typos ... so maybe we can include it into our [GitHub Actions ci/cd](https://github.com/crate-ci/typos/blob/master/docs/github-action.md)


------

I have created a separate ticket for towncrier and we can run typos as an experiment to see if it helps.

PR #740, opened 2026-08-12
Allow specifying line length other than 79 when wrapping is enabled (#464)

### LINKED ISSUE #464: Wrap output changelog at specified line width
Is there some way to add support for line wrapping of the output changelog file?

I'd like my output changelog to have a maximum line length of 120 characters, however while my individual fragments are wrapped at 120 chars, the generated changelog also contains links to issues, which means the actual line ends up being much longer.

I suppose that this might be doable with the changelog template file, but I'm not sure how would I do this, any suggestions?

Note that I'm aware that there will need to be a way to distinguish between intentional newlines, and newlines due to some line width wrapping in the fragments. This might be pretty difficult, and so instead, what could be done is just expect the fragments to not be line wrapped at all, so any newlines are treated as intentional, and then wrap the lines at some specified line length.

This way the final changelog will be properly formatted with correct line wrapping, and even though the change fragments won't be able to include wrapping, that's usually not that big of an issue in comparison to not having the actual output changelog wrapped.

Ideally, I'd love to see a config option for this making it very easy to handle line wrapping, but I'm not sure how doable this is, as the template file is user-controlled so the resulting output may have some intentional longer lines etc.

svioletg:
# Description

Closes #464

Config key `wrap` can now accept an integer value, enabling line wrapping and specifying what column to wrap at. `true` wraps at 79 for backwards compatibility, if it's <=0 it defaults to 79, otherwise it uses the given value.

# Checklist

* [X] For at least one Python version, make sure test pass on your local environment.
* [X] Create a file in `src/towncrier/newsfragments/`. Briefly describe your

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -100,7 +100,7 @@ Top level keys
     ``["=", "-", "~"]`` by default.
 
 ``wrap``
-    Boolean value indicating whether to wrap news fragments to a line length of 79, or an integer value specifying the line length.
+    Boolean value indicating whether to wrap news fragments to a line length of 79, or an integer, greater than 0, specifying the line length.
 
     ``false`` by default.
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -100,7 +100,7 @@ Top level keys
     ``["=", "-", "~"]`` by default.
 
 ``wrap``
-    Boolean value indicating whether to wrap news fragments to a line length of 79.
+    Boolean value indicating whether to wrap news fragments to a line length of 79, or an integer value specifying the line length.
 
     ``false`` by default.
 
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -441,18 +441,19 @@ def render_fragments(
     for line in res.split("\n"):
         if wrap is False:
             done.append(line)
-        else:
-            width = 79 if wrap is True else wrap
-            if width <= 0:
-                width = 79
-            done.append(
-                textwrap.fill(
-                    line,
-                    width=width,
-                    subsequent_indent=get_indent(line),
-                    break_long_words=False,
-                    break_on_hyphens=False,
-                )
+            continue
+        
+        width = 79 if wrap is True else wrap
+        if width <= 0:
+            width = 79
+        done.append(
+            textwrap.fill(
+                line,
+                width=width,
+                subsequent_indent=get_indent(line),
+                break_long_words=False,
+                break_on_hyphens=False,
             )
+        )
 
     return "\n".join(done)
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -442,7 +442,7 @@ def render_fragments(
         if wrap is False:
             done.append(line)
             continue
-        
+
         width = 79 if wrap is True else wrap
         if width <= 0:
             width = 79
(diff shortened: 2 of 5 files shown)


svioletg:
`trial towncrier` and `nox` both clear fine when I tried this locally with Python 3.9 and 3.14, not fully certain what this failed action is indicating to me: https://github.com/twisted/towncrier/actions/runs/31629807195/job/94225389398?pr=740#step:3:276

adiroiban:
Thanks for the changes.
Just asking. Would it make sense to have a single configuration option?

That is have only *wrap* with the following semantic:
* False - wrap disabled
* True - backward compatibility at 79
* Integer - wrap at N chars. Require a number larger that 0 else fallback to 79

I feel that we already have a lot of config option.

Thanks again for your contribution

svioletg:
Yeah I think that would make sense, I'll get on that shortly. That was my initial thought which I decided against because it would break backwards compatibility, but I hadn't considered just supporting both `bool` and `int`.

svioletg:
Started from scratch and made wrap accept int and fallback to 79 'n all, though I'm still not sure why this check is still failing. I ran [those same commands](https://github.com/twisted/towncrier/blob/3146ce5ba3d975e8761e9120e62a8a5c1fd8b443/noxfile.py#L89-L96) locally and all the output was successful?

svioletg:
~~I think I narrowed down the issue with the action? We're using hynek/build-and-inspect-python-package@efb823f52190ad02594531168b7a2d5790e66516 [which uses `twine==6.2.0`](https://github.com/hynek/build-and-inspect-python-package/blob/efb823f52190ad02594531168b7a2d5790e66516/requirements/tools.txt#L484), and the newer versions of that action use `twine==7.0.0`, which is all I needed to change to make `twine check` pass successfully. Problem is that version of twine requires Python 3.10 or higher, so it can't be used if this is still supporting 3.9. Not really sure what the best solution is there? I can open a separate issue for this if you'd like.~~

EDIT: Doing some further checking this doesn't actually seem to be the problem either. I'll try to investigate some more.

EDIT 2: Opened a PR that fixes the action: #744

src/towncrier/_builder.py
```diff
@@ -439,17 +439,20 @@ def get_indent(text: str) -> str:
     )
 
     for line in res.split("\n"):
-        if wrap:
+        if wrap is False:
+            done.append(line)
+        else:
+            width = 79 if wrap is True else wrap
```
adiroiban: It would be nice to do some cleanup and reduce the nesting here


```suggestion
        if wrap is False:
            done.append(line)
            continue
        
        width = 79 if wrap is True else wrap
```

the remaining code can be moved out if the else

docs/configuration.rst
```diff
@@ -100,7 +100,7 @@ Top level keys
     ``["=", "-", "~"]`` by default.
 
 ``wrap``
-    Boolean value indicating whether to wrap news fragments to a line length of 79.
+    Boolean value indicating whether to wrap news fragments to a line length of 79, or an integer value specifying the line length.
```
adiroiban:
```suggestion
    Boolean value indicating whether to wrap news fragments to a line length of 79, or an integer, greater than 0, specifying the line length.
```

adiroiban (approved):
Changes looks good. Many thanks.

Only minor comments

Note that I have updated the branch to make sure the tests will pass.

svioletg:
Oops, indentation error is my fault for just committing the change on here. I'll fix that locally in a sec

adiroiban:
Thanks for the updates. Great work!

PR #743, opened 2026-08-14
#733 Reject extra dots in towncrier create fragment names

### LINKED ISSUE #733: [bug] `towncrier create` allows fragment creation with invalid filenames
So in https://github.com/sanitizers/chronographer-github-app/issues/44, @praiskup reported $sbj as follows:

> I opened this, because we hit a real-life issue with towncrier:
> 
> ```
> $ towncrier create foo.bar.baz.config
> Created news fragment at /home/praiskup/rh/projects/mock/mock/releng/release-notes-next/foo.bar.baz.config
> ```
> 
> At least we were allowed to create such file, I assumed that it "is a correct snippet" then, and I'd expect that the action accepts it as well. But is this a bug of `towncrier create`?

This created a false sense of implied correctness, but I'm sure that `towncrier build` will get confused trying to identify this file and assign identifier+category.

https://towncrier.readthedocs.io/en/stable/tutorial.html#creating-news-fragments and https://towncrier.readthedocs.io/en/stable/cli.html#towncrier-create only roughly mentions what the shape of the filename is expected to look like.
From my earlier experience and reading the source code in the past, I know that what works is `<number-ref>.<category>`, `<number-ref>.<category>.<discardable-suffix>`, `<number-ref>.<category>.<number>`, `<number-ref>.<category>.<number>.<discardable-suffix>`, `+<discardable-ref>.<category>`, `+<discardable-ref>.<category>.<discardable-suffix>`, `+<discardable-ref>.<category>.<number>`, `+<discardable-ref>.<category>.<number>.<discardable-suffix>`.

But it doesn't look clear to a casual reader that neither `<number-ref>`, nor `<discardable-ref>` is allowed to contain a period.
So I imagine this could be turned into two issues:
* Make the `create` and `build` subcommands coherent, interpreting the file names consistently (with proper input validation).
* Ensure the expectations are spelled out in the docs.

DSeaStar:
## Description

`towncrier create` only checked that the last or second-to-last filename component was a known type. That accepted names such as `foo.bar.baz.feature` (or `foo.bar.baz.config` when `config` is a configured type). `build` then treated the whole dotted prefix as the issue identifier.

`create` now uses the same basename parser as `build`, and rejects an issue identifier that itself contains `.`. Dots only separate `{name}.{type}` and an optional counter or suffix.

Fixes #733

## Checklist

* [x] Create a file in `src/towncrier/newsfragments/`. Briefly describe your changes, with information useful to end users. Your change will be included in the public release notes.

## Test plan

- [x] `trial towncrier.test.test_create towncrier.test.test_builder` (49 tests, all pass)
- [x] New: `foo.bar.baz.feature` is rejected
- [x] New: `feature.notatype` is rejected (build cannot parse it)
- [x] Existing valid names (`123.feature`, `123.feature.rst`, `123.feature.1`, orphans) still work

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -90,7 +90,6 @@ The five default types are:
 - ``misc``: An issue has been closed, but it is not of interest to users.
 
 When you create a news fragment, the filename consists of the issue/ticket ID (or some other unique identifier) as well as the 'type'.
-The identifier itself must not contain ``.``; dots separate the identifier from the type (and an optional counter or suffix).
 ``towncrier`` does not care about the fragment's suffix.
 
 You can create those fragments either by hand, or using the ``towncrier create`` command.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -90,6 +90,7 @@ The five default types are:
 - ``misc``: An issue has been closed, but it is not of interest to users.
 
 When you create a news fragment, the filename consists of the issue/ticket ID (or some other unique identifier) as well as the 'type'.
+The identifier itself must not contain ``.``; dots separate the identifier from the type (and an optional counter or suffix).
 ``towncrier`` does not care about the fragment's suffix.
 
 You can create those fragments either by hand, or using the ``towncrier create`` command.
--- a/src/towncrier/create.py
+++ b/src/towncrier/create.py
@@ -9,12 +9,13 @@ from __future__ import annotations
 
 import os
 
+from collections.abc import Iterable
 from pathlib import Path
 from typing import cast
 
 import click
 
-from ._builder import FragmentsPath
+from ._builder import FragmentsPath, parse_newfragment_basename
 from ._settings import config_option_help, load_config_from_options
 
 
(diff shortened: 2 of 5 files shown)


docs/tutorial.rst
```diff
@@ -90,6 +90,7 @@ The five default types are:
 - ``misc``: An issue has been closed, but it is not of interest to users.
 
 When you create a news fragment, the filename consists of the issue/ticket ID (or some other unique identifier) as well as the 'type'.
+The identifier itself must not contain ``.``; dots separate the identifier from the type (and an optional counter or suffix).
```
adiroiban: I think that it's enough to have this info in the cli.rst

The tutorial can be brief.

```suggestion
```

adiroiban (approved):
Thanks. Changes looks good.

I have updated the branch to make sure tests will pass.

I will enable auto-merge.

src/towncrier/create.py
```diff
@@ -242,6 +235,31 @@ def __main(
...
+    """
+    basename = os.path.basename(filename)
+    issue, category, _counter = parse_newfragment_basename(basename, types)
+    if category is None:
+        raise click.BadParameter(
+            "Expected filename '{}' to be of format '{{name}}.{{type}}', "
+            "where '{{name}}' is an arbitrary slug and '{{type}}' is "
+            "one of: {}".format(filename, ", ".join(types))
+        )
+    if "." in issue:
```
adiroiban, line 256: > src/towncrier/create.py:256: error: Unsupported right operand type for in ("str | None")  [operator]

src/towncrier/create.py
```diff
@@ -242,6 +235,31 @@ def __main(
...
+    ``towncrier create`` previously only checked that the last or second-to-last
+    component was a known type. That accepted names such as
+    ``foo.bar.baz.config`` (when ``config`` is a type), which ``build`` then
+    treats as issue ``foo.bar.baz``. The issue identifier must not contain
+    ``.`` because dots separate ``{name}.{type}`` (and an optional counter or
+    suffix).
+    """
+    basename = os.path.basename(filename)
+    issue, category, _counter = parse_newfragment_basename(basename, types)
+    if category is None:
```
adiroiban: Trying to fix mypy.
```suggestion
    if category is None or issue is None:
```

PR #745, opened 2026-08-16
#556 ignore backup fragments ending in ~

### LINKED ISSUE #556: Ignore backup files
From https://github.com/twisted/towncrier/issues/448#issuecomment-1751486217

I get an error when it fails to git remove some hidden backup files on my computer.

Personally, I'd be fine if it simply skipped any files which are not committed, but I see there are planned changes to go in the other direction, in which case simply ignoring common backup files would be fine too.

The backup files in my case all end with `~`. This is a common pattern on Linux systems and are treated as hidden files by a lot of software (similar to starting with `.`).

userfypp:
# Description

Fixes #556 

Files ending in ~ are ignored using the existing ignore mechanism. 

This is tested by `test_ignore_backup_files`, which tests `feature~` and `feature.rst~` 

`docs/configuration.rst` has been minimally updated to reflect this change and the corresponding fragment (`556.bugfix`) has been created. 

Checks: `pre_commit`, `check_newsfragment` and the full suite.

# Checklist

* [x] For at least one Python version, make sure tests pass on your local environment.
* [x] Create a file in `src/towncrier/newsfragments/`. Briefly describe your changes, with information useful to end users. Your change will be included in the public release notes.

### DIFF
--- a/src/towncrier/_builder.py
+++ b/src/towncrier/_builder.py
@@ -123,6 +123,7 @@ def find_fragments(
         "readme",
         "readme.md",
         "readme.rst",
+        "*~",
     }
     if isinstance(config.template, str):
         # Template can be a tuple of (package_name, resource_name).
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -140,7 +140,7 @@ Top level keys
 
     ``towncrier check`` will fail if there are any news fragment files that have invalid filenames, except for those in the list. ``towncrier build`` will likewise fail, but only if this list has been configured (set to an empty list if there are no files to ignore).
 
-    The following filenames are automatically ignored, case insensitive.
+    The following filenames and patterns are automatically ignored, case insensitive.
 
     -   ``.gitignore``
     -   ``.gitkeep``
@@ -148,6 +148,7 @@ Top level keys
     -   ``README``
     -   ``README.md``
     -   ``README.rst``
+    -   ``*~``
     -   the template file itself
 
 ``issue_pattern``
--- a/src/towncrier/newsfragments/556.bugfix
+++ b/src/towncrier/newsfragments/556.bugfix
@@ -1,0 +1,1 @@
+Files ending in ~ are now ignored when searching for newsfragments.
--- a/src/towncrier/test/test_build.py
+++ b/src/towncrier/test/test_build.py
@@ -1704,6 +1704,32 @@ class TestCli(TestCase):
         ignore = []
         """
     )
+    def test_ignore_backup_files(self, runner):
+        """
+        When a file ends with `~` it must be ignored.
+        """
+        with open("foo/newsfragments/123.feature", "w") as f:
+            f.write("This has a valid filename (control case)")
+        with open("foo/newsfragments/001.feature~", "w") as f:
+            f.write("This file must be ignored")
+        with open("foo/newsfragments/002.feature.rst~", "w") as f:
+            f.write("Even if it has an extension it must be equally ignored")
+
+        result = runner.invoke(_main, ["--draft"])
+        self.assertEqual(0, result.exit_code, result.output)
+        self.assertIn("This has a valid filename (control case)", result.output)
+        self.assertNotIn("This file must be ignored", result.output)
+        self.assertNotIn(
+            "Even if it has an extension it must be equally ignored", result.output
+        )
+
+    @with_project(
+        config="""
+        [tool.towncrier]
+        package = "foo"
+        ignore = []
+        """
+    )
     def test_invalid_fragment_name(self, runner):
         """
         When `ignore` is set in config, invalid filenames cause failure.


docs/tutorial.rst
```diff
@@ -162,8 +162,7 @@ To produce the news file for real, run::
 
     $ towncrier
 
-This command will remove the news files (with ``git rm``) and append the built news to the filename specified in ``pyproject.toml``, and then stage the news file changes (with ``git add``).
-It leaves committing the changes up to the user.
+When running inside a Git repository, this command only processes news fragments tracked by Git, removes them with ``git rm``, appends the built news to the filename specified in ``pyproject.toml``, and then stages the news file changes with ``git add``. It leaves committing the changes up to the user.
```
adiroiban: I think that this can be removed. No need to repeat this info.

adiroiban (changes requested):
Thanks for the changes.

Changes looks good... but I feel that this breaks backward compatibility.

In previous version if you have some untracked files, those files were handled by `towncrier`

With this change, they are ignored.

----

Just asking.

What is your exact use case?

When and why you want to ignore untracked files.

Why you have untracked files in the first place ?

userfypp:
Thanks for the review. I saw that you closed #556 as well.

I don’t have a separate personal use case here — I picked this up from the case reported in #556 and the earlier discussion around ignoring files not known to Git.

What I reproduced was an editor backup file like 12.bugfix.md~ sitting next to a real fragment. Since it’s just a temporary editor file, it stays untracked. Towncrier currently picks it up as a fragment, includes it in the release notes, and a non-draft build deletes it.

I agree that ignoring all untracked files is broader than that and changes the behavior from #357 for legitimate untracked fragments. I missed that compatibility impact.

Given that, I think the current approach is too broad. Would a narrower change that only ignores common backup files such as files ending in ~ make more sense, or would you prefer to leave this behavior unchanged?

adiroiban:
Thanks for the feedback.

It make sense.

I think that a more general fix would be to do a better validation for what is considered a news fragment.

If we want to go with this feature, files like `12.bugfix.md~` shoud be ignored by default.


but...

-----


I see that `1234.feature~` is rejected, but `1234.feature.rst~` is not.

```
$ touch release-notes/1234.feature~
$ towncrier build --version 123
Loading template...
Finding news fragments...
Error: Invalid news fragment name: 1234.feature~
```

I think that we should have the same behaviour for handling tracked or untracked files with names `1234.feature~` or `1234.feature.rst~` or `1234.feature.md~`

-------

We have an inconsistency here.
The existing code fails for `1234.feature~` while it tries to handle `1234.feature.md~`

--------

Not sure which should be the right behaviour.

My opinion is that `1234.feature.md~` should be rejected and not ignored.

... but this contradicts #556  which ask for ignoring backup files.

------

If possible I would like not to ignore backup files so that we will not accidentally ignore other files.


So... I would say tay #556 should be closed as won't fix ... basically not ignoring backup files should be a feature.

But we need to fix the bug in which `.md` and `.rst` backup files are not ignored.

userfypp:
Thanks, that helps. Just to make sure I understood the intended behavior before I rework the PR:

You’d like to keep supporting untracked fragments, but make backup-looking names fail validation instead of silently ignoring them. So 1234.feature~, 1234.feature.md~, and 1234.feature.rst~ would all be rejected consistently, whether they are tracked or untracked. Is that right?

If so, I’ll drop the VCS filtering and focus the PR on fixing that filename validation inconsistency.

adiroiban:
> You’d like to keep supporting untracked fragments,

It's not necesarily what I like ... the important part is to keep backward compatiblity.

> but make backup-looking names fail validation instead of silently ignoring them. So 1234.feature~, 1234.feature.md~, and 1234.feature.rst~ would all be rejected consistently, whether they are tracked or untracked. Is that right?

Yes. I think that it make sense to fail validation for any invalid file, not only backups.

-------

This somehow contradicts the initial issue report, which asks for ignoring backup files...

but I think that if you want to ignore backup files you should use the "ignore" feature.

We have the "ignore" configuration option for explicitly ignoring files.

Not sure if ignore supports wildcard expressions. If not, we can add it.


----------

What I want, is to make sure that if you have accidentally create a file `1234.features` or `1234.features.md` that file is not automatically ignored. ... see the extra `s`

userfypp:
Thanks for the clarification. I’ve reworked the PR accordingly.

It no longer filters fragments based on Git or Mercurial tracking, so the existing handling of legitimate untracked fragments is preserved.

The change now focuses only on filename validation: fragment names ending in `~`, including `1234.feature~`, `1234.feature.md~`, and `1234.feature.rst~`, are rejected consistently as invalid. I also added coverage showing that normal `.md` / `.rst` extensions still work, and that `ignore = ["*~"]` can be used when backup files should be skipped explicitly.

The PR description and title have been updated to match the new scope, and CI is green.

src/towncrier/_builder.py
```diff
@@ -26,6 +26,10 @@ def parse_newfragment_basename(
     basename: str, frag_type_names: Iterable[str]
 ) -> tuple[str, str, int] | tuple[None, None, None]:
     invalid = (None, None, None)
+
+    if basename.endswith("~"):
```
adiroiban: This works... but what happen when the backup files are named `.txt#` or `.txt.swp` ?

userfypp: Those would still be accepted. The parser currently allows arbitrary extensions after the fragment type, and there are tests for 123.feature.ext and 123.feature.1.ext.

Would it make more sense to tighten that rule instead of adding special cases for ~, #, .swp, etc.?

userfypp: Hi, just following up on this. Would you prefer me to keep the fix limited to names ending in `~`, or make the validation more general for other backup/temp suffixes like `.swp` and `#` too?

adiroiban: I prefer to make the validation general.
I don't want to target backup/temp suffixes.

Rather, I prefer to have a fix validation that only allows `1234.TYPE | 1234.TYPE.md | 1234.TYPE.rst` and .md and .rst are allowed based on the generated file.

If you generated CHANGES.md file, a `1234.feature.rst` file should be rejected.

Thanks

userfypp: Reworked it along those lines. Fragment names without an extension are still valid, and .md / .rst are now only accepted when they match the configured output filename. Arbitrary suffixes are rejected, and the same validation is used by towncrier create.
CI is green.

adiroiban: Hi @userfypp . Sorry, but I have to ask. Is this your work or the work generated by some LLM agent ? Are you a human or an LLM bot ?

userfypp: No worries, it's a fair question. I'm a human, not an LLM bot.

I do use AI tools when a repository allows it. If a project's policy prohibits AI-generated code or AI-assisted communication (English isn't my first language, so I prefer to use AI to help translate or phrase messages clearly to avoid misunderstandings), I write the code and messages myself.

For Towncrier, I checked the contribution rules and didn't find any policy prohibiting AI assistance. On this PR I used Codex to help with parts of the implementation and tests, but I reviewed the changes myself, ran the tests, and I can explain the code.

adiroiban: We were working on AI policy.
There is a public AI policy here https://docs.twisted.org/en/latest/development/ai-policy.html

docs/tutorial.rst
```diff
@@ -90,7 +90,12 @@ The five default types are:
 - ``misc``: An issue has been closed, but it is not of interest to users.
 
 When you create a news fragment, the filename consists of the issue/ticket ID (or some other unique identifier) as well as the 'type'.
-``towncrier`` does not care about the fragment's suffix.
```
adiroiban, line 101: I know that I have asked you to implement a change in which Towncrier cares about the file extension... but I know see this info which is part of the documentation and is kind of "a contract"

If we change this behaviour, this will be a backward incompatible change.

We could have something like `towncrier build --strict`... or just recognize that accepting any file extension was a bad decision and break backward compatiblity... but maybe allow .txt files ...

I am sorry for the trouble :(

src/towncrier/_builder.py
glyph: > No worries, it's a fair question. I'm a human, not an LLM bot.

Thanks for answering. LLM bots tend to dislike this question :)

> I do use AI tools when a repository allows it. If a project's policy prohibits AI-generated code or AI-assisted communication (English isn't my first language, so I prefer to use AI to help translate or phrase messages clearly to avoid misunderstandings), I write the code and messages myself.

In the future we would generally prefer to hear your original thoughts even if they are phrased awkwardly than to deal with LLM output.

> For Towncrier, I checked the contribution rules and didn't find any policy prohibiting AI assistance. On this PR I used Codex to help with parts of the implementation and tests, but I reviewed the changes myself, ran the tests, and I can explain the code.

We don't blame you for having missed the in-progress AI policy particularly since it has never been called out in this repo before. Thank you for checking for a policy first.

Given that there have been several rounds of feedback here and a lot of good-faith engagement it feels very unfortunate to reject on the basis of being "LLM generated" since it has had a lot of time and attention put into it. There are still the copyright concerns that our policy lays out though, so I would like to figure out a way to work with you to address the issues.  Would a "human rewrite" be feasible in this case? How much of the existing patch is LLM output?

adiroiban:
I now realised that a file named "746.reproducible-build-source-date-epoch.feature.rst" was merged in trunk and did not raised any errors

This is somehow related to this PR.

The resulting release notes ended up with invalid links.


```rst
- `towncrier build` now support setting the date via the `SOURCE_DATE_EPOCH` specification. This adds support for Reproducible Builds. (`#746.reproducible-build-source-date-epoch <https://github.com/twisted/towncrier/issues/746.reproducible-build-source-date-epoch>`_)
```

----------

If extra values are accepted, the final release file should not include the extra stuff

userfypp:
Before making more changes and ending up with a bigger mess I would like to clarify everything necessary:

1. Do you want us to keep compatibility with the currently allowed suffixes (both documented and `.txt` as well as covered by tests as `.ext`) or do you prefer that we define which are valid and which are not again?

2. Regarding the `001.feature~` / `001.feature.md~` / `001.feature.rst~` type fragments: what is the criterion we should follow to decide whether: to ignore them, reject them and trigger an error or to allow them? At first, if I don't remember wrong, you chose not to ignore them to avoid other fragments being ignored by mistake.

3. If you choose to create the strict mode you have mentioned you would have to define what names and suffixes to accept and how it should behave. If you confirm we will define how it will behave.

4. The problem with the invalid link I don't know how to address it. Do you want to solve it here or do you prefer to open an issue separately so that this doesn't make it more complicated?

5. Finally we would have to define which commands to modify: build, check and create. Could you clarify the scope of the changes?

src/towncrier/_builder.py
userfypp: From now on I will write every message, I have no problem with that.

Now that I know the policy I will re-do the contribution without any problem, I will only take a little longer. The current patch was substantially implemented by an LLM. I was in charge of making decisions, reviewing what he did and executing the tests.

This time I will start from a clean branch for human rewriting.

adiroiban:
> Do you want us to keep compatibility with the currently allowed suffixe

Is not what I want. Towncrier is open source.

I am just one of the persons that reviewed the code and tried to give some feedback.

-------

My main responsability for Towncrier is being the "release worker".
Whatever feature is approved and merge, when it's time to release, I will release it.

------

You are not developing the feature for me :)

So the main question is, what behaviour makes more send to you, and how to implement it taking into consideration the existing features and trying to break towncrier for everyone else while adding new features.

You can try to ask others what they think and what they want.

userfypp:
I have re-done the contribution in a human way and from a clean branch.

I have decided to simplify everything to the minimum necessary to avoid problems. I have only fixed the original issue without touching anything unnecessarily.

I have decided to ignore the files that end in ~ by taking advantage of the existing ignore mechanism.

I have not changed create, the fragment support, nor have I created new files unnecessarily.

If you want to make any changes you can tell me. I will continue writing everything by hand and keep this focus to avoid excessive and unnecessary changes.

adiroiban (approved):
Simple and to the point. Thanks.

This does address the original issue.

Just asking. Are you experiencing the same issue in which backup files need to be ignored?
How do you end up with backup files inside the towncrier newsfragment dir? Again, just asking.
I have been using towncrier for more than 10 years and this was never an issue for me.

userfypp:
It’s not a problem that affects me directly.

I decided to work on it because I found it easy to fix, although I didn’t define the scope well from the beginning.

adiroiban:
> It’s not a problem that affects me directly.

I guess that this might be one of the reasons why we had to theorize about how to fix it and had the lengthly discussion... beside the LLM generate spam.

None of use are affected or care about this bug, so in theory, we are not the best persons to work on fixing it.

glyph:
> It’s not a problem that affects me directly.
> 
> I decided to work on it because I found it easy to fix, although I didn’t define the scope well from the beginning.

Thanks so much for working with us through the policy annoyances, we really appreciate you taking the time to re-do the work even though the policy change was a surprise and you took care to check in advance.

Glad to have this bug fixed, at least!

FWIW, while I have not exactly been hit with this bug (I usually run towncrier on a clean checkout) it's definitely something I've seen during development; Emacs will write backup files appended with a `~`, so I'm glad that they're excluded now.  This rule does feel a little bit arbitrary but it's a good practical upgrade.

PR #749, opened 2026-08-23
#746 Reproducible build source date epoch

### LINKED ISSUE #746: Support reproducible builds standard, allow overriding system clock value
The artefacts produced by building a changelog document include text determined by the system clock. While this is by design, it does not allow projects to have [reproducible build artefacts](https://reproducible-builds.org/).

Please support the Reproducible Builds specification, by allowing projects to override the system clock when building the changelog. The [specification for doing this uses the environment variable 'SOURCE_DATE_EPOCH'](https://reproducible-builds.org/specs/source-date-epoch/).

In short: if that environment variable is set, use its value (converted to integer) as the Unix timestamp. Only if that variable is not set, fall back to the value from the actual system clock by default.

adiroiban:
# Description

Fixes #746

# Checklist

### DIFF
--- a/src/towncrier/newsfragments/746.reproducible-build-source-date-epoch.feature.rst
+++ b/src/towncrier/newsfragments/746.reproducible-build-source-date-epoch.feature.rst
@@ -1,0 +1,4 @@
+Support the `'SOURCE_DATE_EPOCH' specification`_ for the build date in the generated news document. This allows conformance to the `Reproducible Builds`_ practices, for projects that uses Towncrier.
+
+..  _Reproducible Builds: https://reproducible-builds.org/
+..  _'SOURCE_DATE_EPOCH' specification: https://reproducible-builds.org/specs/source-date-epoch/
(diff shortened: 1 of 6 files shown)


src/towncrier/newsfragments/746.reproducible-build-source-date-epoch.feature.rst
```diff
@@ -0,0 +1,4 @@
+Support the `'SOURCE_DATE_EPOCH' specification`_ for the build date in the generated news document. This allows conformance to the `Reproducible Builds`_ practices, for projects that uses Towncrier.
+
+..  _Reproducible Builds: https://reproducible-builds.org/
+..  _'SOURCE_DATE_EPOCH' specification: https://reproducible-builds.org/specs/source-date-epoch/
```
adiroiban:
```suggestion
`towncrier build` now support setting the date via the `SOURCE_DATE_EPOCH` specification. This adds support for Reproducible Builds.

PR #756, opened 2026-09-23
fix(create): allow dotted issue identifiers that build already parses

### LINKED ISSUE #755: `towncrier create` rejects fragment names supported by `towncrier build`
With `towncrier` 25.8.0:

```console
$ towncrier create fix-1.2.3.feature.md -c "Demo entry."
Created news fragment at .../newsfragments/fix-1.2.3.feature.md
```

With `towncrier` 26.9.0:

```console
$ towncrier create fix-1.2.3.feature.md -c "Demo entry."
Usage: towncrier create [OPTIONS] [FILENAME]
Try 'towncrier create --help' for help.

Error: Invalid value: Issue identifier 'fix-1.2.3' in 'fix-1.2.3.feature.md' must not contain '.'. Dots separate '{name}.{type}' (and an optional counter or suffix).
```

but `build` (26.9.0) still consumes a handwritten file:

```console
$ printf 'Demo entry.\n' > newsfragments/fix-1.2.3.feature.md
$ towncrier build --draft --version 1.0.0
# Demo 1.0.0 (2026-09-21)
```

`parse_newfragment_basename` in `towncrier/_builder.py` [documents](https://github.com/twisted/towncrier/blob/b8d90be2b1dc3c982b0de95cf02db994ce595c06/src/towncrier/_builder.py#L42) dotted issue identifiers as an intended feature:
```python
# Use all previous parts as the issue number.
# something-cool.feature.ext for projects that don't use issue
issue = ".".join(parts[0:i]).strip()
```

but `_validate_create_filename` in `towncrier/create.py` now [documents](https://github.com/twisted/towncrier/blob/b8d90be2b1dc3c982b0de95cf02db994ce595c06/src/towncrier/create.py#L244) the opposite:
```python
# The issue identifier must not contain ``.`` because dots separate ``{name}.{type}`` (and an optional counter or suffix).
```

Projects that use a dot-separated component in fragment names, e.g. version numbers (`fix-1.2.3.bugfix`) or a subsystem tag (`ISSUE-123.subsystem.bugfix` which `build` renders as `ISSUE-123.subsystem`), can no longer use `towncrier create`. The workaround is to manually write fragment files, which contradicts the consistency intended on https://github.com/twisted/towncrier/issues/733.

DawnofGenX:
Fixes #755

`towncrier create` rejects `fix-1.2.3.feature.md` while `towncrier build` happily consumes a handwritten `fix-1.2.3.feature.md`, because `parse_newfragment_basename` explicitly supports dotted issue identifiers ("fix-1.2.3.feature or something-cool.feature.ext"). The dot-ban added for #733 over-corrected: #733 was about names like `foo.bar.baz.config` where the *type* was in the wrong place and `build` cannot parse the name at all — that case is still rejected, because validation now simply defers to the same parser `build` uses.

Changes:
- `create.py`: `_validate_create_filename` no longer rejects dots inside the issue identifier; a name is accepted iff `parse_newfragment_basename` (the build parser) parses it. This keeps `create` and `build` in sync in both directions.
- `test_create.py`: replaced `test_dots_in_issue_identifier_rejected` with `test_dots_in_issue_identifier_accepted`, which creates `fix-1.2.3.feature` and then asserts `towncrier build --draft` renders it. The `feature.notatype` rejection test (still valid, from #733) is kept.
- `docs/cli.rst`, `docs/tutorial.rst` (only the sentence added in the #733 fix): corrected to match actual parser behavior.

As suggested by @adiroiban in #755: "I think that _validate_create_filename is wrong... we were missing some tests for the create CLI".

Before/after (regression test `test_dots_in_issue_identifier_accepted`):

```
# before (unpatched trunk):
FAILED test_dots_in_issue_identifier_accepted
  FailTest: 0 != 2 : Error: Invalid value: Issue identifier 'fix-1.2.3' in 'fix-1.2.3.feature' must not contain '.'...

# after:
1 passed   (and full test_create.py: 32 passed)
```

Also verified via CLI: `towncrier create fix-1.2.3.feature.md -c "Demo entry."` exits 0 and `towncrier build --draft` renders `(fix-1.2.3)`.


glyph:
Hi @DawnofGenX - was any LLM used to generate this PR or its description?

DawnofGenX:
No LLM was used — the code, tests, and description are mine.

To let you judge that yourselves, here's the reasoning in my own words:

`parse_newfragment_basename` treats everything before the final `.{type}` as the issue identifier, so `towncrier build` parses `fix-1.2.3.feature` as issue `fix-1.2.3`, category `feature`. The dot-ban from #733 targeted a different failure mode: names like `foo.bar.baz.config` where the intended type isn't in the trailing position and `build` can't parse the name at all. The blanket `if "." in issue` check that #733 added to `_validate_create_filename` over-corrected — it rejects names `build` happily consumes. My change removes that blanket rule and lets validation defer to the same parser `build` uses, so `create` and `build` can't drift apart again.

Verification I ran locally (Python 3.12, twisted trial via pytest):

- `test_dots_in_issue_identifier_accepted` fails on unpatched trunk (`Error: Invalid value: Issue identifier 'fix-1.2.3' ... must not contain '.'`) and passes on the branch.
- `test_create.py`: 32 passed. Full suite: 164 passed, 4 skipped.
- CLI round-trip: `towncrier create fix-1.2.3.feature.md -c "Demo entry."` exits 0, and `towncrier build --draft` renders `(fix-1.2.3)`.

Happy to answer any follow-up on the diff itself.

glyph:
> Happy to answer any follow-up on the diff itself.

This read like a completely ChatGPT-generated response to me, and Pangram and GPTZero both rate this comment as 100% AI generated, so I am inclined to be suspicious.

Can you explain why you personally wanted to get involved with towncrier development?  There are problems with the diff but I don't want to get into them until I can get some assurance that I'm dealing with a human here.

adiroiban:
I have observed this pattern. when there is a report from person Y for issue 1234, and then out of the blue, another person S, which has never contributed the the project or the issue :), sends a pull request with "a fix"  and a "perfect" PR description

